PR #2, opened 2020-06-18
EP - multi-cluster service selection

danibachar:


danibachar:
@mangelajo @nyechiel @dfarrell07

lighthouse/multi-cluster-service-selection.md
```diff
@@ -0,0 +1,71 @@
...
+While the world is moving towards micro-services application based architecture, more organizations combine multi-cluster and multi-cloud providers for redundancy, performance, management, and other reasons.
+We would like to optimize the total cost of a Kubernetes deployment in means of communication between multi-clusters, given that we could decide between different approaches. The three approaches possible, currently, are optimizing the full service chain of a single app/deployment, multiple apps/deployments, or a smart local optimization per cluster point of view, without building the full service chain.
+In this Enhancement Proposal, we shall introduce questions, ideas, and requirements to optimize the service selection mechanism of the Lighthouse Plugin.
+
+
+### Requirements
+1. Topology:
+    1. Selection Decision should be made locally, i.e per cluster - decentralized solution.
+    2. The solution must be adaptive for topology changes.
+1. Cost Parameters/Metrics:
```
mangelajo: 1. Service health (if the remote service is unhealthy no ready,or no healthy pods to back the request... ) we must completely avoid sending requests at all.

lighthouse/multi-cluster-service-selection.md
```diff
@@ -0,0 +1,71 @@
...
+    1. Latency
+    2. RTT
+    3. Networking cost
+    4. Network congestion
+    5. CPU/RAM load on each container/pod
```
mangelajo: Going beyond point into point 5 and beyond may be it's like stepping into service mesh territory. Although some sort of metric on how long does the service take to respond would be nice may be.

lighthouse/multi-cluster-service-selection.md
```diff
@@ -0,0 +1,71 @@
...
+    6. Service scale/autoscale (which actually will affect the load on the CPU and RAM)
+
+
+### Open Questions
+1. Topology:
+    1. What topology would we like to optimize?
```
mangelajo: Can you provide examples of what you mean?

lighthouse/multi-cluster-service-selection.md
```diff
@@ -0,0 +1,71 @@
...
+        1. Local optimization per cluster?
+        2. Per App optimization?
+        3. Multi App optimization?
+    2. How often do deployments affects changes in the topology?
+    3. How adaptive would we like to be for changes in the topology?
+    4. Would we like to build the full application graph for better optimization?
```
mangelajo: This sounds like stepping into service mesh territory.

lighthouse/multi-cluster-service-selection.md
```diff
@@ -0,0 +1,71 @@
...
+        1.  How much are we willing to pay for the data sharing through the broker system?
```
mangelajo: Sharing through the broker for changing objects is something we probably need to limit. If we will have objects that change often, and are many, may be we will need better mechanisms than the broker. (I looked into some sort of p2p implementation between the clusters, but we would need to analyze that better...)

lighthouse/multi-cluster-service-selection.md
```diff
@@ -0,0 +1,71 @@
...
+1. Cost Parameters/Metrices:
+    1.  How often would we like to collection the parameters?
+    2.  Would we like to make the collection configurable?
+1.  Optimization:
+    1. Would we like to take the processing time of each service in account?
```
mangelajo: It'd be cool, sounds like stepping into service mesh again, but see my thought on L19. May be we could have annotations for something like the health / liveness probes of pods, but for services or exported services.

lighthouse/multi-cluster-service-selection.md
```diff
@@ -0,0 +1,71 @@
...
+As we would like to comply with the Kubernetes KEP for the Multi-Cluster API, and the current implementation, we would like to offer a simple lightweight solution for selecting a service from a given pool.
+
+#### Priority Queue
+
+It seems that a data structure of a priority queue could be a simple elegant solution for selecting the best service per request.
+
+#### External API
+
+We would need to provide explicit API for updating the queue weights externally, as we would like to keep the implementation simple.
+Maybe adding an additional property such as `cost` to the `status` of a `ServiceExport`
```
mangelajo: Just a few notes:

1) I guess you want to suggest cost here as the local cluster perspective on the service (based on health, response time, stuff like that). because there is an important part of the cost that really depends on which cluster is originating the request, right?


2) Making changes to that object could be complicated, we would need to agree on that from k8s upstream, although we can try.

mangelajo (commented):
some comments/thoughts.

lighthouse/multi-cluster-service-selection.md
danibachar: Yep a good question, in topology I meant what we can see from the deployment perspective (i.e single application deployment/multiple application deployment or indeed just a simple local and somehow smart decision from the cluster perspective without taking into account the whole application graph)... I only asked that because according to literature the more we "know" about the application graph the better optimization we can offer
And indeed like you mentioned in other comments, it stepping into the service mesh territory which we need to understand if we want to tackle - from the comments I can understand we don't want that.

lighthouse/multi-cluster-service-selection.md
danibachar: 1+2) Yes, and I understand it is kind of breaking the Kubernetes Multi-Clusters API.
Although it was discussed as a possible next step (more true for Cluster config then actual service - my mistake):
1) https://groups.google.com/forum/#!msg/kubernetes-sig-network/wXd1D_fKjqU/SEwjsOfpAAAJ
2) https://github.com/kubernetes/kubernetes/issues/85395

The idea is to leave Lighthouse implementation as simple as possible and make the optimization out of plugin scope, with the option to enable some API to make updates to the weight/cost ( I would fix the proposal for weights it explains the idea in a better way).
But until then we should definitely think about either enabling some external API for updates or simplifying the optimization within the  Lighthouse plugin...
But again - might be stepping into the service mesh territory

lighthouse/multi-cluster-service-selection.md
danibachar: 👍

lighthouse/multi-cluster-service-selection.md
danibachar: How would you describe the boundaries of a Service Mesh? 
For example, if we have Autoscale we can have this information using a simple API and share that info...

lighthouse/multi-cluster-service-selection.md
danibachar: Great point, this is exactly the kind of insights we need to discuss... would you say that knowing the application graph or parts of it (enable look ahead for example) under any conditions will be something we would like to avoid?

lighthouse/multi-cluster-service-selection.md
danibachar: Yep, the important point is where to put the line between our solution and Service Mesh which can be looked like some kind of competitor (:

lighthouse/multi-cluster-service-selection.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Requirements
+1. Topology:
+    1. Selection Decision should be made locally, i.e per cluster - decentralized solution.
+    2. The solution must be adaptive for topology changes.
+1. Cost Parameters/Metrics:
+    1. Latency
+    2. RTT
+    3. Networking cost
+    4. Network congestion
```
mkolesnik: How would you measure this?

lighthouse/multi-cluster-service-selection.md
```diff
@@ -0,0 +1,71 @@
...
+In this Enhancement Proposal, we shall introduce questions, ideas, and requirements to optimize the service selection mechanism of the Lighthouse Plugin.
+
+
+### Requirements
+1. Topology:
+    1. Selection Decision should be made locally, i.e per cluster - decentralized solution.
+    2. The solution must be adaptive for topology changes.
+1. Cost Parameters/Metrics:
+    1. Latency
+    2. RTT
```
mkolesnik: This is ping, how'd you measure Latency though? Half this value?

lighthouse/multi-cluster-service-selection.md
mkolesnik: >  Although some sort of metric on how long does the service take to respond would be nice may be.

This is actually exactly service mesh territory. We should stay with basic network metrics & reachability.

lighthouse/multi-cluster-service-selection.md
mkolesnik: I agree with @mangelajo's point here that the cost is only from a local perspective. Metrics such as RTT can differe depending on the source cluster and probably need to be saved elsewhere.

lighthouse/multi-cluster-service-selection.md
```diff
@@ -0,0 +1,71 @@
...
+1.  Optimization:
+    1. Would we like to take the processing time of each service in account?
+
+
+### Design Details
+As we would like to comply with the Kubernetes KEP for the Multi-Cluster API, and the current implementation, we would like to offer a simple lightweight solution for selecting a service from a given pool.
+
+#### Priority Queue
+
+It seems that a data structure of a priority queue could be a simple elegant solution for selecting the best service per request.
```
mkolesnik: It's not really clear how this would work though, when do you remove and add items to the queue? How do you make sure that you're not bombarding a single service while starving the others? What about connection stickiness/tracking (eg TCP needs to reach the same service each time)?

lighthouse/multi-cluster-service-selection.md
mangelajo: May be we should describe what we consider the boundaries of service mesh vs "network metrics & reachability" somewhere in the document, to make it clear.

lighthouse/multi-cluster-service-selection.md
mangelajo: For congestion we could use available bandwidth vs used bandwidth in % or something like that. But really hard to measure over clusters. I'd leave that as a suggestion for future exploration, I agree with mike it's probably not doable, or not easy (unless you have full control of all the intermediate network infrastructure).

lighthouse/multi-cluster-service-selection.md
mangelajo: Right, 1 =(almost)= 2 , If the paths are symmetrical.

May be we just want to consider RTT ?

lighthouse/multi-cluster-service-selection.md
mangelajo: 0. Tunnel status (we may want to implement something like BFD on the gateways, on top of IPSEC/cabledrivers. In such way we get a very early signal (seconds) that something is going wrong with the connectivity. We could revive [1]

[1] https://github.com/jthurman42/go-bfd

lighthouse/multi-cluster-service-selection.md
mangelajo: Not sure where my comment went, but yes on a symmetrical connection.

I'd use just RTT / ping.

Although latency could be important on streaming cases (1 direction video stream for example), may be it's something we could move to a "futures" section.

danibachar:
@mangelajo @mkolesnik 
Would love your second overview of the suggested solution.
I split the solution into 3 stages (the first one is already implemented Round-Robin), the second Weighted-Round-Robin should be simple and does not require any API changes - just a way to collect the relevant metrics...
The third one might look a bit scary - specifically the fact that we would like to "know" the load in the system which can be hard to predict. But please let's discuss it - as the simulations of it seem promising...

lighthouse/service-selection/multi-cluster-service-selection.md
```diff
@@ -0,0 +1,525 @@
+# Multi-cluster service selection optimization - EP
+
+## Summary
+
+This EP, propose a 2 stage development process for improving the
+load balancing of service-to-service communication
```
mkolesnik:
```suggestion
load balancing of pod-to-service communication
```

technically this is the correct nomenclature as it's always pods that are initiating the connection

lighthouse/service-selection/multi-cluster-service-selection.md
```diff
@@ -0,0 +1,525 @@
...
+## Proposal
+
+### User Stories
+
+#### Operator can configure load balancing techniques
+
+1. Round-Robin:
+
+The distribution of traffic (i.e request) among Service-Imports of
+the same service type will be equal in expectation.
```
mkolesnik:
```suggestion
   The distribution of traffic (i.e request) among Service-Imports of
   the same service type will be equal in expectation.
```

Otherwise all these items render as 1.

lighthouse/service-selection/multi-cluster-service-selection.md
```diff
@@ -0,0 +1,525 @@
...
+distribute the load between `c2` and `c3` for `service-1`.
+in expectation over time a `1/3` (33%) of the traffic will
+be passed to `c3` and `2/3` (67%) will be passed to `c2` -
+inversed of their weight.
+Note about `zero weight` (`weight=0`). We have already
+implemented a mechanism that prefers local cluster service
+over remote ones, using `lower` weight -> `higher`
+the probability we will make a safety mechanism to return the
+cluster with the `weight=0` always probability will equal = 1.
+For safety, we will not allow more than 1 cluster to have `weight=0`.
```
mkolesnik:
```suggestion
   Each Service-Import will be assigned with a weight according to predefined
   parameters (e.g latency, RTT, pricing, etc), the `lower`
   the weight the `higher` the probability to `select` that
   specific service, in such way in expectation the traffic
   will distribute according to the weight.
   For example we have a system with 3 clusters (c1,c2,c3)
   and 2 exports for `service-1` from 2 clusters (c2,c3) in
   each cluster we will assign an appropriate weight for each
   export. An option will be `c1 {c2: 10, c3: 5}` (note that
   c1 does not have a local version of `service-1`),
   `c2 {c2: 0, c3: 4}`, `c3 {c2: 4, c3: 0}`.
   Now, let's take cluster `c1` for example that needs to
   distribute the load between `c2` and `c3` for `service-1`.
   in expectation over time a `1/3` (33%) of the traffic will
   be passed to `c3` and `2/3` (67%) will be passed to `c2` -
   inversed of their weight.
   Note about `zero weight` (`weight=0`). We have already
   implemented a mechanism that prefers local cluster service
   over remote ones, using `lower` weight -> `higher`
   the probability we will make a safety mechanism to return the
   cluster with the `weight=0` always probability will equal = 1.
   For safety, we will not allow more than 1 cluster to have `weight=0`.
```

lighthouse/service-selection/multi-cluster-service-selection.md
```diff
@@ -0,0 +1,525 @@
...
+
+1. Optimize:
+
+The system will calculate a smart weight for Service-Imports of
+the same kind in a way that will minimize some `cost` (will be
+explain shortly) the operator defines.
+From there the traffic (i.e request) distribution will work in
+like the Weighted-Round-Robin fashion.
```
mkolesnik:
```suggestion
   The system will calculate a smart weight for Service-Imports of
   the same kind in a way that will minimize some `cost` (will be
   explain shortly) the operator defines.
   From there the traffic (i.e request) distribution will work in
   like the Weighted-Round-Robin fashion.
```

mkolesnik (commented):
Some styling comments

danibachar:
@mkolesnik @mangelajo - fixed PR and rebased against devel - can we merge?

lighthouse/service-selection/multi-cluster-service-selection.md
```diff
@@ -0,0 +1,513 @@
...
+       issues. There are numerous models and
+       distributions for generating such disruptions.
+        1. Testing Tool - We plan to use [Pumba](https://github.com/alexei-led/pumba)
+           which is a chaos testing tool for Docker containers. It allows
+           performing network emulation and application disruptions.
+    1. Pricing:
+       One of the key objectives in this EP is to
+       improve the price one pays for service to service
+       communication. While measuring the actual pricing is a
+       complicated function to build, of different variables such
```
aswinsuryan, line 373: We can make use of topology keys as well to determine the pricing.

]https://kubernetes.io/docs/concepts/services-networking/service-topology/

lighthouse/service-selection/multi-cluster-service-selection.md
```diff
@@ -0,0 +1,513 @@
...
+           parameter in the future.
+
+    1. Functions - we might want to consider and test different
+       functions for combining the different parameters in the function
+
+### Challenges
+
+- Scalability to 5G – thousands of clusters
+- Autoscale utilization
+- Multi-hop
```
aswinsuryan, line 513: Handling sessions and DNS caching may create challenges as well.

lighthouse/service-selection/multi-cluster-service-selection.md
danibachar: Yea, this is a great idea! I will add it to the proposal!

lighthouse/service-selection/multi-cluster-service-selection.md
danibachar: True! Thanks again - will update

dfarrell07:
Is this still planned to be worked on?

maayanf24:
This has been discussed on the issue scrub meeting and decided to close for now.
If this will become a priority it can be reopened

PR #3, opened 2020-07-02
Clarify details about subctl show versions

mangelajo:
Currently the versions are stored in a single field on the operator,
and that's replicated for submariner-* and lighthouse-* (service-discovery).


mkolesnik (dismissed):
There's a further mention you need to change at the end of the document

PR #4, opened 2020-07-24
Support fine-grained cluster connectivity and multiple cable drivers

tpantelis:


submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,70 @@
...
+    id: east
+  rightCluster: 
+    id: west
+  cableDriver: 
+    name: libreswan
+    options:
+      encrypt: true
+```
+
+The `leftCluster` and `rightCluster` fields identify the two clusters. The simplest identifier is the cluster's unique
```
sridhargaddam: So @tpantelis , in a setup where we have two Clusters in OnPrem and one in public Cloud. Do we have to create multiple ClusterLinkConfig CRDs on **each cluster**? 

And is it mandatory to create the CRDs?

If the CRD is not configured, is the assumption that encryption would be used?

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,70 @@
...
+ID or name within the supercluster. This could be extended in the future to allow for a label selector to identify one 
+or more Submariner `Cluster` resources.
+
+While semantically the identifiers for the two clusters uniquely identify`ClusterLinkConfig` instances, the
+`leftCluster` and `rightCluster` field values could be reversed by the user in which case two instances semantically
```
sridhargaddam: Just thinking out loud for other ideas, what is your opinion on the following approach?

We associate each cluster with a new parameter which specifies if encryption is mandatory or optional?
This can go as part of ClusterCRD configuration itself.
An OnPrem cluster can be configured with "encryption: optional", whereas a public cluster will be configured with "encryption: required".
If an OnPrem1 cluster is trying to connect to OnPrem2 cluster, since both of them have Optional encyption, they can use non-encrypted tunnels.
However, if one of the sides have "encryption: required", then encrypted tunnels would be used.
This looks a bit simple, but I'd like to hear views on this.

CC: @nyechiel @skitt @mangelajo @mkolesnik @vthapar 

Screenshot from 2020-07-28 14-07-17

submariner/multiple-cable-drivers.md
skitt: I think we should have a set of defaults, same as currently (strongSwan, ultimately Libreswan, with encryption by default), and only override that if a link CR is present. The link CRs could be created on the broker and synced to the appropriate clusters.

sridhargaddam: Yes, it makes sense to assume encryption by default and override the configuration for specific links using the CRD.

submariner/multiple-cable-drivers.md
sridhargaddam: Eventhough this is simple, we will have less flexibility in certain scenarios. For example, if we have two public clusters (say in the same Region) and we want to avoid encryption between the two clusters, it would not be possible in this approach as we would have configured "Encryption: Required" for both of them. The proposal from @tpantelis would be more flexible.

skitt: Yes, this is really a link property, not something we can derive from cluster properties. At some point we might even need to be able to support multiple links between clouds, and some might be encrypted and others not; this is just an example, but it’s a typical HA requirement in real-world deployments.

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,70 @@
...
+    id: west
+  cableDriver: 
+    name: libreswan
+    options:
+      encrypt: true
+```
+
+The `leftCluster` and `rightCluster` fields identify the two clusters. The simplest identifier is the cluster's unique
+ID or name within the supercluster. This could be extended in the future to allow for a label selector to identify one 
+or more Submariner `Cluster` resources.
```
mkolesnik: So, using a selector makes a lot of sense and goes more in the direction of "connectivity policy" that we discussed, but then there's a bunch of other problems stemming (such as what happens if 2 conflicting policies apply at the same time).

Perhaps we should focus initially on links between 2 clusters?

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,70 @@
...
+There needs to be a way for a user to specify their intent for two clusters to use a specific cable driver
+implementation and/or configuration. At a minimum, the intent must include identifying information for the two
+clusters and the cable driver implementation.
+
+The standard mechanism in Kubernetes to specify intent is by creating a Kubernetes API resource whose effect is eventually
+realized into the cluster's desired state. To extend the Kubernetes API, one could either use the generic Kubernetes
+`ConfigMap` or use a custom resource definition (CRD). There's pros and cons to each but a CRD seems more suitable for
+this enhancement.
+
+This following CRD is proposed:
```
mkolesnik, line 37: So this really seems to me like a definition of a Tunnel, maybe we can simply name the CR that?

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,70 @@
...
+appVersion: v1
+kind: ClusterLinkConfig
+metadata:
+  name: east-west-link
+spec:
+  leftCluster: 
+    id: east
+  rightCluster: 
+    id: west
+  cableDriver:
```
mkolesnik: How does this interact with the Endpoint's Backend and BackendConfig?

submariner/multiple-cable-drivers.md
mkolesnik: @sridhargaddam your proposal is interesting but limited as you and @skitt mentioned.

There is a nice simplicity though to it, so I guess the question is how much flexibility we want to allow. For example, would we even want to allow a supercluster with a mix of cable drivers (libreswan for some connections, wireguard for others)?

Another possibility is to combine the two approaches, but I guess using only one mechanism to control the tunnels makes most sense.

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,70 @@
...
+This could be achieved using different cable driver implementations or the same cable driver implementation with
+differing encryption configurations.
+
+## Proposal
+
+### User-facing API
+
+There needs to be a way for a user to specify their intent for two clusters to use a specific cable driver
+implementation and/or configuration. At a minimum, the intent must include identifying information for the two
+clusters and the cable driver implementation.
```
mkolesnik: So generally is this just a "recommendation" to the submariner engine on how to create the tunnels?

What happens if you specify a cluster ID that doesnt exist?
Also what happens to a tunnel that was already established and then this object is created, would it be destroyed and re-created per the requested intention?

submariner/multiple-cable-drivers.md
tpantelis: > I think we should have a set of defaults, same as currently (strongSwan, ultimately Libreswan, with encryption by default), and only override that if a link CR is present. 

yes - I was thinking the default driver  + config as we have now that is passed via the env would remain the same (at least initially).  

> The link CRs could be created on the broker and synced to the appropriate clusters.

yes. This is a rough first draft - I didn't mention any of these details yet.

submariner/multiple-cable-drivers.md
tpantelis: The intention would be to possibly add selectors later.

submariner/multiple-cable-drivers.md
tpantelis: Whatever semantics we deem appropriate...

submariner/multiple-cable-drivers.md
tpantelis: It wouldn't - TBH I'm not really clear on what the intention of this is/was.

submariner/multiple-cable-drivers.md
tpantelis: > What happens if you specify a cluster ID that doesnt exist?

Implementation detail. This CRD signifies user intent which means they could specify the ID of a  cluster that hasn't been joined yet.  This is similar to exporting a service that hasn't been create yet. If we add a `subctl` command for this, we could check for this case.

> Also what happens to a tunnel that was already established and then this object is created, would it be destroyed and re-created per the requested intention? In either case, if the user made a typo ... we can't distinguish that.

I would imagine so. Again,  implementation detail.

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,70 @@
...
+
+This following CRD is proposed:
+
+```
+appVersion: v1
+kind: ClusterLinkConfig
+metadata:
+  name: east-west-link
+spec:
+  leftCluster:
```
mangelajo: I'd suggest we use standard kubernetes selectors

https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements

This would require that we add labes to the clusters, but it would allow a richer syntax without the need to make 1:1 link configs.

submariner/multiple-cable-drivers.md
mangelajo: I think is not that big of a change, I'd define them from the start.

We can start by just adding in submariner-engine a label to the cluster: id == cluster-id  , and later on we can focus on adding custom labels (via config in the operator...)

submariner/multiple-cable-drivers.md
mangelajo: This would eventually be reflected in the Endpoint cabledriver & backend config. The options may be interpreted by the cable driver, IMO "options" should be a free k/v dictionary.

submariner/multiple-cable-drivers.md
mangelajo: We could resolve this by adding a priority field in the link configs....

The default could be created with priority 100, ... but if we add higher priority ones "1..", they would override  (or inverse numbering, whatever makes more sense)

submariner/multiple-cable-drivers.md
mangelajo: Let's try to define as much of a stable API as we can, as the product matures, changes like that would be hurtful later (we will need to take care of more migration patterns.)

mangelajo (changes requested):
Some comments about using selectors, and perhaps define priorities?

mangelajo:
Sorry for the super late comments I know this is probably not even a priority now, but wanted to share my views.

tpantelis:
I updated the proposal to include connectivity policy.

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,132 @@
...
+
+The standard mechanism in Kubernetes to specify intent is by creating a Kubernetes API resource whose effect is eventually
+realized into the cluster's desired state. To extend the Kubernetes API, one could either use the generic Kubernetes
+`ConfigMap` or use a custom resource definition (CRD). There's pros and cons to each but a CRD seems more suitable for
+this enhancement.
+
+This following CRD is proposed:
+
+```Go
+type ClusterConnectivityPolicies struct {
```
aswinsuryan: Would it be better to have ClusterConnectivityPolicies and ClusterConnectivityPoliciesBinding? So that we can reuse policies. The former defining policies and the latter deciding which clusters it should apply to

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,132 @@
...
+    Options map[string]string `json:"options,omitempty"`
+}
+```
+
+The `ClusterConnectivityPolicies` CRD defines a list of policies that are applied in order of precedence. To resolve the
+connectivity policy for a remote cluster's `Endpoint`, Submariner would iterate the list of policies to search for a policy
+whose left and right label selectors match the corresponding `Cluster` resources of the local and remote `Endpoints`. The
+first matching policy is used. If no matching policy is found, the clusters are not connected.
+
+For full mesh connectivity, a policy with empty left and right label selectors would match all clusters. This would be the
```
aswinsuryan: Should we be providing the option to decide topology per connection? Wouldn't be more realistic to have a uniform topology within a clusterset. Like a clusterset with full-mesh , another one with hub and spoke.

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,132 @@
...
+scalability where  full mesh is not required.
+
+Submariner supports several cable driver implementations (libreswan, wireguard et al) however only one can be
+configured within the same cluster set, that is all clusters that join a broker must use the same cable driver.
+
+There are use cases that involve supporting a mix of cable drivers or cable driver configurations within the same
+cluster set. For example, a cluster set could contain three clusters, two that are on-premise and the other
+in the cloud. Connections to the cloud cluster will need to use encryption while the on-premise clusters can use
+unencrypted connections, as they’re connected by a private link, to avoid the performance overhead of encryption.
+This could be achieved using different cable driver implementations or the same cable driver implementation with
```
mkolesnik, line 17:
```suggestion
This could be achieved using different cable driver implementations of the same cable driver implementation with
```

submariner/multiple-cable-drivers.md
mkolesnik: I tend to agree with @aswinsuryan, clusterset topology is a whole other issue with it's own implications and considerations which we can discuss either as a full equal partner of this proposal, or an entirely separate one.

To me this proposal focuses on the connectivity types and not the topology, and hence the default should apply in case no other policy is found.

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,132 @@
...
+}
+```
+
+The `ClusterConnectivityPolicies` CRD defines a list of policies that are applied in order of precedence. To resolve the
+connectivity policy for a remote cluster's `Endpoint`, Submariner would iterate the list of policies to search for a policy
+whose left and right label selectors match the corresponding `Cluster` resources of the local and remote `Endpoints`. The
+first matching policy is used. If no matching policy is found, the clusters are not connected.
+
+For full mesh connectivity, a policy with empty left and right label selectors would match all clusters. This would be the
+default.
```
mkolesnik: What if this policy doesn't exist?
I think it this case we should fall back to a hard coded (documented) default.

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,132 @@
...
+
+For each connectivity policy, the cable driver and associated configuration options can optionally be specified. If none
+is specified, the default cable driver is used.
+
+The `ClusterConnectivityPolicies` resource would be a singleton with a well-known name and would be maintained on the
```
mkolesnik: Not sure I understand what this means, why does it need to be a singleton with a constant name?
It seems to be quite unlike anything else in K8s..

submariner/multiple-cable-drivers.md
tpantelis: A topology can be realized in this manner. The last example below is hub-and-spoke. One could also define multiple hub-and-spokes.

submariner/multiple-cable-drivers.md
tpantelis: The idea is that there's only one `ClusterConnectivityPolicies` per cluster set so there's no need for reuse.

submariner/multiple-cable-drivers.md
tpantelis: That said, we could instead allow it to be defined in terms of specific topologies, eg

```
spec:
  policies:
     - hub-and-spoke
         hubSelector:
             database-role: server
         spokeSelector:
             database-role: client
      - full-mesh
```

I'm not sure what other topologies there could be. Also we want to be able to satisfy the use case for specifying the cable driver for a set of connections (as in the second example below).  The proposed API here makes it flexible to do that.

aswinsuryan: By clusterset I meant  a group of clusters as defined in [1]. So if we are planning to align with this, I think it would be easier to have uniform topology within a clusterset as it will be easier to program routes.

[1]https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/2149-clusterid

mkolesnik (changes requested):

submariner/multiple-cable-drivers.md
sridhargaddam: I think @tpantelis is trying to say that we can use VxLan cable driver for certain connections and Libreswan for other connections.

submariner/multiple-cable-drivers.md
tpantelis: > Not sure I understand what this means, why does it need to be a singleton with a constant name?
> It seems to be quite unlike anything else in K8s..

The `ClusterConnectivityPolicies` defines all the policies in precedent order so there only needs to be one in a cluster set.

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,132 @@
...
+
+```yaml
+appVersion: v1
+kind: ClusterConnectivityPolicies
+metadata:
+  name: cluster-connectivity-policies
+spec:
+  policies:
+    - leftClusterSelector:
+          location: on-premise
```
sridhargaddam: Are the labels validated against `clustercrd`?

submariner/multiple-cable-drivers.md
mkolesnik: ah right, i misread it

submariner/multiple-cable-drivers.md
tpantelis: yes - the submariner `Cluster` resource. For now anyway - when ClusterClaim becomes a reality then I imaging we'd switch to that.

submariner/multiple-cable-drivers.md
mkolesnik: Why then not just allow multiple such policies to exist?

If multiple policies match the same two clusters (or endpoints?) we can then issue a warning and select arbitrarily one of them (say based on creation date).

submariner/multiple-cable-drivers.md
tpantelis: Of course we can provide a default if no policies are defined.

submariner/multiple-cable-drivers.md
tpantelis: BTW there is precedence for resources with a well-known name: the submariner `Cluster` and the proposed `ClusterClaim`.

tpantelis: > Why then not just allow multiple such policies to exist?
> 
> If multiple policies match the same two clusters (or endpoints?) we can then issue a warning and select arbitrarily one of them (say based on creation date).

We could - that was the original proposal. I'll reverse your question - why not do it this way to avoid those complexities?

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,132 @@
...
+
+The `ClusterConnectivityPolicies` resource would be a singleton with a well-known name and would be maintained on the
+broker cluster and synced to each participating cluster.
+
+### Examples
+
+#### Full mesh (default)
+
+```yaml
+appVersion: v1
```
sridhargaddam: apiVersion?

ditto: in the below examples.

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,132 @@
...
+```Go
+type ClusterConnectivityPolicies struct {
+    metav1.TypeMeta `json:",inline"`
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+
+    Spec ClusterConnectivityPoliciesSpec `json:"spec"`
+}
+
+type ClusterConnectivityPoliciesSpec struct {
+    Policies []ClusterConnectivityPolicy `json:"policies,omitempty"`
```
mangelajo: I believe it's more standard in k8s to have separate policies, instead of a big one with all policies.

As you explained this is to be specific about priorities order, and we talked about not using prioirty numbers (magic numbers), and may be another object where we can provide the order of policies when that needs to be resolved. I guess the spec wasn't still updated since last meeting.

mangelajo (changes requested):

aswinsuryan (approved):

submariner/multiple-cable-drivers.md
bergeramit: A good solution for these concerns can be:
1. separate policies, without a big list of policies, while enforcing a unique naming scheme for policies
2. add a PolicyOrder element that will list the policies by name from the most preferred to the least preferred, here we can also specify the "default" policy (for example - this can include the default cable driver to use).  Although YAML does not support ordered lists directly we can easily use a dictionary for this task - which will also make this added order policy more readable.

The clear benefits of this approach:
1. we can easily figure out the exact order of priorities and scan to understand which cluster set uses which policy
2. we achieve flexibility within the YAML
3. we can generate complex policies with modular and clean policy definitions  - for example: the policy of "clients can access the server but not each other" can be written *below* (in the PolicyOrder) the policy of "Admin clients in each cluster set can speak to each other with a secured connection" - This generates the wanted policy "clients should only talk with the server unless they are Admin clients who can talk to each other but only in a secure manner" in a modular, clean and easy way.

One downside to this approach is that we will not have a single place that specify the entire policy rather a scattered partially written rules. An approach to handle this is to issue a log/ print/ unique config file with the final policies together - every connection and its specs.

tpantelis: There's pros and cons with both approaches. I'm fine with adding a `ClusterConnectivityPolicyOrder` resource. One downside with it is that a referenced policy in the list can become stale if it's deleted and not removed from the list by the user but perhaps we can do that automatically.

mangelajo: I like the cluster connectivity policy order approach. Provides more flexibility, although it has the cons of more computation and being less atomic.

mangelajo: But honestly I’m also good with the current single object approach. 

The advantage of splitting is if we believe that this policy could also be targeted by multiple workers creating or modifying policies in real time (less
contention over a single object) but I cannot envision such case.

so the current form is possibly good.

mangelajo (approved):

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,192 @@
+# Support fine-grained cluster connectivity and multiple cable drivers
```
mkolesnik, line 1: I'm really not sure why these two different proposals are being bundled together.

The ability to support different topologies (with or without connectivity between participating clusters) is not directly related to being able to specify different cable drivers between any two connected clusters.

I suggest these two be split out and each focus on their own problem domain.

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,192 @@
...
+```Go
+type ClusterConnectionPolicies struct {
+    metav1.TypeMeta `json:",inline"`
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+
+    Spec ClusterConnectionPoliciesSpec `json:"spec"`
+}
+
+type ClusterConnectionPoliciesSpec struct {
+    Policies []ClusterConnectionPolicy `json:"policies,omitempty"`
```
mkolesnik, line 112: Having a singleton resource with a list seems like a bad API going forward, we can solve any conflicts between competing policies with other mechanisms such as a ConfigMap that specifies priority order for cable drivers (with fall back to hard coded default if it's empty or unspecified), or giving the policies a weight such that two competing policies are judged by it and the higher value one is selected.

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,192 @@
...
+and associated optional configuration options to use for each cluster connection. To resolve the connection policy for a remote
+cluster's `Endpoint`, Submariner would iterate the list of policies to search for a policy whose left and right label selectors
+match the corresponding `Cluster` resources of the local and remote `Endpoints`. The first matching policy is used. If no matching
+policy is found, the default cable driver is used.
+
+Defining the policies as an ordered list allows the user to resolve ambiguity should a cluster pair match multiple policies.
+Otherwise, if Submariner encountered such a case, it would either have to arbitrarily pick a policy or reject all of them and
+use the default. Neither option is ideal. An alternative to the ordered list would be to have separate resources for each policy
+and define a priority or precedence field. However it seems a single resource without the use of magic priority numbers would be
+easier for the user to maintain.
```
mkolesnik, line 146: I disagree with this premise, using weights won't have any tangible disadvantage versus using a singleton resource.

Using the ordered list doesn't give any better clarity as to which policies affect what since you're lacking the selector part when viewing it, and as an admin would have to query the individual policies for that data anyhow.

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,192 @@
...
+
+The `ClusterConnectionPolicies` CR would be a singleton with a well-known name and would be maintained on the broker cluster
+and synced to each participating cluster.
+
```
mkolesnik, line 150: I think it would be good for debugging purposes if the status field of the `Endpoint` reflected which policy is in effect

submariner/multiple-cable-drivers.md
tpantelis: I assume a singleton ConfigMap? I assume that would have a list in priority order? How would that different than this?

Another approach is using magic weight/priority numbers assigned to each policy by the user. That has disadvantages as outlined elsewhere in this PR.

submariner/multiple-cable-drivers.md
tpantelis: I guess it's a matter of opinion/viewpoint. There's pros and cons to both approaches - we would just need to pick one.

submariner/multiple-cable-drivers.md
bergeramit: I agree. Moreover we should consider adding a way to get statuses of all endpoints or alternatively get a complete view on compatible or not compatible cluster's communication's policy. This can maybe be logged instead of a full feature.

This can be a good debug tool as well

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,192 @@
...
+Submariner currently assumes full mesh by default and provisions VPN connections between all participating clusters. For
+some use cases, it makes sense to offer finer-grained connectivity options. One such use case is a client-server scenario
+where multiple client clusters need to access a server cluster but the user does not want the client clusters to connect
+to one another. Finer-grained connectivity could also improve overall scalability where full mesh is not required.
+
+Submariner supports several cable driver implementations (libreswan, wireguard et al) however only one can be
+configured within the same cluster set, that is all clusters that join a broker must use the same cable driver.
+
+There are use cases that involve supporting a mix of cable drivers or cable driver configurations within the same
```
maryamtahhan, line 13: Just a nit (I'm sorry) can we reserve the "mixing" of cable drivers for the situation where we combine them like VXLAN over IPsec...

```suggestion
There are use cases that involve supporting multiple cable drivers or cable driver configurations within the same
```

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,192 @@
...
+    // RightClusterSelector identifies the Cluster resources representing the clusters on the other end of a connection.
+    // An empty selector indicates wildcard, ie matches any cluster.
+    RightClusterSelector metav1.LabelSelector `json:"rightClusterSelector,omitempty"`
+}
+```
+
+The `ClusterConnectivityPolicy` CR defines the criteria for selecting which cluster pairs can connect. To resolve the
+connectivity policy for a remote cluster's `Endpoint`, Submariner would search for a policy whose left and right label
+selectors match the corresponding `Cluster` resources of the local and remote `Endpoints`. If any matching policy is found,
+the clusters are connected, otherwise they are not.
```
maryamtahhan, line 61: So silly questions, will gateways respond to changes in the policies? so for example if the selectors for an existing policy modified/extended what happens? or do we just assume this is a new policy...

I guess it's not entirely clear to me what happens to the clusters (more specifically what they connect to) if no matching policies are found - is this an error/should they fall back to a default... In other words is it implicit that any clusters that don't meet the an defined policies will fall back to full mesh connectivity with any other clusters that don't meet the selector policy

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,192 @@
...
+
+If there are no policies defined, Submariner would default to full mesh connectivity.
```
maryamtahhan, line 63: Do you still need to define a default cable driver for full mesh connectivity somewhere? or do we assume that full mesh connectivity and multiple cable drivers are mutually exclusive?

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,192 @@
...
+
+    CableDriver *CableDriverSpec `json:"cableDriver,omitempty"`
+}
+
+type CableDriverSpec struct {
+    // Name of the cable driver implementation.
+    Name string `json:"name"`
+
+    // Options specifies key/value configuration parameters for the cable driver.
+    Options map[string]string `json:"options,omitempty"`
```
maryamtahhan, line 132: Currently I am looking more at the mixing of cable drivers rather than the enablement + advertising of multiple cable drivers, but my thinking was to go down the route of using ConfigMaps... the reason being that every time I wanted to extend/modify the Custom Resource Definition I would need to release a new definition, as well as being able to override certain params if needed locally and being able to prioritise them base on the order they are loaded... 

it would be good to get alignment on best practice for Submariner as regards configuration parameters for the cable driver work (both multiple interface enablement as well as the mixing/combining of backends):

I was thinking that configuration parameters fall into two categories based on their scope:
Pod Local - these are parameters that hold configuration parameters specific to each Pod (values vary from Pod to Pod). 
Namespace Local - these are configuration parameters that you would like to share across a namespace (in this case the submariner-operator namespace). 

I was originally thinking that something like the following would be useful:
- To use environment variables/secrets injected into the Gateway Pod where a parameter will have a varying value in each Pod. (involves extending the submariner.io_submariners.yaml for some parameters)... 
 - To use a mounted ConfigMap/secrets for configurations that need to be shared across pods.
But it would be good to indicate if there is a preference for something else in terms of cable driver  configurations... 

I guess not all configuration parameters are created equally - so some rough guidelines would be useful...

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,192 @@
...
+
+The `ClusterConnectionPolicies` CR defines a list of policies that are applied in order of precedence to select the cable driver
+and associated optional configuration options to use for each cluster connection. To resolve the connection policy for a remote
+cluster's `Endpoint`, Submariner would iterate the list of policies to search for a policy whose left and right label selectors
+match the corresponding `Cluster` resources of the local and remote `Endpoints`. The first matching policy is used. If no matching
+policy is found, the default cable driver is used.
+
+Defining the policies as an ordered list allows the user to resolve ambiguity should a cluster pair match multiple policies.
+Otherwise, if Submariner encountered such a case, it would either have to arbitrarily pick a policy or reject all of them and
+use the default. Neither option is ideal. An alternative to the ordered list would be to have separate resources for each policy
```
maryamtahhan, line 144: Would it be worth building in an option to load balance/round robin across multiple matching policies/backends (if no priority is defined or if the priority for those backends is equal) as well as being able to prioritise the policies?

maryamtahhan (commented):
@tpantelis Sorry about all the questions - just want to make sure I'm fully understanding the proposal...

submariner/multiple-cable-drivers.md
maryamtahhan: actually NVM this round robin suggestion - this would probably be a pain in reality in terms of the co-ordination it would require from both ends of a connection

submariner/multiple-cable-drivers.md
tpantelis: Sorry not sure I follow.

submariner/multiple-cable-drivers.md
tpantelis: It's been quite a while since I looked at this - trying to re-familiarize...  If there's no matching policy for a particular remote Endpoint, it won't create a connection. Does that answer your question?

submariner/multiple-cable-drivers.md
tpantelis: The connectivity policy concept outlined here is separate from the cable driver specification concept. The  connectivity policy says which clusters can connect to one another and the cable driver spec says how they connect.

submariner/multiple-cable-drivers.md
tpantelis: > Currently I am looking more at the mixing of cable drivers rather than the enablement + advertising of multiple cable drivers, but my thinking was to go down the route of using ConfigMaps... the reason being that every time I wanted to extend/modify the Custom Resource Definition I would need to release a new definition, 

Not necessarily - with a CRD you can easily make changes in a backwards compatible manner - we've done that several times. Regardless, you'd have the same backwards compatibility considerations using a `ConfigMap`. There's pros and cons to using `ConfigMap` vs CRD. A `ConfigMap` is suitable for simple key/value params but this is more complex than that I think.

> as well as being able to override certain params if needed locally and being able to prioritise them base on the order they are loaded...

How does a CRD preclude that?

submariner/multiple-cable-drivers.md
maryamtahhan: Ok so then if you are using the fine grained cluster connectivity it's up to you to make sure that there are no dangling clusters - in other words you need to define enough policies to make sure you avoid the situtation, there are no preset defaults

submariner/multiple-cable-drivers.md
maryamtahhan: Yeah I appreciate what you are looking at with the policy structures is more complex than Key value pair params as such the CRD is more appropriate. I was specifically referring to using a ConfigMap for the map parameter for the Cable driver structure. Just trying to make sure we are aligned on that... but I can work with a map...

submariner/multiple-cable-drivers.md
tpantelis: By dangling cluster I assume you mean one that isn't included in a policy. Since the policies are a whitelist then such a cluster is naturally excluded from connectivity. The preset default is full mesh connectivity. Being a whitelist, technically the absence of any policy should mean no connectivity but we can imply a wildcard policy out-of-the-box even if we don't explicitly create it.

The other alternative is to use a blacklist approach but that doesn't seem suitable.

submariner/multiple-cable-drivers.md
donaldh: On the one hand, this could be separated into two proposals. But they share a common foundation which is defining the attributes of cluster adjacencies. My suggestion would actually be to align them more closely and use the same policy CRD for defining adjacencies and their cable properties.

submariner/multiple-cable-drivers.md
maryamtahhan: yeah that's what I meant by dangling cluster...  what got me thinking about it was if you had 7 clusters and 5 of those matched the policy (you definitely have 5 connected clusters) but the question I had then was regarding the 2 remaining clusters, are they automatically connected...
But I think there might be situations where maybe folks don't want to automatically connect those clusters - so probably better to let the end user create a wild card policy to fall back on...

submariner/multiple-cable-drivers.md
tpantelis: I think there was an original iteration where they were combined but then we decided to separate them. There are different use cases for the what vs the how.

submariner/multiple-cable-drivers.md
tpantelis: > But I think there might be situations where maybe folks don't want to automatically connect those clusters - so probably better to let the end user create a wild card policy to fall back on...

But then wouldn't they be back to automatic full-mesh and thus defeat the purpose of the whitelist policy?

So, as mentioned before, there's an implicit wildcard policy out-of-the-box - when the user creates the first actual policy, that also implicitly (logically) removes the wildcard policy.

maryamtahhan: Are we saying that choosing to fine tune the interconnects on some of the clusters precludes you from meshing the rest of the clusters in a cluster set?

I just assumed that the user could add a wild card policy as the lowest priority policy and so, any clusters that didn't match any existing policies would be meshed... (of course they can choose to do something else as the fall back policy)

so it comes down to what the use case is, is it 
1. to let someone fine tune the interconnects for all clusters in a cluster set in one go
2. to let someone fine tune the interconnects for some of the clusters in a cluster set at a time.
3. Both :)

I guess I'm not understanding the scope... :(

tpantelis: Referring back to the example use cases outlined above, the purpose is to allow the user to restrict which remote clusters a cluster can access, for security, scalability etc. Connectivity is a binary decision and thus a prioritized policy list with fallback is not suitable, as it is with the cable engine config. 

> Are we saying that choosing to fine tune the interconnects on some of the clusters precludes you from meshing the rest of the clusters in a cluster set?

Yes but I wouldn't say preclude - by defining policies the users is choosing specifically which clusters to connect.

Let's say a user defines policies A and B. The algorithm would cycle thru the policies - first check policy A; if the clusters match the selectors then connect them; otherwise check B. If neither match then they're not connected. Now if the user added a wildcard policy C then every cluster pair would automatically match and thus be connected. This wouldn't be desirable.

Perhaps there's a use case where a user has 50 clusters and wants all to interconnect except, say 3. A whitelist policy would be cumbersome so we could introduce blackbox policies as well to facilitate this scenario, in which case a whitelist wildcard policy would be relevant.

This is my understanding of the problem domain and how to solve it. Perhaps you have a different idea or something I'm missing.

donaldh: Sorry to reopen the the conversation, but migrating from no policy (full-mesh connectivity) to policy (no connectivity by default) seems problematic. If I have a cluster set of several clusters connected full-mesh that are sharing services, and I want to add a satellite cluster that will connect only to the geographically nearest of the existing clusters, I would want to define a policy for the new satellite cluster without affecting the existing mesh. There does not seem to be a way to get from one configuration to the other without affecting connectivity during the transition.

tpantelis: Yeah I'm sure there's various other use cases. It's open for ideas. What would you suggest? Perhaps adding blacklist policies?

donaldh: I was originally thinking about this from the perspective of selective service import and wondering how we might establish connectivity only where it is needed for the declared service consumption. That's what started me considering it from the perspective of a single cluster. I will explore whether we can define the deny/allow default on a per cluster basis and see if I can come up with something worth proposing.

tpantelis: > If I have a cluster set of several clusters connected full-mesh that are sharing services, and I want to add a satellite cluster that will connect only to the geographically nearest of the existing clusters

I think you could handle that scenario with a backlist policy. Say the satellite cluster is labeled with "_satellite_" and the other with "_east_", then create a backlist policy with label selectors **satellite** and **!east** - this would disallow _satellite_ connecting with any other cluster except for _east_. With no whitelist policies defined, it would default to the all-inclusive wild card policy which would allow everything else.

maayanf24:
This has been discussed on the issue scrub meeting and decided to close for now.
If this will become a priority it can be reopened

PR #6, opened 2020-07-28
Initial proposal for adding a built-in benchmarking tool

maayanf24:
Signed-off-by: Maayan Friedman <maafried@redhat.com>

### DIFF
--- a/issue-40-built-in-benchmarking-tool.md
+++ b/issue-40-built-in-benchmarking-tool.md
@@ -1,12 +1,12 @@
 ## Add Built-in Benchmarking Tool
 
-https://github.com/submariner-io/submariner-website/issues/40
+[submariner-operator#494](https://github.com/submariner-io/submariner-operator/issues/494)
 
 ## Summary
 
 This is a proposal to support a built-in benchmarking tool in Submariner.
-The goal is to provide users with easy way to benchmark the network performance across clusters, and experiment with different tuning of the system.
-
+The goal is to provide users with easy way to benchmark the network performance across clusters, and experiment with different tuning of
+the system.
 
 ## Proposal
 
@@ -17,41 +17,46 @@ Scheduling server and client can be placed within the same cluster or spread acr
 The user can specify source and destination nodes via node labels.
 
 If node labels are not specified, the following tests will be run by default:
+
 * non-gateway to non-gateway  
 * gateway to gateway
 
 In case server-options are not included, the command will use defaults.
 
-**Note** - In phase 1 of the implementations there will be no additional params other then `<kubeconfing1>` and 
+**Note** - In phase 1 of the implementations there will be no additional params other then `<kubeconfing1>` and
 `<kubeconfig2>` and only default tests will be performed.
 
-#### Backend tools and params
+### Backend tools and params
 
-The backend for latency test will be netperf.   
+The backend for latency test will be netperf.
 The default params that will be passed to netperf:
-```
+
+```console
 -t TCP_RR  -- -o min_latency,mean_latency,max_latency,stddev_latency,transaction_rate
 ```
+
 * -t: testname. TCP_RR - user-space to user-space ping with no connection time (request/response test)  
-* -o: sets the style of output to “CSV” where there will be one line of comma-separated values, preceded by one line of column names  
+* -o: sets the style of output to “CSV” where there will be one line of comma-separated values, preceded by one
+line of column names  
 
 The backend for throughput tests will be iperf. The default params that will be passed to iperf:
-```
+
+```console
 -w 8M -P 10 -c
 ```
-* -w: TCP window size `n[KM]`  
-* -P: The number of simultaneous connections to make to the server    
-* -c: client mode    
+
+* -w: TCP window size `n[KM]`
+* -P: The number of simultaneous connections to make to the server
+* -c: client mode
 
 #### Source code location
 
-Source code should go into `submariner-operator`.   
+Source code should go into `submariner-operator`.
 There is going to be a dependency on the nettest image from Shipyard, which will need to have the necessary tools.
-
 
 ### Subctl command
 
-```
+```console
 subctl benchmark throughput <kubeconfig1> <kubeconfig2>
 subctl benchmark latency <kubeconfig1> <kubeconfig2>
 ```
--- a/issue-40-built-in-benchmarking-tool.md
+++ b/issue-40-built-in-benchmarking-tool.md
@@ -1,0 +1,31 @@
+## Add Built-in Benchmarking Tool
+
+## Summary
+
+There is a need to support a built-in benchmarking tool in Submariner.
+
+## Proposal
+
+Add a new ability to have a built-in benchmark tool.
+Implement a new Subctl command that will use iperf3 internally to benchmark the throughput between the two nodes.
+
+The feature Should support scheduling server and client in the same cluster.
+
+The user can specify source and destination nodes via node names. (Or via lables?)
+
+If nodes are not specified, the following will be the default tests: 
+* non-gateway to non-gateway  
+* gateway to gateway
+
+
+#### Source code location
+
+Source code should go into Submariner. 
+There is going to be a dependency on Shipyard (iperf image)
+
+
+### Subctl command
+
+```
+subctl test throughput --server-config, --server-node, --client-config and --client-node
+```
--- a/issue-40-built-in-benchmarking-tool.md
+++ b/issue-40-built-in-benchmarking-tool.md
@@ -22,7 +22,8 @@ If node labels are not specified, the following tests will be run by default:
 
 In case server-options are not included, the command will use defaults.
 
-*Note* - In phase 1 of the implementations there will be no additional params other then `<kubeconfing1>` and `<kubeconfig2>`
+**Note** - In phase 1 of the implementations there will be no additional params other then `<kubeconfing1>` and 
+`<kubeconfig2>` and only default tests will be performed.
 
 #### Backend tools and params
 
@@ -44,13 +45,13 @@ The backend for throughput tests will be iperf. The default params that will be 
 
 #### Source code location
 
-Source code should go into Submariner.   
+Source code should go into `submariner-operator`.   
 There is going to be a dependency on the nettest image from Shipyard, which will need to have the necessary tools.
 
 
 ### Subctl command
 
 ```
-subctl test throughput <kubeconfig1> <kubeconfig2>
-subctl test latency <kubeconfig1> <kubeconfig2>
+subctl benchmark throughput <kubeconfig1> <kubeconfig2>
+subctl benchmark latency <kubeconfig1> <kubeconfig2>
 ```
--- a/issue-40-built-in-benchmarking-tool.md
+++ b/issue-40-built-in-benchmarking-tool.md
@@ -1,20 +1,23 @@
 ## Add Built-in Benchmarking Tool
 
+https://github.com/submariner-io/submariner-website/issues/40
+
 ## Summary
 
-There is a need to support a built-in benchmarking tool in Submariner.
-This is to provide users with easy way to benchmark the network performance within and across clusters, and experiment with different tuning of the system.
+This is a proposal to support a built-in benchmarking tool in Submariner.
+The goal is to provide users with easy way to benchmark the network performance across clusters, and experiment with different tuning of the system.
+
 
 ## Proposal
 
 Add a new ability to have a built-in benchmark tool.
-Implement a new Subctl command that will use iperf3 internally to benchmark the throughput between the two nodes.
+Implement a new Subctl command that will use iperf3 internally to benchmark the throughput between two nodes.
 
 scheduling server and client can be placed within the same cluster or spread across multiple clusters.
 
 The user can specify source and destination nodes via node names. (Or via lables?)
 
-If nodes are not specified, the following will be the default tests: 
+If nodes are not specified, the following tests will be run by default:
 * non-gateway to non-gateway  
 * gateway to gateway
 
@@ -22,7 +25,7 @@ If nodes are not specified, the following will be the default tests:
 #### Source code location
 
 Source code should go into Submariner. 
-There is going to be a dependency on Shipyard (iperf image)
+There is going to be a dependency on the nettest image from Shipyard (Which has iperf).
 
 
 ### Subctl command
--- a/issue-40-built-in-benchmarking-tool.md
+++ b/issue-40-built-in-benchmarking-tool.md
@@ -11,16 +11,27 @@ The goal is to provide users with easy way to benchmark the network performance 
 ## Proposal
 
 Add a new ability to have a built-in benchmark tool.
-Implement a new Subctl command that will use iperf3 internally to benchmark the throughput between two nodes.
+Implement a new Subctl command that will use uperf internally to benchmark the throughput and round trip latency between two nodes.
 
-scheduling server and client can be placed within the same cluster or spread across multiple clusters.
+Scheduling server and client can be placed within the same cluster or spread across multiple clusters.
 
-The user can specify source and destination nodes via node names. (Or via lables?)
+The user can specify source and destination nodes via node labels.
 
-If nodes are not specified, the following tests will be run by default:
+If node labels are not specified, the following tests will be run by default:
 * non-gateway to non-gateway  
 * gateway to gateway
 
+In case server-options are not included, the command will use defaults.
+
+The backend for latency test will be netperf. The default params that will be passed to netperf:
+```
+-t TCP_RR  -- -o min_latency,mean_latency,max_latency,stddev_latency,transaction_rate
+```
+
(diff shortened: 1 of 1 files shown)


issue-40-built-in-benchmarking-tool.md
```diff
@@ -0,0 +1,31 @@
...
+#### Source code location
+
+Source code should go into Submariner. 
+There is going to be a dependency on Shipyard (iperf image)
+
+
+### Subctl command
+
+```
+subctl test throughput --server-config, --server-node, --client-config and --client-node
```
sridhargaddam, line 30: Please include kubeconfigs for both the clusters.

skitt: `--server-config` and `--client-config` specify the kubeconfigs for both clusters.

nyechiel: @skitt are there any parameters (e.g TCP Window) we need to expose at the "subctl test throughput" level or are we assuming that everything goes into the config files?

issue-40-built-in-benchmarking-tool.md
```diff
@@ -0,0 +1,31 @@
...
+## Summary
+
+There is a need to support a built-in benchmarking tool in Submariner.
+
+## Proposal
+
+Add a new ability to have a built-in benchmark tool.
+Implement a new Subctl command that will use iperf3 internally to benchmark the throughput between the two nodes.
+
+The feature Should support scheduling server and client in the same cluster.
```
nyechiel, line 12: I would say that server and client can be placed within the same cluster or spread across multiple clusters.

issue-40-built-in-benchmarking-tool.md
```diff
@@ -0,0 +1,31 @@
+## Add Built-in Benchmarking Tool
+
+## Summary
+
+There is a need to support a built-in benchmarking tool in Submariner.
```
nyechiel, line 5: More context: this is to provide users with easy way to benchmark the network performance within and across clusters, and experiment with different tuning of the system.

issue-40-built-in-benchmarking-tool.md
```diff
@@ -0,0 +1,32 @@
...
+This is to provide users with easy way to benchmark the network performance within and across clusters, and experiment with different tuning of the system.
+
+## Proposal
+
+Add a new ability to have a built-in benchmark tool.
+Implement a new Subctl command that will use iperf3 internally to benchmark the throughput between the two nodes.
+
+scheduling server and client can be placed within the same cluster or spread across multiple clusters.
```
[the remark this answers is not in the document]
nyechiel: I agree that the main use case we should target is testing *across* different clusters.. but technically one should be able to use the tool as is to also benchmark within a single cluster. Or do you suggest we explicitly block this option?

issue-40-built-in-benchmarking-tool.md
sridhargaddam: Even in the datapath-connectivity tests, we kind of spawn a server/client (i.e., listener pod, connector pod), so it makes sense to align them for consistency.

The current command is as follows. So, if we decide to do the change, it will not require any changes to client scripts as well and would just require changes to subctl alone.

subctl verify \<kubeConfig1\> \<kubeConfig2\> [flags]

sridhargaddam: > `--server-config` and `--client-config` specify the kubeconfigs for both clusters.

If --server-config/--client-config is for kubeconfig's, how do we provide the server/client options (like protocol/bitrate/time/windowSize etc)? as config-file or as command-line-args?

issue-40-built-in-benchmarking-tool.md
```diff
@@ -0,0 +1,32 @@
...
+
+The user can specify source and destination nodes via node names. (Or via lables?)
```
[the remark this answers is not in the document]
sridhargaddam: Personally, I prefer nodeNames here for simplicity, but the same can be achieved via selectors as well. 
Instead of taking "cluster2-worker" as input from user, we would take "kubernetes.io/hostname=cluster2-worker" (or any other label of choice). 

I'm fine with labels, but this needs to be properly documented as the same label can be applied on multiple nodes and nettest could be scheduled on one of them.

issue-40-built-in-benchmarking-tool.md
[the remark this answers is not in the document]
greenren: While measuring Submariner's network performance, we found it useful to have some sort of baseline within a cluster to compare the cross-cluster traffic to. For example, if intercluster network performance is low, it makes sense that intracluster network performance is also limited. So it's not an exact comparison, but it does provide a useful context for the Submariner benchmark.

issue-40-built-in-benchmarking-tool.md
greenren: > @skitt are there any parameters (e.g TCP Window) we need to expose at the "subctl test throughput" level or are we assuming that everything goes into the config files?

I'm also curious about how a user will be able to adjust the iperf3 parameters and what the defaults will be.
Will the benchmark loop through different number of parallel client streams by default?

issue-40-built-in-benchmarking-tool.md
```diff
@@ -0,0 +1,35 @@
...
+Add a new ability to have a built-in benchmark tool.
+Implement a new Subctl command that will use iperf3 internally to benchmark the throughput between two nodes.
+
+scheduling server and client can be placed within the same cluster or spread across multiple clusters.
+
+The user can specify source and destination nodes via node names. (Or via lables?)
+
+If nodes are not specified, the following tests will be run by default:
+* non-gateway to non-gateway  
+* gateway to gateway
```
sridhargaddam, line 22: @skitt: Based on the ODCN experience, do we want to give an option to the user to schedule pods with HostNetworking?

@nyechiel @mkolesnik @maayanf24 thoughts?

issue-40-built-in-benchmarking-tool.md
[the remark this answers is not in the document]
mkolesnik: Thanks for the input @greenren 

I feel that this tool's strong point would be to measure the metrics across clusters, hence to me that should be the focus.
Obviously you can measure intra-cluster traffic also, but as submariner is not involved in the networking there (we strive to be as agnostic as possible about the specific cluster CNI in fact), it doesnt make much sense for these measurements to be a focus for this tool.

How I see this tool being useful is try out different submariner configurations across clusters and measuring and comparing that, for example to check submariner with and without globalnet and compare the impact of using that feature on the data plane. Another comparison is to check different cable drivers or cable driver configurations.

While it's true that intra-cluster network performance would affect the inter-cluster results, it would do so in a consistent matter which is comparable (if the disturbance isn't transient), but comparing inter-cluster results to intra-cluster results would give no benefit as it's akin to comparing the speed of a virtual NIC to the speed of a physical NIC.

To be clear, I'm not saying this tool couldn't be possibly used to measure intra-cluster performance, I just think that as it's not the focus nor the intention of this tool, this mode should not be advertised or supported. I.E. it would be provided "as is" without any guarantee to serve this purpose.

I'm sure there are probably many networking tools that can be used to measure intra-cluster networking performance, possibly even in the K8s ecosystem.

issue-40-built-in-benchmarking-tool.md
mkolesnik: As I mentioned before, to me it seems that the aim of this tool is to measure inter-cluster traffic in order to test and compare different configurations of submariner (or perhaps you could compare it with other solutions, though I'm not sure that's something we'd like to support or not).

As such, I argue that the best path forward is to give a simple set of tests that would be performed by the tool and reported so they can be easily compared.

We need to decide which metrics make sense and which deployments make sense to test.

Obviously this list can be further expanded, but I feel the tool should remain simple and intuitive and give clear (and few) options to the user to facilitate ease of use.

We're not trying to build "the ultimate K8s network benchmarking tool" here, so I think we should keep it simple and focus on the things that give the most value to the most people.

issue-40-built-in-benchmarking-tool.md
mkolesnik: As per my comment on line 22, I don't think it makes sense to expose any of the tool internals to the user. If we want to change the internal benchmarking tool later, exposing these options as API either binds us to this implementation, or forces us to jump through loops or break the API.

issue-40-built-in-benchmarking-tool.md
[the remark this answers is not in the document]
mangelajo: I see @mkolesnik point here about... not measuring something that it's specific to submariner.

But since supporting intra-cluster should not make it specifically much more complicated (it adds the intra-node / node to node dimension) .. I think it's good to have this so we can provide a good baseline to compare, which proved very useful as a baseline as @greenren said.

May be this is something we can propose as a second iteration (given that it requires extra work, although not much). ?

issue-40-built-in-benchmarking-tool.md
mangelajo: I agree with @mkolesnik here on providing a set of standard tests which would let people compare deployment to deployment easily.. may be we can add a selector for specific parts (for example if one path is being optimized and they need to re-run a specific bit, but I'd say that's a later enhancement).

issue-40-built-in-benchmarking-tool.md
```diff
@@ -0,0 +1,35 @@
...
+
+
+#### Source code location
+
+Source code should go into Submariner.
```
mangelajo, line 27:
```suggestion
Source code should go into `submariner-operator`. 
```

issue-40-built-in-benchmarking-tool.md
mangelajo: May be it's easier to not provide lots of options here as a start, and define a standardized testing, 

Like gw to gw and non-gw to non-gw to start with. Then we can start thinking about how to add fine tuning options, or the option to test in-cluster to add a baseline (may be we can include that part from the automated testing?)

issue-40-built-in-benchmarking-tool.md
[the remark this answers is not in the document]
sridhargaddam: I agree with @greenren and @mangelajo. Since Submariner also programs certain iptable chains/routing rules on the nodes, it will be helpful if we allow the tool to validate throughput within the local cluster. This way, we can use the tool to compare the throughput with and without Submariner in the local cluster as well.

issue-40-built-in-benchmarking-tool.md
sridhargaddam: I have a slightly different opinion on the scope of the tool.

Personally, I feel that "subctl test throughput" is more like a wrapper/facilitator to schedule Pods on required nodes and run the performance tests with whatever configuration user provides and display the output as-is to the user.
The job of the tool is NOT to interpret the results or recommend any changes, it would be the job of the user running the tool.

So "subctl test throughput" would take the following inputs:
1. The backend tool that would be used to run the tests with default tool as netperf (or whatever we decide)
2. label of node where the Server/Client has to be scheduled
3. Server Options: This can be read from a configuration file and will be passed "as-is" to the respective tool while running the server side of the tool.
   Example: If "-s -1" is included in the configuration file, then iperf would be executed as "iperf -s -1"
4. Similarly client options: as above.

Each tool (like iperf or netperf) can display the output in its own manner. After the execution of the tests, subctl will simply output the same to the user. What are your thoughts on this approach?

sridhargaddam: @maayanf24 Okay, based on the discussion we had yesterday, please identify the options that can be used for validating throughput and latency to begin with and we can expand the options (in future) based on feedback.

issue-40-built-in-benchmarking-tool.md
sridhargaddam: Okay, for now, we can skip this and we can consider this as a future item.

issue-40-built-in-benchmarking-tool.md
```diff
@@ -0,0 +1,47 @@
...
+## Summary
+
+This is a proposal to support a built-in benchmarking tool in Submariner.
+The goal is to provide users with easy way to benchmark the network performance across clusters, and experiment with different tuning of the system.
+
+
+## Proposal
+
+Add a new ability to have a built-in benchmark tool.
+Implement a new Subctl command that will use uperf internally to benchmark the throughput and round trip latency between two nodes.
```
sridhargaddam, line 14:
```suggestion
Implement a new Subctl command that will use iperf/netperf internally to benchmark the throughput and round trip latency between two nodes.
```

issue-40-built-in-benchmarking-tool.md
```diff
@@ -0,0 +1,47 @@
...
+
+Scheduling server and client can be placed within the same cluster or spread across multiple clusters.
+
+The user can specify source and destination nodes via node labels.
```
sridhargaddam, line 18: Would be helpful if you provide an example here

issue-40-built-in-benchmarking-tool.md
```diff
@@ -0,0 +1,47 @@
...
+
+In case server-options are not included, the command will use defaults.
+
+The backend for latency test will be netperf. The default params that will be passed to netperf:
+```
+-t TCP_RR  -- -o min_latency,mean_latency,max_latency,stddev_latency,transaction_rate
+```
+
+The backend for throughput will be iperf. The default params that will be passed to iperf:
+```
```
sridhargaddam, line 32:
```suggestion
The backend for throughput tests will be iperf. The default params that will be passed to iperf:
```
```

File: issue-40-built-in-benchmarking-tool.md [hunk 11]
```diff
@@ -0,0 +1,47 @@
...
+-w 8M -P 10 -c
```
sridhargaddam, line 33: Please elaborate on the significance of each of these arguments 
-w window size/ socket buffer size 
-P <....> 
-c <....>
etc

issue-40-built-in-benchmarking-tool.md
```diff
@@ -0,0 +1,47 @@
...
+
+If node labels are not specified, the following tests will be run by default:
+* non-gateway to non-gateway  
+* gateway to gateway
+
+In case server-options are not included, the command will use defaults.
+
+The backend for latency test will be netperf. The default params that will be passed to netperf:
+```
+-t TCP_RR  -- -o min_latency,mean_latency,max_latency,stddev_latency,transaction_rate
```
sridhargaddam, line 28: Please elaborate the default args

issue-40-built-in-benchmarking-tool.md
```diff
@@ -0,0 +1,47 @@
...
+
+The backend for throughput will be iperf. The default params that will be passed to iperf:
+```
+-w 8M -P 10 -c
+```
+
+#### Source code location
+
+Source code should go into Submariner. 
+There is going to be a dependency on the nettest image from Shipyard (Which has iperf).
```
sridhargaddam, line 39: Instead of explicitly specifying iperf, probably we can just say that the image has necessary tools.

issue-40-built-in-benchmarking-tool.md
```diff
@@ -0,0 +1,47 @@
...
+
+
+### Subctl command
+
+```
+subctl test throughput <kubeconfig1> <kubecinfig2> --node-label --server-options
+subctl test latency <kubeconfig1> <kubecinfig2> --node-label --server-options
```
sridhargaddam, line 46:
```suggestion
subctl test throughput <kubeconfig1> <kubeconfig2> --node-label --server-options
subctl test latency <kubeconfig1> <kubeconfig2> --node-label --server-options
```

sridhargaddam: "--node-label" is not followed by the necessary label?
Similarly "--server-options"

issue-40-built-in-benchmarking-tool.md
maayanf24: @mangelajo - this was discussed already and agreed that the test will be part of the Submariner

mangelajo: I'm but why submariner?, I envisioned this as a part of subctl, so you could test from a single tool? , 

What will we do, building another tool and publishing another tool somewhere?, I think it's more practical (and makes more sense) to have it in subctl.

issue-40-built-in-benchmarking-tool.md
```diff
@@ -0,0 +1,56 @@
...
+
+Source code should go into Submariner.   
+There is going to be a dependency on the nettest image from Shipyard, which will need to have the necessary tools.
+
+
+### Subctl command
+
+```
+subctl test throughput <kubeconfig1> <kubeconfig2>
+subctl test latency <kubeconfig1> <kubeconfig2>
```
mangelajo, line 55:
```suggestion
subctl benchmark throughput <kubeconfig1> <kubeconfig2>
subctl benchmark latency <kubeconfig1> <kubeconfig2>
```

mangelajo:
This was never merged, I'm merging regardless of the votes since it was already implemented I think it's better to keep it for reference and move on.

PR #9, opened 2020-08-26
Add CODEOWNERS file

dfarrell07:
Signed-off-by: Daniel Farrell <dfarrell@redhat.com>

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+* @mangelajo @tpantelis @Oats87


CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @mangelajo @tpantelis @Oats87
```
skitt, line 1: What are the criteria for the initial code owners? Wouldn’t it make more sense to use the list of current contributors, *i.e.* @dfarrell07, @maayanf24 and @mangelajo?

skitt (changes requested):

CODEOWNERS
dfarrell07: Per https://submariner.io/contributing/community-membership/#project-owner, they are members of the https://github.com/orgs/submariner-io/teams/submariner-core group.

skitt: Ah, thanks. Wouldn’t it make sense however to include existing contributors when adding `CODEOWNERS`, at least the major contributors, as per “History of review and authorship”?

dfarrell07: Yeah, that does make some sense, but I think it doesn't quite line up with the governance. The detailed version of the committer requirements is: https://submariner.io/contributing/community-membership/#committer-requirements. The idea is that the person should develop a history of review/contribution, be sponsored by project owners, and then publically apply. I think the current contributors should apply to be code owners, but I don't think we should skip those other steps of the process.

skitt: Right, it doesn’t align with the governance, but we’re transitioning this project to align with the governance by adding `CODEOWNERS` ;-). I guess this matches what we did in the past, telling people that their permissions would be reduced until they applied and got sponsored.

skitt (approved):

CODEOWNERS
skitt: So really my thoughts on this is that we should ensure that all new projects are immediately created with this `CODEOWNERS` file, and no administrators (once we settle the project configuration settings so that not having administrators is viable, if that is in fact possible).

PR #11, opened 2020-10-02
Support OVN-Kubernetes making route-agent pluggable

mangelajo:
Proposes refactoring out the route-agent into a pluggable
architecture that can be reused for OVN, as well as
a separate deployment which would take care of synchronizing
with the network plugin as necessary.
    
Co-authored-by: Sridhar Gaddam <sgaddam@redhat.com>


mangelajo:
@sridhargaddam added this, but it's super draft. I'll continue working on it on monday.

submariner/pluggable-network-plugin-and-ovn.md
```diff
@@ -0,0 +1,50 @@
+## Add support for OVN with pluggable network-plugin support
+
+[submariner#778](https://github.com/submariner-io/submariner/issues/778)
+
+## Summary
+
+Today submariner supports working with existing network-plugins which are based
+on the kubeproxy and iptables implementations. Any network-plugin outside of this
+case won't work, OVN-Kubernetes is an example.
```
sridhargaddam, line 9:
```suggestion
Today Submariner supports working with existing network-plugins which are based
on the kube-proxy with iptables backend. Any network-plugin outside of this
case won't work, OVN-Kubernetes is an example.
```

submariner/pluggable-network-plugin-and-ovn.md
```diff
@@ -0,0 +1,50 @@
...
+
+This is a proposal to support pluggingable network drivers, starting by OVN-kubernetes.
```
sridhargaddam, line 11:
```suggestion
This is a proposal to support pluggable network drivers, starting with OVN-Kubernetes.
```

submariner/pluggable-network-plugin-and-ovn.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Proposal
+
+Re-architect submariner-routeagent in a way that different network-plugin implementations
+can be provided. Those drivers would be notified of different transitions on a node:
```
sridhargaddam, line 16:
```suggestion
Re-architect Submariner route-agent in a way such that different network-plugin implementations
can be supported. Those drivers would be notified of different events like:
```

submariner/pluggable-network-plugin-and-ovn.md
```diff
@@ -0,0 +1,50 @@
...
+
+* Transition of a gateaway node in your cluster
+* Connection of a new cluster (and it's CIDR details)
+* Disconnection of existing clusters (and it's CIDR details)
+* ... any other event necessary -TBD-
```
sridhargaddam, line 21: Initialization events to handle any CNI Specific tweaks, creating necessary chains etc
Notification when new route-agent pods are scheduled in the local cluster (needed for vxlan based drivers)

submariner/pluggable-network-plugin-and-ovn.md
```diff
@@ -0,0 +1,50 @@
...
+of the underlying network plugin the current cluster uses.
+
+These network-plugin specific drivers will interface to the underlying network-plugin
+and prepare it's configuration as necessary, as well as any routing or linux network
+stack changes necessary on the worker/gateway nodes of the cluster.
+
+
+#### Technologies explored and alternatives
+
+* Talk about eBPF
```
sridhargaddam, line 33: I think we can briefly summarize why we did not consider eBPF and provide a link to the eBPF Analysis document here

submariner/pluggable-network-plugin-and-ovn.md
```diff
@@ -0,0 +1,50 @@
...
+
+Source code should go into `submariner`, specifically the route-agent.
+
+
+#### Work items
+
+[ ] Define an initial architecture for the network plugin drivers
+[ ] Improve the submariner-operator to pass down the network-plugin type down
+    to the route-agent.
+[ ] Converting the existing implementation into kubeproxy-iptables default driver,
```
sridhargaddam, line 45:
```suggestion
[ ] Converting the existing implementation into kube-proxy iptables default driver, 
```

sridhargaddam (commented):
Thanks for putting this together @mangelajo

submariner/pluggable-network-plugin-and-ovn.md
sridhargaddam: https://docs.google.com/document/d/1KcrxZROb3v-1EW8B-lOKls4GNwWl73Zg02LytNEbuck/edit

submariner/pluggable-network-plugin-and-ovn.md
mangelajo: lol, I wrote plugginable X'D

submariner/pluggable-network-plugin-and-ovn.md
mangelajo: +1, I'll add that

submariner/pluggable-network-plugin-and-ovn.md
```diff
@@ -0,0 +1,289 @@
...
+Source code should go into `submariner`, specifically the route-agent.
+
+#### Work items
+
+* [ ] Define an initial architecture for the network plugin drivers
+* [ ] Improve the submariner-operator to pass down the network-plugin type down
+      to the route-agent.
+* [ ] Converting the existing implementation into kube-proxy iptables default driver,
+      while refining the initially defined interfaces.
+* [ ] Program the integration to OVN: connection to NBDB (& SBDB ?)
```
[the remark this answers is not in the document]
mangelajo: No changes to those necessary (as far as I've seen) I mean programming our own connection to NBDB, we need to get the credentials, etc..

submariner/pluggable-network-plugin-and-ovn.md
```diff
@@ -0,0 +1,289 @@
...
+#### Source code location
+
+Source code should go into `submariner`, specifically the route-agent.
+
+#### Work items
+
+* [ ] Define an initial architecture for the network plugin drivers
+* [ ] Improve the submariner-operator to pass down the network-plugin type down
+      to the route-agent.
+* [ ] Converting the existing implementation into kube-proxy iptables default driver,
```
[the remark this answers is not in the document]
mangelajo: There will be no changes from the user point of view.

We will default to the new created kube-proxy iptables driver (I'll make that clear).

submariner/pluggable-network-plugin-and-ovn.md
```diff
@@ -0,0 +1,289 @@
...
+  * eBPF would not work on Windows hosts, while, OVN-Kubernetes can work on Windows hosts.
+
+A more detailed view can be found in:
+[eBPF/XDP analysys](https://docs.google.com/document/d/1KcrxZROb3v-1EW8B-lOKls4GNwWl73Zg02LytNEbuck)
+
+#### Source code location
+
+Source code should go into `submariner`, specifically the route-agent.
+
+#### Work items
```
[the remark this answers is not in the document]
mangelajo: +1 Correct, I have thought about this but forgot to add it. I'll add it, thanks.

submariner/pluggable-network-plugin-and-ovn.md
```diff
@@ -0,0 +1,289 @@
...
+#### Alternatives
+
+* eBPF/XDP was explored as an alternative to the current proposal (for the OVN interface),
+it was discarded because, while eBPF provides a very powerful interface for dataplane
+packet processing, when used on top of an existing solution (OVN/OVS/iptables) we found
+the following issues:
+  * There is no interface to manipulate/work with the existing connection tracking entries
+    on the host.
+  * We needed to design a whole network-plugin in parallel to the existing network plugin
+    for traffing going to the remote clusters, and traffic comming back.
```
mangelajo, line 268: traffing -> traffic

submariner/pluggable-network-plugin-and-ovn.md
```diff
@@ -0,0 +1,289 @@
...
+    iptables -I FORWARD -i ens5 -s 10.132.0.0/14        -d 172.30.0.0/16 -o ovn-k8s-gw0 -j ACCEPT
+
+
+    ip rule add from 10.132.0.0/14 to 10.128.0.0/14 table 5
+    ip rule add from 10.132.0.0/14 to 172.30.0.0/14 table 5
+    ip route add default via $SUBM_EXT_PORT_IP_A table 5
+
+
+
+
```
mangelajo, line 256: unnecessary empty lines

submariner/pluggable-network-plugin-and-ovn.md
mangelajo: comming -> coming

mangelajo:
@nyechiel @tpantelis @sridhargaddam  ^ :)

nyechiel (approved):
LGTM, this should be enough to get us started and introduce initial support in v0.8

submariner/pluggable-network-plugin-and-ovn.md
```diff
@@ -0,0 +1,300 @@
...
+case won't work, OVN-Kubernetes is an example.
+
+This is a proposal to support pluggable network drivers, starting with OVN-Kubernetes.
+
+## Proposal
+
+Re-architect Submariner route-agent in a way such that different network-plugin implementations
+can be supported. Those drivers would be notified of different events like:
+
+* `Initialization` event
```
sridhargaddam, line 18: Screenshot from 2020-10-14 11-25-29

When this is rendered, its displayed as a single sentence and "Called" is capitalized in all the events.

mangelajo: Ouch, ok... if it's ok, I'd prefer to merge it as it is... it's only a reference for our development.

submariner/pluggable-network-plugin-and-ovn.md
```diff
@@ -0,0 +1,300 @@
...
+    We may want to add an indication on the driver, to say if the driver needs leader election
+    or not. kube-proxy iptables does not need it, and leader election loads the K8s API.
+
+* `Transition to Non-Gateway` event
+    During Gateway to Non-Gateway events the route agent on the specific node
+    may need to perform specific cleanups
+
+* `Transition to Gateway` event
+    When a node becomes Gateway some settings may need to be adjusted on the node
+    by the driver.
```
mkolesnik, line 44: This sounds very specific to iptables but I'm not sure if it makes sense for other driver types..

Also if we model it more "micro-services" like then I'm not sure we need too many specific events as we could just have the different parts react to different states of resources in K8s.

mkolesnik (dismissed):
I'm not sure that the "route agent" we have right now isn't an iptables driver already.. For other connectivity options, we might want different services altogether.

submariner/pluggable-network-plugin-and-ovn.md
```diff
@@ -0,0 +1,300 @@
...
+
+* `Uninstall` event
+    Called once if the route agent is being removed from a node, being Submariner un-installed,
+    or the route-agent driver being switched, a full cleanup will be necessary in this case.
+
+* `Leader`/`Non-leader` events
+    Some parts of the drivers could need a single leader handling specific synchronization
+    work with the network plugin. Those events will be called for drivers that require
+    this type of behavior.
+
```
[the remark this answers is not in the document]
mangelajo: Yes, probably it would make sense as a 2nd iteration. Specially since not all drivers would need something like that. 

I'm ok with documenting it for the long term and start by the simpler approach.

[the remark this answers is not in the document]
mangelajo: I will add those comments.

submariner/pluggable-network-plugin-and-ovn.md
mangelajo: This part is also important for OVN, ovn will need to know which one is the active gateway, so it can re-assign the chassis-gateway for the router external port, receiving events for that it's important.

sridhargaddam: AFAIU this will be required for most of the drivers as Gatewaynode requires different configuration compared to other nodes in the cluster.

submariner/pluggable-network-plugin-and-ovn.md
sridhargaddam: Ack

sridhargaddam (approved):

submariner/pluggable-network-plugin-and-ovn.md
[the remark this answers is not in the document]
mangelajo: Ok, I refectored it, after thinking more about it I believe it makes more sense to spawn a separate deployment from the start just for this.

submariner/pluggable-network-plugin-and-ovn.md
```diff
@@ -74,6 +70,16 @@ These network-plugin specific drivers will interface to the underlying network-p
...
 stack changes necessary on the worker/gateway nodes of the cluster.
 
+### Synchronizer implementation
+
+Some parts of the drivers could need a single worker handling specific synchronization
+work with the network plugin.
+
+The drivers declare in a flag if they need this type of instance.
+
+The synchronizer instances receive the same exact events, but instead of configuring
```
sridhargaddam, line 80: Two questions @mangelajo 

1. If we are going to spawn a new deployment (with replica count as 1), will we use the same image for that?

2. When you say that synchronization instance will receive the same exact events. Will it receive **all the events** listed above? Won't it be an overkill for the benefit of having it inside the driver itself (i.e., allow the driver implementation to use its own leader election and remove the leader events)

mangelajo: > Two questions @mangelajo
> 
> 1. If we are going to spawn a new deployment (with replica count as 1), will we use the same image for that?

We could use the same image/binary, just on a different mode, so we don't need to have two different type of drivers/interfaces. As they would receive the same settings etc...

but just an idea, trying to make things simpler. We can always split to more images/binaries if we need.

> 2. When you say that synchronization instance will receive the same exact events. Will it receive **all the events** listed above? Won't it be an overkill for the benefit of having it inside the driver itself (i.e., allow the driver implementation to use its own leader election and remove the leader events)


I think some of the events would become irrelevant for OVN on that synchronizer instance, we need to figure those details out. May be there are events that we can filter if the driver don't need them (in a later optimization phase?) but I think we may want to keep things as simple as possible if we want to have something this sprint.

May be we should have a meeting with @mkolesnik about these topics :-)

sridhargaddam: As you know, creating a new deployment has implications on the installer tools like operator and helm. Also, if we use the same image, then within the driver code, we have to include checks if its doing regular event processing or if its purpose is to program the CNI. If we use a different image, then we will have associated issues like versioning, mis-matches in driver vs new image etc etc

I'm not against this proposal, but IMHO I get a feeling that splitting the route-agent driver into two pods will create more maintenance headaches. 

If we have the code as part of driver itself, it can use a leader election (if required) or the route-agent running on gatewayNode can act as leader. BTW, figuring out the active leader is very easy if we use "kubectl describe cm <configMapName>" - the output includes information about the active leader holding the lock.

Lets discuss with @mkolesnik for his inputs.

sridhargaddam (approved):
LGTM, thanks @mangelajo

submariner/pluggable-network-plugin-and-ovn.md
```diff
@@ -0,0 +1,342 @@
...
+* `Transition to Gateway` event
+
+    When a node becomes Gateway some settings may need to be adjusted on the node
+    by the driver.
+
+* `Periodic` event
+
+   We could provide a periodic event to the driver, but may be an alternative
+   is doing this directly in the driver, using a timer.
+
```
mkolesnik, line 57:
```suggestion
```

This one seems too broad for me, we could just have specific events for any periodic things that we want all "drivers" to support but they'd be better defined explicitly

submariner/pluggable-network-plugin-and-ovn.md
```diff
@@ -0,0 +1,342 @@
...
+* `Update of a remote Endpoint`
+
+* `Deletion of a local Endpoint`
+
+* `Deletion of a remote Endpoint`
+
+* `Node creation` on the cluster
+
+* `Node removal` on the cluster
```
mkolesnik, line 72: Are these relevant for us?

mkolesnik (changes requested):

submariner/pluggable-network-plugin-and-ovn.md
mangelajo: Yes, for both the route agent and the sync agent. Now the route agent listen on pods, but just to find Node ips. We plan to change that.

submariner/pluggable-network-plugin-and-ovn.md
mangelajo: We ended removing this one from the PR we posted this morning, this is implementation specific, if any driver/event handler needs a periodic timer can be created internally by himself... doesn't make sense to have it here, I agree.

mangelajo:
Opps I'll fix the last commit

mkolesnik (approved):

skitt (approved):

PR #20, opened 2020-12-16
Define a stable release strategy

mangelajo:

### DIFF
--- a/releases/stable-branches-and-release-cycle.md
+++ b/releases/stable-branches-and-release-cycle.md
@@ -1,0 +1,113 @@
+# Producing and supporting stable releases
+
+## Summary
+
+Today the Submariner project releases one minor version a month without capacity
+for stable branches or bugfix releases. In the following proposal we define a plan
+to make this possible starting with 0.8.
+
+## Proposal
+
+### Alignment of branch -> image releases
+
+We currently produce `<image>:devel` for every master push, along with `<image>:x.y.z` for
+the releases.
+
+Changing the `master` branch name to `devel` would make this homogeneous so our image releases
+are simply `<image>:<branch-or-tag>`.
+
+### Creating one stable branch for every minor release
+
+Approaching the first release candidate `rc0` of every minor or major version, a new stable
+branch will be forked off `devel` branch in the form of `release-x.y` by the release scripts,
+and a patch or PR pinning the shipyard-base-image to `release-x.y` will be applied by the
+automated process.
+
+Any following changes related to pinning go modules in submariner, admiral, lighthouse, etc...
+will be done in the `release-x.y` branch directly (no cherry picks).
+
+Once the branch exists, any patches which should be part of the release will need to be
+merged in `devel`and backported to `release-x.y`.
+
+### Maintaining a stable branch
+
+#### Latest stable branch (N)
+
+The only patches that should be submitted to the latest stable branch are
+bugfixes and not features, unless for exceptions previously agreed on the
+community meeting.
+
+Any patches that are necessary for fixing bugs in a stable branch will
+need to be merged first into `devel`, and then cherry-picked to the stable
+branch.
+
+`release-0.9` <-- `devel` <-- `bug-fix patch`
+
+*No-change version bumps* will be done if a base image has been updated
+due to a security bugfix, our code would remain the same, but our versions
+would be bumped to pickup the new images.
+
+#### Previous stable branch (N-1)
+
+The only patches that should be submitted to the previous stable branch (N-1)
+are critical bugfixes, being the definition of critical bugfix: security
+related fixes, and CVEs.
+
+The patches necessary for the critical bugfixes should be cherry-picked from
+previous  versions, though sometimes they could need specific adaptations
+to the branch.
+
+`release-0.8`  <-- `release-0.9` <-- `devel` <-- `bug-fix patch`
+
+*No-change version bumps* will be necessary if a base image has been updated
+due to a security bugfix, our code would remain the same, but our versions
+would be bumped to pickup the new images.
+
+#### Releasing a stable branch
+
+When a patch is merged to a stable branch the images for such branch will be updated, for
+example `quay.io/submariner/submariner:release-x.y`.
+
+This image allows for testing a set of bugfix patches, sometimes multiple patches are expected
+to be required for one fix. Once the release branch is ready we should create a new x.y.**z**
+patch release via the release repository.
+
+### How long is a stable branch maintained
+
+The current stable branch (N) will be maintained for normal bugfixes until
+the next stable version is released. The previous stable branch (N-1) will be
+maintained only for critical bugfixes, however this can change in the future if
+we introduce the concept of long term support releases.
+
+### Reducing the cadence of minor releases
+
+Based on the previous ideas, the existing minor version cadence doesn't seem realistic
+for maintaining stable branches. We don't want "fake" stable branches where patches
+are never backported.
+
+With the new cadence, after each sprint we would produce a milestone release,
+for example
+
+* `0.9-milestone-1-rc0`, ... , `0.9-milestone-1`
+* `0.9-milestone-2-rc0`, ... , `0.9-milestone-2`
+* `0.9-milestone-3-rc0`, ... , `0.9-milestone-3`
+* `0.9-milestone-4-rc0`, ... , `0.9-milestone-4`
+
+In this case we won't do pinning of shipyard base image, but we will release and update
+the admiral & shipyard `go.mod` git references, without pinning shipyard.
+
+After a number of milestones instead we would follow the usual pattern for a release:
+
+* `0.9.0-rc0` along with its stable branch `release-0.9` as described
+  previously, ..., `0.9.0`
+
+The actual cadence and release times is TBD by the community, it could be aligned with
+k8s for example, slightly after a release of k8s we could adapt to the new changes
+in k8s API and produce a final release.
+
+## Work items
+
+* [ ] all repositories: produce `<image>:<branch>` images on push
+* [ ] releases repo support for forking repositories approaching a stable release.
+* [ ] all necessary changes in the releases repository
+* [ ] all necessary changes in CI scripts.


releases/stable-branches-and-release-cycle.md
```diff
@@ -0,0 +1,79 @@
...
+
+We produce currently `<image>:devel` for every master push, along with `<image>:x.y.z` for
+the releases.
+
+Changing the `master` branch name to `devel` would make this homogeneous so our image releases
+are simply `<image>:<branch-or-tag>`.
+
+### Creating one stable branch for every minor release
+
+Approaching the release of every minor or major version a new stable branch will be forked
```
mkolesnik, line 21:
```suggestion
Approaching the first release of every minor or major version, a new stable branch will be forked
```

releases/stable-branches-and-release-cycle.md
```diff
@@ -0,0 +1,79 @@
...
+off `devel` branch in the form of `release-x.y` for every existing repository.
+
+Any changes related to pinning shipyard base image will be done in that branch.
+
+At that point in time, any patches which belong to the release will need to be merged in master
+and backported to `release-x.y`. For this reason, the later we branch, the less extra work
```
sridhargaddam, line 27: Maybe we can create the branch on release-x.y-rc0

releases/stable-branches-and-release-cycle.md
```diff
@@ -0,0 +1,79 @@
...
+
+Changing the `master` branch name to `devel` would make this homogeneous so our image releases
+are simply `<image>:<branch-or-tag>`.
+
+### Creating one stable branch for every minor release
+
+Approaching the release of every minor or major version a new stable branch will be forked
+off `devel` branch in the form of `release-x.y` for every existing repository.
+
+Any changes related to pinning shipyard base image will be done in that branch.
```
mkolesnik, line 24: Generally all changes related to release, such as pinning to admiral or lh/subm as well.

releases/stable-branches-and-release-cycle.md
```diff
@@ -0,0 +1,79 @@
...
+
+At that point in time, any patches which belong to the release will need to be merged in master
```
mkolesnik, line 26:
```suggestion
Once the branch exists, any patches which should be part of the release will need to be merged in `devel`
```

releases/stable-branches-and-release-cycle.md
```diff
@@ -0,0 +1,79 @@
...
+and backported to `release-x.y`. For this reason, the later we branch, the less extra work
+will be necessary to produce a stable release.  
+
+### Maintaining a stable branch
+
+Any patches that are necessary for fixing bugs in a stable branch will need to be merged
```
sridhargaddam, line 32: Until how many stable branches do we want to continue backporting?

releases/stable-branches-and-release-cycle.md
mkolesnik, line 27: This though only applies to the first release of a branch, any subsequent releases will have to have (the desired) patches merged to both.

releases/stable-branches-and-release-cycle.md
```diff
@@ -0,0 +1,79 @@
...
+first into master, and the later stable branches.
```
mkolesnik, line 33:
```suggestion
first into `devel`, and the later stable branches.
```

releases/stable-branches-and-release-cycle.md
```diff
@@ -0,0 +1,79 @@
...
+
+`release-0.8`  <-- `release-0.9` <-- `devel`
+
+Patches could be simple cherry-picks, or sometimes they could need specific adaptations
```
mkolesnik, line 37:
```suggestion
Patches could be simple cherry-picks, though sometimes they could need specific adaptations
```

releases/stable-branches-and-release-cycle.md
```diff
@@ -0,0 +1,79 @@
...
+to the branch.
+
+When a patch is merged to a stable branch the images for such branch are updated, for example
+`quay.io/submariner/submariner:release-0.9`
```
sridhargaddam, line 41: Instead of updating release-0.9 image itself, we should probably create a new image like `quay.io/submariner/submariner:release-0.9.x" and increment x as we go along. WDYS?

releases/stable-branches-and-release-cycle.md
```diff
@@ -0,0 +1,79 @@
...
+
+Any patches that are necessary for fixing bugs in a stable branch will need to be merged
+first into master, and the later stable branches.
+
+`release-0.8`  <-- `release-0.9` <-- `devel`
+
+Patches could be simple cherry-picks, or sometimes they could need specific adaptations
+to the branch.
+
+When a patch is merged to a stable branch the images for such branch are updated, for example
```
mkolesnik, line 40:
```suggestion
When a patch is merged to a stable branch the images for such branch will be updated, for example
```

releases/stable-branches-and-release-cycle.md
```diff
@@ -0,0 +1,79 @@
...
+Approaching the release of every minor or major version a new stable branch will be forked
+off `devel` branch in the form of `release-x.y` for every existing repository.
+
+Any changes related to pinning shipyard base image will be done in that branch.
+
+At that point in time, any patches which belong to the release will need to be merged in master
+and backported to `release-x.y`. For this reason, the later we branch, the less extra work
+will be necessary to produce a stable release.  
+
+### Maintaining a stable branch
```
mkolesnik, line 30: I think we're missing here that we're only expecting backports of bugs not features (except exceptions).

releases/stable-branches-and-release-cycle.md
```diff
@@ -0,0 +1,79 @@
...
+
+With the new cadence after each sprint we would produce a milestone release,
+for example
+
+* `0.9-milestone-1-rc0`, ... , `0.9-milestone-1`
+* `0.9-milestone-2-rc0`, ... , `0.9-milestone-2`
+* `0.9-milestone-3-rc0`, ... , `0.9-milestone-3`
+* `0.9-milestone-4-rc0`, ... , `0.9-milestone-4`
+
+In this case we won't do pinning of shipyard base image, but we will release and update
```
mkolesnik, line 62: I think this can apply to stable branches as well, e.g. when we create the branch we'll pin it to shipyard 'release-x.y.z' which will always remain on the compatible version.
Makes sense?

releases/stable-branches-and-release-cycle.md
```diff
@@ -0,0 +1,79 @@
...
+the admiral & shipyard `go.mod` references.
+
+After a number of milestones instead we would follow the usual pattern for a release:
+
+* `0.9.0-rc0`, ..., `0.9.0` along with it's stable branch `release-0.9` as described
+previously.
```
mkolesnik, line 68: And at GA of that release we create the branch right?

releases/stable-branches-and-release-cycle.md
```diff
@@ -0,0 +1,79 @@
...
+
+The actual cadence and release times is TBD by the community, it could be aligned with
+k8s for example, slightly after a release of k8s we could adapt to the new changes
+in k8s API and produce a final release.
```
mkolesnik, line 72: I think it makes sense to release once theres what to release (eg enough features).

mkolesnik (changes requested):
Thanks for capturing the discussion into an enhancement proposal, Miguel!

I've some comments and suggestions

releases/stable-branches-and-release-cycle.md
SteveMattar: I think patches should be created in `devel` first and then cherry pick anywhere else needed

releases/stable-branches-and-release-cycle.md
SteveMattar: +1 hotfixes or critical bugs

releases/stable-branches-and-release-cycle.md
SteveMattar: I agree as we do not expect such changes so often
so a patch would increment the z-stream number

releases/stable-branches-and-release-cycle.md
skitt: I rather like the idea of preparing the release in the branch, and creating the branch in preparation for rc0. That way the devel branch is always pinned to devel Shipyard in particular (we’d still need to bump dependencies in `go.mod`).

releases/stable-branches-and-release-cycle.md
skitt: I’m not sure we have enough input from downstream users to know, but it would be nice to set expectations.

releases/stable-branches-and-release-cycle.md
skitt: I understand the `release-0.9` images as being “branch” images, _i.e._ updated to always point to the tip of the corresponding branch. _Release_ images would be tagged as they are currently, 0.9.0 etc.

releases/stable-branches-and-release-cycle.md
skitt: Shouldn’t we create the branch when we start preparing the release? (_I.e._ pin dependencies etc.)

releases/stable-branches-and-release-cycle.md
```diff
@@ -0,0 +1,79 @@
...
+### Reducing the cadence of minor releases
+
+Based on the previous ideas the existing minor version cadence doesn't seem realistic
+for maintaining stable branches. We don't want fake stable branches where patches
+are never backported.
+
+With the new cadence after each sprint we would produce a milestone release,
+for example
+
+* `0.9-milestone-1-rc0`, ... , `0.9-milestone-1`
```
dfarrell07, line 57: Maybe it would make things simpler to just call these RCs? That's effectivally what they are, snapshots that folks can use to test/preview the upcoming release. It would also keep our get.submariner.io logic untouched, just use the existing RC flag to get the latest milestone/RC.

releases/stable-branches-and-release-cycle.md
Jaanki: or sync with AMC

Jaanki (changes requested):

releases/stable-branches-and-release-cycle.md
```diff
@@ -0,0 +1,79 @@
...
+to make this possible starting with 0.8.
+
+## Proposal
+
+### Alignment of branch -> image releases
+
+We produce currently `<image>:devel` for every master push, along with `<image>:x.y.z` for
+the releases.
+
+Changing the `master` branch name to `devel` would make this homogeneous so our image releases
```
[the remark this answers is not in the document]
dfarrell07: There's a general push to move away from "master". The new default GH branch is "main", so I think folks will already be fairly primed for a different name.

https://github.com/submariner-io/submariner/issues/895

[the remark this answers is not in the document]
mangelajo: Yeah, I agree it's somehow confusing, but other repositories around the world are doing the same.

releases/stable-branches-and-release-cycle.md
mangelajo: Makes sense.

releases/stable-branches-and-release-cycle.md
mangelajo: Yet, but this mean, what to do with patches that are bugfixes to make the release possible.

releases/stable-branches-and-release-cycle.md
mangelajo: It's a compromised between how many backports/cherry-pics will need to do, and how clean ' devel' branch stays

mangelajo: I have clarified that we do this for the first rc (`rc0`) while it's up for discussion I think it would be a good approach to try.

releases/stable-branches-and-release-cycle.md
mangelajo: At the start probably it has to be more than critical bugs,

normally in other software, first it's general bugs, for some time, and then at some point, it becomes only hotfixes and critical bugs.

releases/stable-branches-and-release-cycle.md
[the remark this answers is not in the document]
mkolesnik: Currently we use `devel` to refer to the bleeding edge anyway, it'll make more sense if anything.

Also `master` is being retired due to politically correctness, GitHub specifically is switching to `main` but I'm not sure this change is ubiquitous.

releases/stable-branches-and-release-cycle.md
mkolesnik: We should create the branch whenever we start the release so rc0 makes sense, but as presented we'd have milestones before that which will pre-release without creating a branch for them.

releases/stable-branches-and-release-cycle.md
mangelajo: No, it should not do that.

The for doing that we should make a release explicitly, it would be a stable release 0.9.1 once we consider that release-0.9 has all the right patches.

releases/stable-branches-and-release-cycle.md
mkolesnik: We can separate the procedure from the policy, this section deals with the procedure.

Also I think if there's demand for a fix on an actively maintained branch then we'll need to cherry pick down into it (though the latter stable branches).

releases/stable-branches-and-release-cycle.md
mangelajo: The problem is that sometimes a new stable version will need multiple patches. Or sometimes we will want to control the cadence at which new patch releases are created.

mkolesnik: > I understand the `release-0.9` images as being “branch” images, _i.e._ updated to always point to the tip of the corresponding branch. _Release_ images would be tagged as they are currently, 0.9.0 etc.

Yes that's the gist, the `release-0.9` image in this example is the "latest stable" but not a "released" image, the "released" image will be tagged as always e.g. `0.9.3` or whatever the version will be.

releases/stable-branches-and-release-cycle.md
mangelajo: Not a bad idea but it conflicts with some clatifications  I added above based on comments (like branching off right before `rc0` is created).

We can add logic to grab milestones to get.submariner.io like VERSION=milestone, would that be ok?

mkolesnik: That's not the purpose of RC though. RC is "release candidate" so in the end you're expected to release a GA.

Probably since it's milestones and not actual releases it'll make more sense to go with `pre0`, `pre1` etc till the actual milestone gets released (as a pre-release itself).
This way we're not implying an imminent release.

releases/stable-branches-and-release-cycle.md
mkolesnik: So we branch off at `rc0` then? As long as we don't branch along the milestones I think it makes sense.

releases/stable-branches-and-release-cycle.md
mangelajo: We could make exceptions at some point, but I believe that setting a fixed cadence is good for projects, users will know what to expect, when to expect, we will know how many stable branches we need to maintain.

But as you say, at the start it's probably good to allow freedom to create an additional release, as long as we are aware of the cost resulting of that.

releases/stable-branches-and-release-cycle.md
mangelajo: L-25 you mean, or something different?

releases/stable-branches-and-release-cycle.md
mangelajo: Whatever you guys think is better.

I believe branching near `ga` removes cherry-pick work... but complicates other stuff, like the pinning in go-mod. We can try any of those ways and see how it goes.

skitt: Right, I don’t think we should branch for the milestones, as long as they are just snapshots of `devel`. My idea is that any non-`devel`-related changes should be on a branch; so for example we tag Shipyard, then change Admiral to pin the Dapper image on a branch leading to rc0, etc.

releases/stable-branches-and-release-cycle.md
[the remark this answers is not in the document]
jberkus: The change to "main" is universal for new projects.  So, yeah, OSS developers are already prepared for your primary branch to be something other than "master".

releases/stable-branches-and-release-cycle.md
skitt: They still need to go through the main branch. We ran into lots of problems before we enforced this rule in ODL: in quite a few instances, fixes were made on release branches, but forgotten in the main branch, so the next release failed again for all the same reasons...

releases/stable-branches-and-release-cycle.md
mkolesnik: I think it's best to explicitly specify the policy then, for instance we can say the latest stable gets all butgfixes and the stable n-1 gets just critical fixes.

releases/stable-branches-and-release-cycle.md
mkolesnik: No I mean that specifically each stable branch will be pinned to the corresponding shipyard version, but we won't be pinning to a fixed version.

So for example, `release-0.8` branch on `submariner` will be pinned to `shipyard-dapper-base:release-0.8` and won't be changed when releasing (as it makes less sense if we're already on a stable branch).

releases/stable-branches-and-release-cycle.md
mkolesnik: So probably `rc` should be the point at which we branch (which is not a stage during a milestone).

That's also why I suggested not to use `rc` terminology here - https://github.com/submariner-io/enhancements/pull/20/files#r545148130

releases/stable-branches-and-release-cycle.md
skitt: Agreed, we also define (or point to our definition elsewhere) of what constitutes a critical bugfix. It might be worth mentioning here too what we want to do regarding security issues; for example, we might want to allow no-change bumps to rebuilds container images to get fixed dependencies in our container images.

mangelajo: What do you mean by "allow no-change bumps to rebuilds container images to get fixed dependencies in our container images."

skitt: I imagine there will be circumstances where one of the packages we use in our images, or even the base image itself, needs to be upgraded because of a security issue or critical bug. When that happens, we’ll want to rebuild our images and release them, with a new version number, even if there are no changes in our code — so the code for, say, 0.8.2 and 0.8.3 would be identical, but the container images would be different and include important fixes.

mangelajo: Ah, makes sense. Thanks for the clarification.

releases/stable-branches-and-release-cycle.md
mangelajo: Yes, but in this paragraph I'm talking about the milestones, right?, on the milestones we don't pin. On the RC's yes. I'll clarify that.

mangelajo: Ah, I think I see what you mean, but go.mod can't be pointed to "release branches", only to vx.x.x tags, or git shas..

mkolesnik: I'm talking about the 'Dockerfile.dapper' files only, not 'go.mod'

mangelajo: > I'm talking about the 'Dockerfile.dapper' files only, not 'go.mod'

Ack, still not understanding it :), when you have the time specific text changes would help me I guess ':D sorry, I'm a bit thick today...

releases/stable-branches-and-release-cycle.md
```diff
@@ -0,0 +1,113 @@
...
+`release-0.9` <-- `devel` <-- `bug-fix patch`
+
+*No-change version bumps* will be done if a base image has been updated
+due to a security bugfix, our code would remain the same, but our versions
+would be bumped to pickup the new images.
+
+#### Previous stable branch (N-1)
+
+The only patches that should be submitted to the previous stable branch (N-1)
+are critical bugfixes, being the definition fo critical bugfix: security
```
sridhargaddam, line 53:
```suggestion
are critical bugfixes, being the definition of critical bugfix: security
```

releases/stable-branches-and-release-cycle.md
```diff
@@ -0,0 +1,113 @@
...
+would be bumped to pickup the new images.
+
+#### Releasing a stable branch
+
+When a patch is merged to a stable branch the images for such branch will be updated, for
+example `quay.io/submariner/submariner:release-x.y`.
+
+This image allows for testing a set of bugfix patches, sometimes multiple patches are expected
+to be required for one fix. Once the release brach is ready we should create a new x.x.**x**
+patch  release via the release repository.
```
sridhargaddam, line 73:
```suggestion
to be required for one fix. Once the release branch is ready we should create a new x.y.**z**
patch release via the release repository.
```

releases/stable-branches-and-release-cycle.md
```diff
@@ -0,0 +1,113 @@
...
+
+### How long is a stable branch maintained
+
+Initially stable branches (N) will be maintained until the next stable
```
sridhargaddam, line 77: Is 'N' a single Stable Branch or a number of stable branches?

mangelajo: ohh it's quite unclear, I meant current stable branch = N, previous = N-1 let me fix it.

mangelajo:
@sridhargaddam addressed the comments, I hope it looks better now.

sridhargaddam (approved):

releases/stable-branches-and-release-cycle.md
```diff
@@ -0,0 +1,113 @@
...
+would be bumped to pickup the new images.
+
+#### Releasing a stable branch
+
+When a patch is merged to a stable branch the images for such branch will be updated, for
+example `quay.io/submariner/submariner:release-x.y`.
+
+This image allows for testing a set of bugfix patches, sometimes multiple patches are expected
+to be required for one fix. Once the release branch is ready we should create a new x.y.**z**
+patch  release via the release repository.
```
nyechiel, line 73:
```suggestion
patch release via the release repository.
```

releases/stable-branches-and-release-cycle.md
```diff
@@ -0,0 +1,113 @@
...
+
+### How long is a stable branch maintained
+
+The current stable branch (N) will be maintained for with normal bugfixes until
```
nyechiel, line 77:
```suggestion
The current stable branch (N) will be maintained for normal bugfixes until
```

releases/stable-branches-and-release-cycle.md
```diff
@@ -0,0 +1,113 @@
...
+the next stable version is released. The previous stable branch (N-1) will be
+maintained only for critical bugfixes, however this can change in the future if
+we introduce the concept of long term support releases.
+
+### Reducing the cadence of minor releases
+
+Based on the previous ideas the existing minor version cadence doesn't seem realistic
```
nyechiel, line 84:
```suggestion
Based on the previous ideas, the existing minor version cadence doesn't seem realistic
```

releases/stable-branches-and-release-cycle.md
```diff
@@ -0,0 +1,113 @@
...
+for maintaining stable branches. We don't want fake stable branches where patches
```
nyechiel, line 85:
```suggestion
for maintaining stable branches. We don't want "fake" stable branches where patches
```

releases/stable-branches-and-release-cycle.md
```diff
@@ -0,0 +1,113 @@
...
+are never backported.
+
+With the new cadence after each sprint we would produce a milestone release,
```
nyechiel, line 88:
```suggestion
With the new cadence, after each sprint we would produce a milestone release,
```

releases/stable-branches-and-release-cycle.md
```diff
@@ -0,0 +1,113 @@
...
+
+* `0.9-milestone-1-rc0`, ... , `0.9-milestone-1`
+* `0.9-milestone-2-rc0`, ... , `0.9-milestone-2`
+* `0.9-milestone-3-rc0`, ... , `0.9-milestone-3`
+* `0.9-milestone-4-rc0`, ... , `0.9-milestone-4`
+
+In this case we won't do pinning of shipyard base image, but we will release and update
+the admiral & shipyard `go.mod` git references, without pinning shipyard.
+
+After a number of milestones instead we would follow the usual pattern for a release:
```
nyechiel, line 99: Should we define the number of miltestones?

releases/stable-branches-and-release-cycle.md
```diff
@@ -0,0 +1,113 @@
...
+
+* `0.9.0-rc0` along with its stable branch `release-0.9` as described
+  previously, ..., `0.9.0`
+
+The actual cadence and release times is TBD by the community, it could be aligned with
```
nyechiel, line 104: In addition to that, I would add something like "New versions of Submariner are released based on completion of feature work that has been scheduled for the release" to make it clear that we plan feature-based releases and not time-based.

releases/stable-branches-and-release-cycle.md
```diff
@@ -0,0 +1,113 @@
...
+
+Today the Submariner project releases one minor version a month without capacity
+for stable branches or bugfix releases. In the following proposal we define a plan
+to make this possible starting with 0.8.
+
+## Proposal
+
+### Alignment of branch -> image releases
+
+We produce currently `<image>:devel` for every master push, along with `<image>:x.y.z` for
```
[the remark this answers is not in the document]
mangelajo: Done

releases/stable-branches-and-release-cycle.md
```diff
@@ -0,0 +1,113 @@
...
+the releases.
+
+Changing the `master` branch name to `devel` would make this homogeneous so our image releases
+are simply `<image>:<branch-or-tag>`.
+
+### Creating one stable branch for every minor release
+
+Approaching the first release candidate `rc0` of every minor or major version a new stable
```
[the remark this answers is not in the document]
mangelajo: Done.

releases/stable-branches-and-release-cycle.md
```diff
@@ -0,0 +1,113 @@
...
+
+*No-change version bumps* will be done if a base image has been updated
+due to a security bugfix, our code would remain the same, but our versions
+would be bumped to pickup the new images.
+
+#### Previous stable branch (N-1)
+
+The only patches that should be submitted to the previous stable branch (N-1)
+are critical bugfixes, being the definition of critical bugfix: security
+related fixes, and CVEs.
```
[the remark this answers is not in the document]
mangelajo: Right, it's intended, in this case we are talking of N-1 previous stable branch.

That would be now 0.7.0 (only security bug fixes)
Then 0.8.0 (all fixes)
0.9.0 being developed.

At least we support two stable branches, and we scope the work we will do. May be it's too much promising to maintain 0.8.0 later with security bugfixes.

releases/stable-branches-and-release-cycle.md
mangelajo: Done

releases/stable-branches-and-release-cycle.md
mangelajo: oops. Done

releases/stable-branches-and-release-cycle.md
mangelajo: Done.

releases/stable-branches-and-release-cycle.md
mangelajo: Done

releases/stable-branches-and-release-cycle.md
mangelajo: I propose we leave them in free form for now. No strings attached? :-) Then we can choose when it's the best time to release.

releases/stable-branches-and-release-cycle.md
mangelajo: I feel that contradicts a little bit with the existing paragraph. Either we align with the cadence of other project, or we do feature based releases.

How could we phrase it?

tpantelis (approved):

releases/stable-branches-and-release-cycle.md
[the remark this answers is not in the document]
nyechiel: Oh, you are correct.

releases/stable-branches-and-release-cycle.md
nyechiel: Fine with me.

releases/stable-branches-and-release-cycle.md
nyechiel: Maybe it's out of scope for this specific proposal anyway. Let's discuss this in the one of the upcoming upstream calls.

nyechiel (approved):

skitt (approved):

releases/stable-branches-and-release-cycle.md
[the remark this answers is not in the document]
mkolesnik: > At least we support two stable branches, and we scope the work we will do. May be it's too much promising to maintain 0.8.0 later with security bugfixes.

To me it sounds like a reasonable policy, and I think it's widely accepted in open source projects.
We can always change the policy later (even expand it) depending on the bandwidth we have.

releases/stable-branches-and-release-cycle.md
mkolesnik: The proposal is already quite open ended so I think the current text is fine.

mkolesnik (approved):

PR #23, opened 2021-02-05
Enhancement to support fine-grained connectivity

aswinsuryan:
Signed-off-by: Aswin Surayanarayanan <asuryana@redhat.com>


submariner/fine-grained-connectivity.md
```diff
@@ -0,0 +1,86 @@
...
+        ClusterCIDR []string `json:"cluster_cidr"`
+        GlobalCIDR  []string `json:"global_cidr"`
+}
+```
+
+#### Submariner Broker
+
+There will be a single broker running in one of the participating clusters. The clusters will specify
+the Clusterset it wants to be a part of when it joins the broker. A cluster can be a part of multiple Clustersets.
+The broker ensures that resources are only synced within a Clusterset.
```
tpantelis, line 58: A broker is basically a namespace hosted on an API server. I think it makes sense for each cluster set to have its own broker namespace so they don't conflict so, in terms of the broker, a cluster set equates to a namespace. So logically there would be multiple brokers, one per cluster set. Otherwise, with a single broker namespace, we'd need some other way to route resources to the correct clusters (probably annotate the resources with the cluster set ID). The brokers could be hosted on one API server or multiple, that would be up to the admin.

submariner/fine-grained-connectivity.md
```diff
@@ -0,0 +1,86 @@
...
+#### GlobalNet
+
+Global net currently allocates CIDR per cluster. So across Clustersets each cluster will have non-overlapping CIDR.
+
+## Alternatives
+
+One broker per Clusterset: With multiple brokers either we need multiple Submariner components each having its own
+namespace and talk to a specific broker or we need to have a submariner component talking to multiple brokers.
+Either option requires changes across Submariner and Lighthouse. A significant change will be required in the
+submariner-operator and deployment as well.
```
tpantelis, line 103: Either way, there would be broker and Submariner components per cluster set, whether they run in the same process/pod or separate ones. Of course there's pros and cons to each,.

tpantelis (commented):
This proposal mainly covers multiple cluster sets in general. Whether we use multiple cluster sets to implement hub-spoke/client-server connectivity semantics is another story. One issue is a cluster being part of multiple cluster sets which violates the current u/s KEP. Another issue would be the complexity overhead for the admin to manage multiple cluster sets which could be dozens or hundreds per client-server domain.

submariner/fine-grained-connectivity.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Summary
+
+Submariner currently provisions full mesh topology by default between the participating clusters.
+For some use-cases, we need fine-grained connectivity options. This proposal covers only the topology
+aspect of connectivity.
+
+This proposal aims to address the below use-case
+
+Hub-Spoke Topology -  Here only the client can talk to the server but not to the other clients to
```
sridhargaddam: Can you please clarify which is Client and which is Server component in the Hub-Spoke topology. Also, include notes if Hub can talk to Spoke or not to avoid ambiguation.

submariner/fine-grained-connectivity.md
```diff
@@ -0,0 +1,86 @@
...
+which the server is connected to.
+
+To achieve each hub and spoke will be grouped together into an isolated group of clusters called Clusterset.
+There will be a full mesh topology within the Clusterset, though in this use case it will be just two clusters
+in the Clusterset. A cluster can be a part of multiple Clusterset. So the hub can be a part of multiple
+Clustersets, where it connects with a spoke in each Clusterset
```
sridhargaddam: Can a Spoke act like a Hub (in a different ClusterSet) for additional Spokes in this architecture?

submariner/fine-grained-connectivity.md
```diff
@@ -0,0 +1,86 @@
...
+Submariner currently provisions full mesh topology by default between the participating clusters.
+For some use-cases, we need fine-grained connectivity options. This proposal covers only the topology
+aspect of connectivity.
+
+This proposal aims to address the below use-case
+
+Hub-Spoke Topology -  Here only the client can talk to the server but not to the other clients to
+which the server is connected to.
+
+To achieve each hub and spoke will be grouped together into an isolated group of clusters called Clusterset.
```
sridhargaddam:
```suggestion
To achieve each hub and spoke will be grouped together into an isolated group of clusters called Clusterset.
A Hub cluster which is part of multiple Clustersets does not perform routing between the Spokes in different Clustersets.
```

submariner/fine-grained-connectivity.md
```diff
@@ -0,0 +1,86 @@
...
+
+The broker ensures that the ServiceImports are synced only within the Clustersets. So a cluster should be a
+part of the Clusterset from which the service was exported to receive the service import.  But the same
+service in the same namespace across Clustersets would be considered an instance the same service.
+
+#### Dataplane Connectivity
+
+Similar to service imports endpoint objects are synced only within the Clustersets. A full mesh of tunnels
+will be created within the members of the Clustersets. If a cluster is a part of two Clusterset multiple
+tunnels will be created.
```
sridhargaddam:
```suggestion
tunnels will be created, but will not forward traffic from one Clusterset to other.
```

submariner/fine-grained-connectivity.md
```diff
@@ -0,0 +1,86 @@
...
+
+#### GlobalNet
+
+Global net currently allocates CIDR per cluster. So across Clustersets each cluster will have non-overlapping CIDR.
```
sridhargaddam:
```suggestion
Globalnet currently allocates CIDR per cluster. So across Clustersets each cluster will have non-overlapping CIDR.
```

submariner/fine-grained-connectivity.md
sridhargaddam: Right. But there is a problem if multiple Submariner Components are run (even if they are in different namespace) since the Submariner Pods run with HostNetworking, they modify the routingRules/ipTables/Tunnels etc on the Host There will be race conditions if multiple Submariner Pods are run. So IMHO its better to avoid that path and use single instance of Submariner Pods even if the Cluster is part of multiple Clustersets and use Broker for intelligently distributing endpoint/clusters based on the Clusterset.

sridhargaddam (changes requested):
Thanks for the PR @aswinsuryan . Some comments inline.

submariner/fine-grained-connectivity.md
tpantelis: As mentioned elsewhere, a cluster being part of multiple Clustersets is going to be a sticking point with the u/s proposal. Perhaps we should not try to align with that for this use case and think in terms of a "_Broker domain_" (ie a broker namespace).  Maybe a cluster can only be part of one Clusterset but it can be part of multiple Broker domains. I think if we focus on the latter then we can avoid Clusterset, at least for now, since it's just an initial KEP right now.

submariner/fine-grained-connectivity.md
```diff
@@ -0,0 +1,86 @@
...
+## Proposal
+
+### Terminology
+
+#### Clusterset
+
+A Clusterset is a group of clusters that forms a trust boundary. The design of Clusterset can be aligned
+to the MCS KEP. More details are available in the below link.
+
+[Cluster-Id KEP]: http://www.reddit.comhttps://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/2149-clusterid
```
nyechiel:
```suggestion
[Cluster-Id KEP]: https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/2149-clusterid
```

submariner/fine-grained-connectivity.md
```diff
@@ -0,0 +1,86 @@
...
+Clustersets, where it connects with a spoke in each Clusterset
+
+## Proposal
+
+### Terminology
+
+#### Clusterset
+
+A Clusterset is a group of clusters that forms a trust boundary. The design of Clusterset can be aligned
+to the MCS KEP. More details are available in the below link.
```
nyechiel: You mention the MCS KEP (https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api) but then link to the Cluster ID KEP?

submariner/fine-grained-connectivity.md
nyechiel: I find it confusing that we include design details ("each hub and spoke will be grouped together into an isolated group of clusters called Clusterset") as part of the use-case definition. I think it's better to only define the problems/requirements in this section, and explain the proposed approach under the Proposal section below.

submariner/fine-grained-connectivity.md
```diff
@@ -0,0 +1,86 @@
...
+
+#### Service Discovery
+
+The broker ensures that the ServiceImports are synced only within the Clustersets. So a cluster should be a
+part of the Clusterset from which the service was exported to receive the service import.  But the same
+service in the same namespace across Clustersets would be considered an instance the same service.
+
+#### Dataplane Connectivity
+
+Similar to service imports endpoint objects are synced only within the Clustersets. A full mesh of tunnels
```
nyechiel:
```suggestion
Similar to ServiceImports, endpoint objects are synced only within the Clustersets. A full mesh of tunnels
```

submariner/fine-grained-connectivity.md
aswinsuryan: The hub and spoke will be a special case of clustersets were only two clusters are present,  no specific server will be marked as a hub or a spoke with this proposal. So they can talk in both ways. A cluster which is a part of multiple clustersset will be a hub logically but it is not explicit. I can add the details based on how we conclude on other comments.

submariner/fine-grained-connectivity.md
sridhargaddam: I agree that Broker is just a namespace today. Inorder to support multiple Brokers for each Clusterset, we will have to modify ClusterCRD to include Broker URL/Credentials for each ClusterSet. This way, we will have Single Submariner Components (which will avoid race conditions while applying rules on the nodes) and will be able to support multiple Brokers. But this will be a bit intrusive change and also LH might need awareness of Clusetset (and the associated Broker in the ClusterSet).

tpantelis: yes each component that syncs to/from the broker would need to be aware of each broker URL, broker namespace and local namespace. Of course today we pass that via env vars but could be `ConfigMap`(s). It doesn't seem like submariner would need to be aware of the concept of a Clusterset - that seems like a higher-level logical organization.

submariner/fine-grained-connectivity.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Summary
+
+Submariner currently provisions full mesh topology by default between the participating clusters.
+For some use-cases, we need fine-grained connectivity options. This proposal covers only the topology
+aspect of connectivity.
+
+This proposal aims to address the below use-case
+
+Hub-Spoke Topology -  Here only the client can talk to the server but not to the other clients to
```
mkolesnik: Lets not call it hub and spoke, this term means that all nodes can communicate.
https://en.wikipedia.org/wiki/Spoke%E2%80%93hub_distribution_paradigm

Perhaps we can call it just "central cluster" or something and explain the use case.

submariner/fine-grained-connectivity.md
aswinsuryan: @tpantelis  As discussed we can have singe broker and single broker n/s and use the clusterset/colorcode in the clusterCRD to identify how the cluster needs to be grouped together. Lighthouse will have to add the clusterset/colorcode as a label in ServiceImport. So the high-level design will remain more or less the same as described in the proposal. I will add more details about the design, once we decide to go ahead with this proposal.

submariner/fine-grained-connectivity.md
aswinsuryan: Done.

submariner/fine-grained-connectivity.md
aswinsuryan: Yes spoke can be a hub too. Submariner does not differentiate between hub and spoke. It is based on the topology user creates certain clusters will be hub.

submariner/fine-grained-connectivity.md
aswinsuryan: Changed the name.

submariner/fine-grained-connectivity.md
aswinsuryan: As discussed we will be going with one broker and one broker n/s for now.

aswinsuryan:
Closing in favour of https://github.com/submariner-io/enhancements/pull/4

PR #29, opened 2021-05-06
Add Vxlan Cable Driver

### LINKED ISSUE #1313: Write an enhancement proposal for un-encrypted vxlan tunnels
Write an enhancement proposal for un-encrypted vxlan tunnels

**Why is this needed**:

This is to discuss the design details of vxlan tunnel.

aswinsuryan:
Fixes: submariner-io/submariner/issues/1313

Signed-off-by: Aswin Surayanarayanan <asuryana@redhat.com>


submariner/vxlan-tunnel.md
```diff
@@ -0,0 +1,34 @@
+# Encryption-less connections between clusters with VXLAN
+
+## Summary
+
+All current cable drivers involve encrypting content, which is great for privacy but involves some overhead. 
+On private connections, it would be useful to enable unencrypted connections, for example using IP-on-IP or VxLAN.
```
tpantelis:
```suggestion
On private connections, it would be useful to allow unencrypted connections, for example using IP-on-IP or VxLAN.
```

submariner/vxlan-tunnel.md
```diff
@@ -0,0 +1,34 @@
...
+
+This enhancement proposes to use VXLAN for un-encrypted connection between clusters. A new VXLAN interface "vxlan-tunnel"
+will be added to the active gateway node of each cluster and it will act as the VTEP. The traffic which is destined to other
+clusters will be forwarded via this interface to the appropriate clusters.
+
+## Design Details
+
+* A new interface "vxlan-tunnel" will be added 
+* The VXLAN interface will be one-to-many, that is a single interface will be used to connect to all the joined clusters. 
+* For the VXLAN VTEP IP, the prefix "241" and will use the rest of the blocks from the endpoint IP
```
tpantelis: Can you clarify this?

submariner/vxlan-tunnel.md
```diff
@@ -0,0 +1,34 @@
...
+* The remote clusters IP (either public or private ) will be added to the Forwarding Database
```
tpantelis:
```suggestion
* The remote cluster endpoint IPs (either public or private ) will be added to the Forwarding Database
```

submariner/vxlan-tunnel.md
```diff
@@ -0,0 +1,34 @@
...
+* The routes will be added to forward all the Service CIDR and POD CIDR to the respective remote VTEP IP.
```
tpantelis:
```suggestion
* The routes will be added to forward all the Service and Pod CIDR traffic to the respective remote VTEP IP.
```

submariner/vxlan-tunnel.md
aswinsuryan: Added an example

submariner/vxlan-tunnel.md
```diff
@@ -0,0 +1,34 @@
+# Encryption-less connections between clusters with VXLAN
+
+## Summary
+
+All current cable drivers involve encrypting content, which is great for privacy but involves some overhead.
+On private connections, it would be useful to allow unencrypted connections, for example using IP-on-IP or VxLAN.
```
nyechiel:
```suggestion
On private connections, it would be useful to allow unencrypted connections, for example using IP-on-IP or VXLAN.
```

submariner/vxlan-tunnel.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Proposal
+
+This enhancement proposes to use VXLAN for un-encrypted connection between clusters. A new VXLAN interface "vxlan-tunnel"
+will be added to the active gateway node of each cluster and it will act as the VTEP. The traffic which is destined to other
+clusters will be forwarded via this interface to the appropriate clusters.
+
+## Design Details
+
+* A new interface "vxlan-tunnel" will be added
```
nyechiel:
```suggestion
* A new interface "vxlan-tunnel" will be added.
```

submariner/vxlan-tunnel.md
```diff
@@ -0,0 +1,34 @@
...
+* The VXLAN interface will be one-to-many, that is a single interface will be used to connect to all the joined clusters.
+* For the VXLAN VTEP IP, the prefix "241" and will use the rest of the blocks from the endpoint private IP address. For example,
```
nyechiel:
```suggestion
* For the VXLAN VTEP IP, the prefix "241" will be used, followed by the rest of the blocks from the endpoint private IP address. For example,
```

submariner/vxlan-tunnel.md
```diff
@@ -0,0 +1,34 @@
...
+if the private IP address is 10.2.96.1 the VTEP IP would be 241.2.96.1.
+* The remote cluster endpoint IPs (either public or private ) will be added to the Forwarding Database
```
nyechiel:
```suggestion
* The remote cluster endpoint IPs (either public or private) will be added to the Forwarding Database.
```

submariner/vxlan-tunnel.md
```diff
@@ -0,0 +1,34 @@
...
+* The routes will be added to forward all the Service and Pod CIDR traffic to the respective remote VTEP IP.
+* Table 150 shall be used to add the routes, which is otherwise used for host network traffic routes. We do not require any
+extra rules to handle host network traffic for the VXLAN driver.
+* The IPSEC_NATT_PORT used port shall be used as the VTEP port. The parameter name shall be changed to TUNNEL_PORT
```
nyechiel:
```suggestion
* The IPSEC_NATT_PORT used port shall be used as the VTEP port. The parameter name shall be changed to TUNNEL_PORT.
```

submariner/vxlan-tunnel.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Work items
+
+* [New cable driver](https://github.com/submariner-io/submariner/issues/1149)
+* [CI](https://github.com/submariner-io/submariner/issues/1148)
+* [Deployment(Operator, subctl)](https://github.com/submariner-io/submariner-operator/issues/1101)
+* [Deployment (helm)](https://github.com/submariner-io/submariner-charts/issues/116)
+* [Metrics](https://github.com/submariner-io/submariner/issues/1150)
+* [Docs](https://github.com/submariner-io/submariner-website/issues/450)
+* [cloud-prepare](https://github.com/submariner-io/submariner/issues/1319)
```
nyechiel: IIUC, there is nothing to change in cloud-prepare. The implementation there is already abstracted and not specific to IPsec, i.e:

{Port: nattPort, Protocol: "udp"},
{Port: natDiscoveryPort, Protocol: "udp"},

submariner/vxlan-tunnel.md
nyechiel: This is both in Submariner and in the Operator, right? Can you please identify other references to IPsec in the code that should be abstracted? Even `subctl` says things like 

cmd.Flags().BoolVar(&natTraversal, "natt", true, "enable NAT traversal for IPsec")
cmd.Flags().BoolVar(&disableNat, "disable-nat", false, "disable NAT for IPsec")

nyechiel (commented):
Couple of nits and one follow up question, thanks @aswinsuryan!

submariner/vxlan-tunnel.md
aswinsuryan: Added the repos where this change will be required.

submariner/vxlan-tunnel.md
aswinsuryan: Removed from to do

submariner/vxlan-tunnel.md
```diff
@@ -0,0 +1,34 @@
...
+
+* A new interface "vxlan-tunnel" will be added.
+* The VXLAN interface will be one-to-many, that is a single interface will be used to connect to all the joined clusters.
+* For the VXLAN VTEP IP, the prefix "241" will be used, followed by the rest of the blocks from the endpoint private IP address. For example,
+if the private IP address is 10.2.96.1 the VTEP IP would be 241.2.96.1.
+* The remote cluster endpoint IPs (either public or private) will be added to the Forwarding Database.
+* The routes will be added to forward all the Service and Pod CIDR traffic to the respective remote VTEP IP.
+* Table 150 shall be used to add the routes, which is otherwise used for host network traffic routes. We do not require any
+extra rules to handle host network traffic for the VXLAN driver.
+* The IPSEC_NATT_PORT used port shall be used as the VTEP port. The parameter name shall be changed to TUNNEL_PORT. This requires changes
```
mangelajo: we can set it now as gateway.submariner.io/udp-port label. May be we can call it just GATEWAY_UDP_PORT
 for uniformity?, also in subctl join --gateway-udp-port ? or

An important problem with vxlan, at least in the past is that src & dst port should be equal, so it doesn't work with non-uniform udp ports.  Unless the kernel has changed (it was a kernel limitation). We need to verify if that works, and otherwise, we probably need to log an error when a connection is attempted between endpoints not sharing the same port number.

submariner/vxlan-tunnel.md
```diff
@@ -0,0 +1,34 @@
...
+
+This enhancement proposes to use VXLAN for un-encrypted connection between clusters. A new VXLAN interface "vxlan-tunnel"
+will be added to the active gateway node of each cluster and it will act as the VTEP. The traffic which is destined to other
+clusters will be forwarded via this interface to the appropriate clusters.
+
+## Design Details
+
+* A new interface "vxlan-tunnel" will be added.
+* The VXLAN interface will be one-to-many, that is a single interface will be used to connect to all the joined clusters.
+* For the VXLAN VTEP IP, the prefix "241" will be used, followed by the rest of the blocks from the endpoint private IP address. For example,
```
mangelajo, line 18: We keep adding to the reserved 240.0.0.0/4 class E range of IPv4, I wonder if there's a better solution to that.
we are at risk that such range would eventually used for something and submariner would sink :P

@sridhargaddam Do we know if it would be possible to assign IPv6 addresses to the gateways (we could use LLAs) and use those as destination of IPv4 networks?, I know it's quite bizarre, but, technically it may be possible, as in the end, we only need the destination MAC address for routing, but may be the kernel won't support this.

mangelajo (changes requested):

submariner/vxlan-tunnel.md
aswinsuryan: Done. I have added item in the EPIC to validate the non-uniform port case.
https://github.com/submariner-io/submariner/issues/674

submariner/vxlan-tunnel.md
sridhargaddam: @mangelajo I'll have to check if it works. I've seen some conversations in the past where people completely disable IPv6 in the underlay, so even if it works, for such clusters there could be an issue (though a remote possibility). Its worth spending time on this and trying to understand if it works.

In the meantime, as you know, currently we have a hardcoded 240.0.0.0/8 range. Shall we expose that as an environment variable (even if its hidden to the users) so that we can customize it, if necessary, just to mitigate situations where 240.0.0.0/8 is not allowed for some reason.

mangelajo: I think it would be ok to make IPv6 stack a requirement if it worked (i'm not sure). The risk of the 240/4 range is dangerous ':)

An env var could help. but to which range would they migrate? 

We probably should book some time in the future (0.11?) to investigate alternatives to this. But for now 241.x.x.x should be fine I think.

submariner/vxlan-tunnel.md
```diff
@@ -0,0 +1,34 @@
...
+if the private IP address is 10.2.96.1 the VTEP IP would be 241.2.96.1.
+* The remote cluster endpoint IPs (either public or private) will be added to the Forwarding Database.
+* The routes will be added to forward all the Service and Pod CIDR traffic to the respective remote VTEP IP.
+* Table 150 shall be used to add the routes, which is otherwise used for host network traffic routes. We do not require any
```
mangelajo: Please note that in OVN we use:

149 for remote-clusters -> local cluster traffic steering
150 for host net to remote cluster steering (I believe..)

mangelajo (approved):

submariner/vxlan-tunnel.md
sridhargaddam: > An env var could help. but to which range would they migrate?

Any other range in the Reserved Class E network. This is to get Submariner working instead of sinking :P 

>We probably should book some time in the future (0.11?) to investigate alternatives to this. But for now 241.x.x.x should be fine I think.

If we make it configurable, we can update the CIDR to 241.0.0.0/8 if required. But yes, we should look for alternatives and find a better solution.

submariner/vxlan-tunnel.md
aswinsuryan: @mangelajo I used 150 since host-networking routes was using it, should that be changed?

submariner/vxlan-tunnel.md
aswinsuryan: Raised https://github.com/submariner-io/submariner/issues/1321 to track.

submariner/vxlan-tunnel.md
aswinsuryan: Changed to 100

nyechiel (approved):
Thanks @aswinsuryan

mangelajo (approved):

PR #33, opened 2021-06-30
Enhancement proposal for transparent multicluster and namespace mapping

mangelajo:

### DIFF
--- a/lighthouse/transparent-multicluster-and-namespace-mapping-support.md
+++ b/lighthouse/transparent-multicluster-and-namespace-mapping-support.md
@@ -1,0 +1,84 @@
+# Transparent multicluster support
+
+## Summary
+
+Lighthouse is a component from Submariner which can be used independently of the dataplane components,
+and exchanges information in the form of ServiceImport and EndpointSlice CR/CRDs over the broker.
+
+Lighthouse implements a CoreDNS plugin to perform dns resolution for clusterset services.
+
+Lighthouse implements the
+[MCS API](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api)
+with some additional extensions.
+
+In terms of DNS resolution, the MCS API specifies that multicluster services can be resolved with the following
+FQDN format:
+
+* $service-name.$namespace.svc.clusterset.local
+
+In the context of [transparent multicluster](https://github.com/kcp-dev/kcp/tree/main/docs/investigations),
+two extra features are necessary:
+
+* The ability to resolve multicluster services via $service-name.$namespace or $service-name, in a way
+  that a pod running on a different cluster would resolve such service as if it was a local service,
+  hence removing the need to modify any deployment logic for the application.
+
+* The ability to map namespaces, removing the namespace sameness requirement. This is necessary to cover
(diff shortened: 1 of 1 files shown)


lighthouse/transparent-multicluster-and-namespace-mapping-support.md
```diff
@@ -0,0 +1,84 @@
...
+#### DNS resolution from Pods
+
+Pods from the namespace "foo" will resolve multicluster services in the following way:
+
+* $service will resolve as $service.bar.svc.clusterset.local
+* $service.foo will resolve as $service.bar.svc.clusterset.local
+* $service.foo.svc.clusterset.local will resolve as $service.bar.svc.clusterset.local
+
+This means that, when this mode is enabled, the lighthouse-coredns plugin will need to track
+a list of pod-ip to namespace, so DNS requests can be traced back to namespaces. This
```
smarterclayton, line 75: Source IP of the DNS request used to determine what to serve?  What happens for host network pods (may not be a use case that is required to work, just haven't thought about it)?  I do think host network pods wouldn't be part of the transparent multi-cluster use case by default (the bit where users don't know what cluster they're on)

mangelajo: That is correct, host network pods cannot be mapped this way. But I don't see another way of mapping them, 

An alternative to this specific implementation (long term) could be having the kubelet inject[1] the search paths in /etc/resolv.conf for clustersets, (before the local ones), but that would not solve the case of hostNetwork, the host /etc/resolv.conf is used, and there are no pod specific tweaks that we can make I believe.

[1] https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/network/dns/dns.go#L168

lighthouse/transparent-multicluster-and-namespace-mapping-support.md
```diff
@@ -0,0 +1,84 @@
...
+
+In terms of DNS resolution, the MCS API specifies that multicluster services can be resolved with the following
+FQDN format:
+
+* $service-name.$namespace.svc.clusterset.local
+
+In the context of [transparent multicluster](https://github.com/kcp-dev/kcp/tree/main/docs/investigations),
+two extra features are necessary:
+
+* The ability to resolve multicluster services via $service-name.$namespace or $service-name, in a way
```
mangelajo, line 22: comment from @smarterclayton on slack: we would also need to resolve $service-name.$namespace.svc.cluster.local as the multicluster service

lighthouse/transparent-multicluster-and-namespace-mapping-support.md
```diff
@@ -0,0 +1,84 @@
...
+create the right DNS entries for the OpenShift DNS operator. (needs investigation to make sure that we can
+properly fallback to the normal kubernetes resolution).
+
+### Namespace mapping
+
+Namespace mapping could be performed by annotating a namespace with the target clusterset namespace, for example.
+
+If we take this example, in the physical cluster A:
+
+   kubectl annotate namespace foo multicluster.kubernetes.io/namespace=bar
```
mangelajo, line 54: -> Better to have a CRD because that would be an indication of supported/not supported.

dfarrell07:
Closing for now, if this becomes a priority/someone can work on it we'll re-open.

PR #59, opened 2021-12-13
Broker Multi-tenancy enhancement added

davidohana:
Signed-off-by: David Ohana <david.ohana@ibm.com>

### DIFF
--- a/submariner-operator/broker-multi-tenancy.md
+++ b/submariner-operator/broker-multi-tenancy.md
@@ -1,0 +1,321 @@
+# Submariner Broker Multi-Tenancy Support
+
+Extending Submariner Broker to Support Multiple ClusterSets/Tenants.
+
+## Summary
+
+Currently, Submariner Broker supports a single ClusterSet (i.e., a set of
+clusters with a high degree of trust and sharing). This design document
+proposes an enhancement to support multiple disjoint ClusterSet’s (e.g.,
+each belonging to a different “tenant”) hosted on a single Broker cluster.
+ The design addresses multiplicity and isolation concerns.
+
+## Motivation
+
+In SaaS offerings, such as managed OpenShift, it may be desirable to
+minimize the cost of running a dedicated Broker cluster for each ClusterSet.
+When all clusters share a high level of trust, the Broker can be co-located
+with workloads and does not require a dedicated cluster. However, this level
+of trust is neither uncommon nor desirable in a multi-tenant environment.
+Furthermore, in such settings, the Submariner Broker cluster is owned
+and managed independently of the workload clusters.
+
+Submariner repository currently has the following related issues:
+
+- [submariner#1120](https://github.com/submariner-io/submariner/issues/1120):
+Deploy multiple ClusterSet's within the same "Submariner"
+- [submariner-operator#1040](https://github.com/submariner-io/submariner-operator/issues/1040):
+Support deployment within a custom namespace
+
+In addition, a recent PR which provides partial support was recently merged:
+
+- [submariner-operator PR 1682](https://github.com/submariner-io/submariner-operator/pull/1682)
+
+### Pitfalls of the current state
+
+The current design places all resources relating to a ClusterSet in a single,
+predefined, namespace on the Broker cluster. This works well for a single
+ClusterSet and users often choose to run independent Submariner Broker clusters
+for different ClusterSets. The use of a single predefined Broker namespace
+is limiting:
+
+1. The selected namespace name should be configurable to accommodate
+requirements of different use cases.
+2. It prevents Submariner from managing multiple independent ClusterSets in a
+single Broker cluster.
+
+In a SaaS environment, we can consider three distinct user types in the system,
+with different privileges:
+
+- **Offering Admin** owns and manages the Submariner Broker(s) cluster. They are
+responsible for creating a per ClusterSet configuration space.
+- **ClusterSet/Broker Admin** manages workload clusters belonging to a ClusterSet.
+They manage cluster membership in a given ClusterSet. Admins of different ClusterSets
+should be isolated from each other.
+- **Cluster/Service Admin** manages the import and export of services from/to
+workload clusters in the ClusterSet.
+
+The first two are somewhat conflated in the current Submariner design.
+
+### Goals
+
+- Allow usage of configurable Broker namespace names.
+- Enable multiple ClusterSets to coexist in the same Submariner Broker(s)
+cluster.
+- Provide isolation (via namespaces and RBAC) between objects of different
+ClusterSets to ensure users can’t modify configuration of another user’s ClusterSet.
+- Provide clearer distinction between Offering Admin and ClusterSet Admin,
+by allowing the latter to deploy a Broker with restricted namespace permission.
+- Provide backward compatibility for existing deployments (e.g.,
+explicit opt-in, defaulting to current Submariner behavior and settings by default).
+
+### Non-Goals
+
+- Support live migration of an existing Broker to support multi-tenancy.
+- Support for workload clusters joining more than one ClusterSet
+(in the case of multiple tenants in workload clusters, we could run a
+separate cluster control plane per tenant).
+- Any changes needed in runtime and data plane elements (e.g., metrics collection,
+ Lighthouse, etc.) to support configurable ClusterSet naming.
+- Support multi-tenancy in workload clusters.
+
+Note that non-goals are defined as such in the context of the currently proposed
+work and may be added in the future. They are optional for the current phase
+and may be implemented if straightforward.
+
+## Proposal
+
+The work will be carried out in stages:
+
+1. Make Broker namespace configurable (e.g. in CRDs and subctl) to
+allow multiple Broker namespaces to co-exist. The default namespace value
+can match the current fixed value to allow migration. The use of multiple
+or single namespace should be under global configuration to allow backwards
+compatibility. The change does not introduce new functionality (e.g.,
+isolation and protection) beyond configurable namespace names. This allows
+for multiple namespaces to exist concurrently, in a cooperative manner
+(i.e., discretionary access control by referencing the correct namespace).
+2. Support stricter isolation between namespaces. This step builds upon the
+previous and creates relevant service accounts, RBAC policies, etc., to ensure
+users can only access their designated Broker namespace (previous step assumed
+cooperative access).
+3. Separate Broker namespace creation and management to separate the Offering
+and ClusterSet Admin roles and privileges. This could be done via a Broker (full)
+ reconciliation loop (note: currently the broker is not active, but acts
+ as a storage namespace only). Changes should be made to minimize disruption to existing
+workflows (e.g., hide changes behind the existing `subctl` command line).
+
+The first and second steps are mostly complete by
+[submariner-operator PR 1682](https://github.com/submariner-io/submariner-operator/pull/1682).
+
+## Design Details
+
+### Overview of Proposed Changes
+
+- Add a new “central” controller/operator called Axon controller (may think
+of a better name.) to the Broker cluster. The controller runs in a new,
+singleton, Axon namespace. This controller is responsible for provisioning
+submariner tenants (i.e., a ClusterSet). The `subctl` CLI will be extended
+to support tenant management by adding the following commands (see details below):
+  - `subctl deploy axon-cp`
+  - `subctl create|delete|get|list tenant [tenant_id]`
+- Create the submariner CRDs when deploying the Axon operator, assuming that
+tenants (i.e., ClusterSet Admins) will not have permissions to create those.
+The `deploy broker` command will ensure the CRDs exist.
+- A submariner tenant is represented by a tenant object in the Axon namespace
+ and a dedicated tenant namespace. Each tenant has a string ID and the namespace
+ is some constant prefix + tenant ID.
+- A SA is created per tenant and a *partial* `broker-info.subm` file is generated
+ by `subctl` upon tenant creation, with credentials (client secret) to access
+ the tenant namespace only.
+- Tenant can be deleted by deleting the tenant object, and the Axon operator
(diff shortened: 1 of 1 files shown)


submariner-operator/broker-multi-tenancy.md
```diff
@@ -0,0 +1,321 @@
...
+should be isolated from each other.
+- **Cluster/Service Admin** manages the import and export of services from/to
+workload clusters in the ClusterSet.
+
+The first two are somewhat conflated in the current Submariner design.
+
+### Goals
+
+- Allow usage of configurable Broker namespace names.
+- Enable multiple ClusterSets to coexist in the same Submariner Broker(s)
```
mkolesnik, line 63: So I think we need to clearly define what's a ClusterSet, since if it's "all submariner resources in an arbitrary namespace" then the first goal should cover it already.

submariner-operator/broker-multi-tenancy.md
```diff
@@ -0,0 +1,321 @@
...
+cluster.
+- Provide isolation (via namespaces and RBAC) between objects of different
+ClusterSets to ensure users can’t modify configuration of another user’s ClusterSet.
```
mkolesnik, line 66: This is partially related to submariner but partially isn't, I think the goal should be that each submariner instance working against a single broker namespace can only access and influence objects in that broker namespace.
We can't enforce isolation beyond that, as it's out of scope for submariner.

submariner-operator/broker-multi-tenancy.md
```diff
@@ -0,0 +1,321 @@
...
+- Provide clearer distinction between Offering Admin and ClusterSet Admin,
+by allowing the latter to deploy a Broker with restricted namespace permission.
```
mkolesnik, line 68: I'm not sure what this means in the context of Submariner.
We use whatever access the given kubeconfig provides, there either is an ability to create certain resources in K8s or there isn't, but it's out of scope for Submariner.

submariner-operator/broker-multi-tenancy.md
```diff
@@ -0,0 +1,321 @@
...
+- Provide backward compatibility for existing deployments (e.g.,
+explicit opt-in, defaulting to current Submariner behavior and settings by default).
+
+### Non-Goals
+
+- Support live migration of an existing Broker to support multi-tenancy.
+- Support for workload clusters joining more than one ClusterSet
```
mkolesnik, line 75: I'm not sure what this means, you would have one cluster be a part of two cluster sets?
How would that work out?

submariner-operator/broker-multi-tenancy.md
```diff
@@ -0,0 +1,321 @@
...
+(in the case of multiple tenants in workload clusters, we could run a
+separate cluster control plane per tenant).
+- Any changes needed in runtime and data plane elements (e.g., metrics collection,
+ Lighthouse, etc.) to support configurable ClusterSet naming.
+- Support multi-tenancy in workload clusters.
```
mkolesnik, line 80: What does this mean?

submariner-operator/broker-multi-tenancy.md
```diff
@@ -0,0 +1,321 @@
...
+and may be implemented if straightforward.
+
+## Proposal
+
+The work will be carried out in stages:
+
+1. Make Broker namespace configurable (e.g. in CRDs and subctl) to
+allow multiple Broker namespaces to co-exist. The default namespace value
+can match the current fixed value to allow migration. The use of multiple
+or single namespace should be under global configuration to allow backwards
```
mkolesnik, line 93: Why not default to `submariner-k8s-broker` namespace (like now), and just use another namespace if requested?
I don't understand what "global configuration" is required here.

submariner-operator/broker-multi-tenancy.md
```diff
@@ -0,0 +1,321 @@
...
+compatibility. The change does not introduce new functionality (e.g.,
+isolation and protection) beyond configurable namespace names. This allows
+for multiple namespaces to exist concurrently, in a cooperative manner
+(i.e., discretionary access control by referencing the correct namespace).
+2. Support stricter isolation between namespaces. This step builds upon the
+previous and creates relevant service accounts, RBAC policies, etc., to ensure
```
mkolesnik, line 99: We already create SAs and Roles that are namespaced to the broker, is there anything else needed?

submariner-operator/broker-multi-tenancy.md
```diff
@@ -0,0 +1,321 @@
...
+
+- Add a new “central” controller/operator called Axon controller (may think
+of a better name.) to the Broker cluster. The controller runs in a new,
+singleton, Axon namespace. This controller is responsible for provisioning
+submariner tenants (i.e., a ClusterSet). The `subctl` CLI will be extended
+to support tenant management by adding the following commands (see details below):
+  - `subctl deploy axon-cp`
+  - `subctl create|delete|get|list tenant [tenant_id]`
+- Create the submariner CRDs when deploying the Axon operator, assuming that
+tenants (i.e., ClusterSet Admins) will not have permissions to create those.
```
mkolesnik, line 123: Honestly, I don't think this should be a part of Submariner core.

You could, however, add this as another service on top of Submariner (i.e. similar to OCM add on).

submariner-operator/broker-multi-tenancy.md
```diff
@@ -0,0 +1,321 @@
...
+- Tenant can be deleted by deleting the tenant object, and the Axon operator
+ will clean up its resources (i.e., remove namespace and all contained objects)
+- The `deploy broker` command of subctl will require some changes but will
+ also maintain backwards compatibility with the single broker operation mode:
+  - Add an optional `namespace` field to the `.subm` file, to identify the
+  tenant. If the field exists, it takes precedence over the `--namespace` CLI
+  argument. If neither is provided, uses the default broker namespace.
+  - Use the access credentials and the Broker URL that are already specified in
+  the `.subm` file to access the Broker, if they exist. Otherwise - fallback to
+  details from kubecontext.
```
mkolesnik, line 140: I don't believe we do that even now (fallback to kubecontext), why would we introduce this?

submariner-operator/broker-multi-tenancy.md
```diff
@@ -0,0 +1,321 @@
...
+tenants (i.e., ClusterSet Admins) will not have permissions to create those.
+The `deploy broker` command will ensure the CRDs exist.
+- A submariner tenant is represented by a tenant object in the Axon namespace
+ and a dedicated tenant namespace. Each tenant has a string ID and the namespace
+ is some constant prefix + tenant ID.
+- A SA is created per tenant and a *partial* `broker-info.subm` file is generated
+ by `subctl` upon tenant creation, with credentials (client secret) to access
+ the tenant namespace only.
+- Tenant can be deleted by deleting the tenant object, and the Axon operator
+ will clean up its resources (i.e., remove namespace and all contained objects)
```
mkolesnik, line 132: This is akin to uninstalling Submariner and should be handled properly.

submariner-operator/broker-multi-tenancy.md
```diff
@@ -0,0 +1,321 @@
...
+to support tenant management by adding the following commands (see details below):
+  - `subctl deploy axon-cp`
+  - `subctl create|delete|get|list tenant [tenant_id]`
+- Create the submariner CRDs when deploying the Axon operator, assuming that
+tenants (i.e., ClusterSet Admins) will not have permissions to create those.
+The `deploy broker` command will ensure the CRDs exist.
+- A submariner tenant is represented by a tenant object in the Axon namespace
+ and a dedicated tenant namespace. Each tenant has a string ID and the namespace
+ is some constant prefix + tenant ID.
+- A SA is created per tenant and a *partial* `broker-info.subm` file is generated
```
mkolesnik, line 128: So why not create a `broker-info.subm.template` or something?

submariner-operator/broker-multi-tenancy.md
```diff
@@ -0,0 +1,321 @@
...
+  tenant. If the field exists, it takes precedence over the `--namespace` CLI
+  argument. If neither is provided, uses the default broker namespace.
+  - Use the access credentials and the Broker URL that are already specified in
+  the `.subm` file to access the Broker, if they exist. Otherwise - fallback to
+  details from kubecontext.
+  - The PSK will still be generated only in the `deploy broker` command, and will
+  be added to the `.subm` file (backing-up the old file).
+  - Deprecate the `--ipsec-psk-from` CLI option, as its name is not representative
+  anymore, and using `--broker-info-from` option as a synonym.
+  - The submariner operator in the Broker cluster will not be deployed anymore,
```
mkolesnik, line 145: AFAIK the operator creates the CRDs and such, and should still be deployed in case no "axon controller" is used as to not impact current behavior.
You could however make sure that it doesn't try to recreated the CRDs if they already exist (which would be a smart idea regardless).

I think the operator should still be responsible for creating the resources it creates, and this responsibility shouldnt be moved elsewhere.

submariner-operator/broker-multi-tenancy.md
```diff
@@ -0,0 +1,321 @@
...
+1. Get target cluster from kube context
+1. Deploy `axon-operator` (consider a better name)
+    1. Ensure (create if missing) NS exist: `axon-operator`
+    1. Ensure CRD exist: `tenants.submariner.io`
+    1. Ensure CRDs relevant to submariner broker exist:
+        - brokers
+        - clusters
+        - endpoints
+        - serviceimports
+        - serviceexports
```
mkolesnik, line 179: This should remain the responsibility of core submariner

submariner-operator/broker-multi-tenancy.md
```diff
@@ -0,0 +1,321 @@
...
+
+## Open Questions
+
+- Do we need to update helm too?
+- Do we still need to support the `--namespace` CLI arg? Is it still required to
+support multiple ClusterSets in classic (non-Axon) submariner?
+- How do we authenticate and authorize users beyond the Broker Admin (e.g., k8s
+service accounts)? Should we support additional principles (e.g., users internal
+to a tenant) and permissions within a ClusterSet?
+- Do we create a new repo for axon-cp operator? Or put it in
```
mkolesnik, line 320: IMHO it should be a separate project building on top of Submariner, with it's own tooling and CLI

mkolesnik (changes requested):

submariner-operator/broker-multi-tenancy.md
davidohana: Currently, the broker cannot be deployed (at least by `subctl`) without having cluster-admin permissions, since it needs to create CRDs and also because it deploys the operator to a shared namespace on the broker cluster (`submariner-operator`). We would like `subctl` to be able to deploy a broker, also if the current cluster context has access to a single namespace only.

submariner-operator/broker-multi-tenancy.md
davidohana: Yes, that's the intent. But notice it's a non-goal. Just clarify that we don't mean to support that for now.

submariner-operator/broker-multi-tenancy.md
davidohana: That is the intent. I will delete the "global configuration" statement as it is confusing.

submariner-operator/broker-multi-tenancy.md
davidohana: The namespaced RBAC that is implemented today seems adequate currently.

submariner-operator/broker-multi-tenancy.md
davidohana: Yes, I already understood that and working to align the design so that tenant management is outside Submariner core.

submariner-operator/broker-multi-tenancy.md
davidohana: So that user can already use the generated `subm` file with `subctl` to connect to the to the broker cluster with the correct user and namespace context

submariner-operator/broker-multi-tenancy.md
davidohana: The intent was to create the CRDs required for the broker outside of the operator as an initial setup step.
If we have namespace permission we cannot create CRDs while deploying the broker. We also cannot deploy the operator because it has its own namespace.

submariner-operator/broker-multi-tenancy.md
sridhargaddam: +1. 

Today with ACM/OCM we can deploy multiple ClusterSets on the same Hub Cluster but each one in a different broker namespace. Once subctl is enhanced to specify a namespace as part of `deploy-broker`/`join` commands and when it avoids using shared namespace (i.e., submariner-operator), one could use the same Broker Cluster for different ClusterSets just like ACM/OCM. With this, the first goal should cover it.

submariner-operator/broker-multi-tenancy.md
```diff
@@ -0,0 +1,321 @@
...
+by allowing the latter to deploy a Broker with restricted namespace permission.
+- Provide backward compatibility for existing deployments (e.g.,
+explicit opt-in, defaulting to current Submariner behavior and settings by default).
+
+### Non-Goals
+
+- Support live migration of an existing Broker to support multi-tenancy.
+- Support for workload clusters joining more than one ClusterSet
+(in the case of multiple tenants in workload clusters, we could run a
+separate cluster control plane per tenant).
```
sridhargaddam, line 77: I'm not sure if I understood it correctly. With this enhancement proposal will we have a **single** submariner-operator pod running on the Broker cluster managing all the ClusterSets that belong to different tenants?

davidohana: Actually, it's a new operator called AxonOperator and it shall be single in the broker cluster. Should be mainly responsible for tenant (namespace) allocation/management for now, but probably more responsibility will be added in the future. 
SubmarinerOperator is not deployed anymore in the broker anymore according to this enhancement. If it does, it should be deployed per tenant, in the tenant namespace and not in a global namespace.

Issue #53, opened 2021-11-02
Epic: Restructure/refactor submariner-operator repo


Jaanki:
## Epic Description
This will ensure that submariner-operator repo structure follows go standards and is optimised.

## Acceptance Criteria
submariner-operator code is as much optimised and modular as possible.

## Definition of Done (Checklist)

* [x] Code complete
* ~~[ ] Relevant metrics added~~
* [x] The acceptance criteria met
* [x] Unit/e2e test added & pass
* [x] CI jobs pass
* ~~[ ] Deployed using cloud-prepare+subctl~~
* ~~[ ] Deployed using ACM/OCM addon~~
* ~~[ ] Deploy using Helm~~
* ~~[ ] Deployed on supported platforms (for e.g kind, OCP on AWS, OCP on GCP)~~
* [x] Run subctl verify, diagnose and gather
* ~~[ ] Uninstall~~
* ~~[ ] Troubleshooting (gather/diagnose) added~~
* ~~[ ] Documentation added~~
* ~~[ ] Release notes added~~

## Work Items
- [x] Make subctl a library Depends on https://github.com/submariner-io/submariner-operator/issues/1329

PR #60, opened 2021-12-15
Decouple diagnose implementation from subctl

vthapar:
Decouple `subctl diagonse` CLI parts from actual implementation
so it can be run from OCM UI without requiring `subctl`.

### DIFF
--- a/submariner-operator/diagnose-refactor-ocm.md
+++ b/submariner-operator/diagnose-refactor-ocm.md
@@ -15,25 +15,34 @@ this Job, track its completion and fetch results once done.
 
 * A new option `--output json` will be added to `subctl diagnose` to capture output in a json format instead of current
  user readable text format.
-* `subctl diagnose` will also update some Prometheus metrices whenever it is run.
-* A `SubmarinerDiagnoseJob` Job will be created which will run `subctl diagnose` with relevant flags as `command`. This
+* `subctl diagnose` will also update some Prometheus metrics whenever it is run.
+* A `submariner-diagnose` Job will be created which will run `subctl diagnose` with relevant flags as `command`. This
  job will run once to completion. This Job will always run with new option of `--output json`.
 * A new `SubmarinerDiagnose` CRD will be added to SubmarinerAddon. OCM UI will create this CR with relevant values
  to trigger running diagnose from OCM UI.`
-* A new `SubmarnerDiagnose` Controller will be added to OCM SubmarinerAddon. This controller will spawn the SubmarinerDiagnose
+* A new `SubmarnerDiagnoseController` Controller will be added to OCM SubmarinerAddon. This controller will spawn the SubmarinerDiagnose
  Job when SubmarinerDiagnose CR is created, track completion.
-* When SubmarinerDiagnose Job is completed, SubmarinerDiagnoseControler will fetch the logs from diagnose Pod, extract the
+* Only one active `submariner-diagnose` Job will be run in a cluster by SubmarinerAddon at a given time. Users can create
+ multiple SubmarinerDiagnose Jobs manually and we won't enforce any checks to prevent that.  
+* When `submariner-diagnose` Job is completed, SubmarinerDiagnoseController will fetch the logs from diagnose Pod, extract the
  results in json format and populate the relevant Status fields in SubmarinerDiagnose CR.
-* Multiple SubmarinerDiagnose Jobs can be created and run at same time. Note that some of the diagnostic subcommands might have
- issues with multiple concurrent runs as they may interfere with each other.
 * Support will be added to OCM UI to run diagnostics on a cluster with a single click and then display the results. For now,
- it will run with suboption `all`
+ it will run with sub-option `all`
 * Results of diagnose will also be exposed as Prometheus metrics. This will allow any Prometheus based observability
   solutions to consume the results of diagnose. This will also provide a quick and easy way to show results in the UI while
   SubmarinerAddon changes are in-progress.
 * [Initial scope] For initial implementation, SubmarinerDiagnose will always run with `all` but without Firewall Options.
  This is to reduce complexity on OCM UI side to pass credentials for remote cluster which are required for Firewall option.
  This support will be added in later release.
+
+### Alternate Design
+
+* Instead of `--output json` option to provide results in json format for SubmarinerAddon to consume, it will just retain
+ existing text output.
+* UI will provide a link to logs of `submariner-diagnose` Pod and show results as is to the user.
+* Advantage of this design is that newer features can be added to `subctl diagnose` without requiring corresponding changes
+ in UI to consume those results.
+* This will also simplify the code in SubmarinerAddon as it will no longer need to extract and parse diagnose results.
 
 ### SubmarinerDiagnose CRD
 
@@ -80,8 +89,9 @@ FirewallStatus FirewallStatus `json:"firewallStatus,omitempty"`
 }
 
 type ConnectionsStatus struct {
-	// TBD
+// TBD
 }
+
 type FirewallPortStatus string
 
 const (
@@ -108,11 +118,15 @@ spec:
       containers:
       - name: submariner-diagnose-xyz123
         image: submariner-diagnose
-        command: ["subctl",  "diagnose", "all"]
+        command: ["subctl",  "diagnose", "all", "--in-cluster"]
       restartPolicy: Never
+      serviceAccount: submariner-diagnose
+      serviceAccountName: submariner-diagnose
+  backoff-limit: 0
 ```
 
 ### Prometheus Metrics
+
 * submariner_diagnose_errors Total count of error detected, label: cluster-name: counter, will only go up, +1 for each
  error detected by running `subctl diagnose`
 * submariner_k8s_version_support K8s version supported label: cluster-name(1 - for true, 0 - for false)
--- a/submariner-operator/diagnose-refactor-ocm.md
+++ b/submariner-operator/diagnose-refactor-ocm.md
@@ -1,0 +1,137 @@
+# Refactor subctl diagonose for use with OCM
+
+## Summary
+
+Currently only way to run diagnostics on a Submariner installation is through `subctl` CLI. This requires user or admin
+to run it manually from a host. There is no way to run it through UI like OCM or deployment tools like helm charts etc.
+
+## Proposal
+
+This enhancement proposes a means to run `subctl diagnose` through a `SubmarinerDiagnose` Job. User or Admin can deploy
+this Job manually. In case of OCM, a new SubmarinerDiagnose Controller will be added to SubmarinerAddon which will deploy
+this Job, track its completion and fetch results once done.
+
+## Design Details
+
+* A new option `--output json` will be added to `subctl diagnose` to capture output in a json format instead of current
+ user readable text format.
+* `subctl diagnose` will also update some Prometheus metrices whenever it is run.
+* A `SubmarinerDiagnoseJob` Job will be created which will run `subctl diagnose` with relevant flags as `command`. This
+ job will run once to completion. This Job will always run with new option of `--output json`.
+* A new `SubmarinerDiagnose` CRD will be added to SubmarinerAddon. OCM UI will create this CR with relevant values
+ to trigger running diagnose from OCM UI.`
+* A new `SubmarnerDiagnose` Controller will be added to OCM SubmarinerAddon. This controller will spawn the SubmarinerDiagnose
+ Job when SubmarinerDiagnose CR is created, track completion.
+* When SubmarinerDiagnose Job is completed, SubmarinerDiagnoseControler will fetch the logs from diagnose Pod, extract the
+ results in json format and populate the relevant Status fields in SubmarinerDiagnose CR.
+* Multiple SubmarinerDiagnose Jobs can be created and run at same time. Note that some of the diagnostic subcommands might have
+ issues with multiple concurrent runs as they may interfere with each other.
+* Support will be added to OCM UI to run diagnostics on a cluster with a single click and then display the results. For now,
+ it will run with suboption `all`
+* Results of diagnose will also be exposed as Prometheus metrics. This will allow any Prometheus based observability
+  solutions to consume the results of diagnose. This will also provide a quick and easy way to show results in the UI while
+  SubmarinerAddon changes are in-progress.
+* [Initial scope] For initial implementation, SubmarinerDiagnose will always run with `all` but without Firewall Options.
+ This is to reduce complexity on OCM UI side to pass credentials for remote cluster which are required for Firewall option.
+ This support will be added in later release.
+
+### SubmarinerDiagnose CRD
+
+```Go
+// SubmarinerDiagnoseSpec defines the desired configuration to run SubmarinerDiagnose
+type SubmarinerDiagnoseSpec struct {
+All             bool         `json:"all,omitempty"`
+CNI             bool         `json:"CNI,omitempty"`
+Connections     bool         `json:"connections,omitempty"`
+Deployment      bool         `json:"deployment,omitempty"`
+Firewall        bool         `json:"firewall,omitempty"`
+GatherLogs      bool         `json:"gatherLogs,omitempty"`
+K8sVersion      bool         `json:"k8sVersion,omitempty"`
+KubeProxyMode   bool         `json:"kubeProxyMode,omitempty"`
+FirewallOptions FirewallOptions `json:"firewallOptions,omitempty"`
+
+// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+// Important: Run "make manifests" to regenerate code after modifying this file
+// Add custom validation using kubebuilder tags: https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html
+}
+
+type FirewallOptions struct {
+InterCluster bool `json:"interCluster,omitempty"`
+IntraCluster bool `json:"intraCluster,omitempty"`
+Metrics      bool `json:"metrics,omitempty"`
+RemoteCluster            string `json:"remoteCluster,omitempty"`
+RemoteK8sAPIServer       string `json:"remoteK8sAPIServer,omitempty"`
+RemoteK8sAPIServerToken  string `json:"remoteK8sAPIServerToken,omitempty"`
+RemoteK8sCA              string `json:"remoteK8sCA,omitempty"`
+RemoteK8sSecret          string `json:"remoteK8sSecret,omitempty"`
+RemoteK8sRemoteNamespace string `json:"remoteK8sRemoteNamespace,omitempty"`
+}
+
+// SubmarinerDiagnoseStatus defines the observed result of SubmarinerDiagnose
+type SubmarinerDiagnoseStatus struct {
+Conditions []metav1.Condition `json:"conditions"`
+K8sVersion string `json:"k8sVersion,omitempty"`
+CNIType string `json:"cniType,omitempty"`
+KubeProxyMode bool `json:"kubeProxyMode,omitempty"`
+ConnectionsStatus ConnectionsStatus `json:"connectionsStatus,omitempty"`
+FirewallStatus FirewallStatus `json:"firewallStatus,omitempty"`
+// DeploymentStatus already captured in SubmarinerStatus, no need to duplicate information.
+// DeploymentStatus SubmarinerStatus `json:"DeploymentStatus,omitempty"`
+}
+
+type ConnectionsStatus struct {
+	// TBD
+}
+type FirewallPortStatus string
+
+const (
+Allowed = "allowed"
+Blocked = "blocked"
+Unknown = "unknown"
+)
+
+type FirewallStatus struct {
+Metrics FirewallPortStatus `json:"metricsStatus,omitempty"`
+VxLanTunnel FirewallPortStatus `json:"vxLanTunnel,omitempty"`
+IPSecTunnel FirewallPortStatus `json:"IPSecTunnel,omitempty"`
+}
+```
+
+```yaml
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: submariner-diagnose-xyz123
+spec:
+  template:
+    spec:
+      containers:
+      - name: submariner-diagnose-xyz123
+        image: submariner-diagnose
+        command: ["subctl",  "diagnose", "all"]
+      restartPolicy: Never
+```
+
+### Prometheus Metrics
+* submariner_diagnose_errors Total count of error detected, label: cluster-name: counter, will only go up, +1 for each
+ error detected by running `subctl diagnose`
+* submariner_k8s_version_support K8s version supported label: cluster-name(1 - for true, 0 - for false)
+* submariner_cni_support Cni support . label: cluster-name, cni name (1 - for true, 0 - for false)
+* submariner_globalnet_overlapping_CIDR Globalnet - is overlapping CIDR label: cluster-name(1 - true, 0 - false)
+* submariner_kube_proxy_mode_support Kube proxy mode support label: cluster-name (1 - true, 0 - false)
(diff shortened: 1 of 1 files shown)


Jaanki:
This is a part of https://github.com/submariner-io/enhancements/issues/53

vthapar:
> This is a part of #53

It is not a simple refactor of code, but redesign. It kinda depend son #53 because we want that to be done before we modify the code for this redesign. Edited title to reflect the difference.

submariner-operator/diagnose-refactor-ocm.md
```diff
@@ -0,0 +1,105 @@
...
+## Proposal
+
+This enhancement proposes decouple the diagnose implementation from `subctl` into a Debug controller running
+in the `submariner-operator` namespace. A new CRD `SubmarinerDebug` will be created. User or Admin can create this CR to
+trigger `diagnose` run in that cluster and results will be stored in the corresponding Status fields of SubmarinerDebug
+CR created.
+
+## Design Details
+
+* A new CRD `SubmarinerDebug` will be created. Config fields will match current flag sin `subctl` and will have Status
```
maayanf24:
```suggestion
* A new CRD `SubmarinerDebug` will be created. Config fields will match current flags in `subctl` and will have Status
```

submariner-operator/diagnose-refactor-ocm.md
```diff
@@ -0,0 +1,105 @@
...
+
+type FirewallStatus struct {
+	Metrics FirewallPortStatus `json:"metricsStatus,omitempty"`
+	VxLanTunnel FirewallPortStatus `json:"vxLanTunnel,omitempty"`
+	IPSecTunnel FirewallPortStatus `json:"IPSecTunnel,omitempty"`
+}
+```
+
+### Prometheus Metrics
+TBD
```
maayanf24:
```suggestion
* submariner_diagnose_errors Total count of error detected, label: cluster-name: counter, will only goes up, +1 for each error detected by running subctl diagnose
* submariner_k8s_version_support K8s version supported label: cluster-name(1 - for true, 0 - for false)
* submariner_cni_support Cni support . label: cluster-name, cni name (1 - for true, 0 - for false)
* submariner_globalnet_overlapping_CIDR Globalnet - is overlapping CIDR label: cluster-name(1 - true, 0 - false)
* submariner_kube_proxy_mode_support Kube proxy mode support label: cluster-name (1 - true, 0 - false)
* submariner_deployment_status Deployment status, label:cluster-name, daemonset / deployment name (1 - all good, 0 - error)
* submariner_pod_status Pod status, label: cluster-name, pod name, status ( 1 - all good, 0 - error)
* submariner_pod_restart_count Pod restart count (if count > 5), label: cluster-name, pod name, status . value = restart count
* submariner_firewall_vxlan_tunnel Firewall intra-cluster, label: cluster-name (1 - all good, 0 - error)
* submariner_firewall_ipsec_tunnel Firewall inter-cluster, labels: local-cluster, remote-cluster (1 - all good, 0 - error)

```

maayanf24 (commented):
Thanks for adding this.
I've added the suggested metrics

submariner-operator/diagnose-refactor-ocm.md
```diff
@@ -0,0 +1,105 @@
...
+// SubmarinerDebugSpec defines the desired configuration to run SubmarinerDebug
+type SubmarinerDebugSpec struct {
+	All             bool         `json:"all,omitempty"`
+	CNI             bool         `json:"CNI,omitempty"`
+	Connections     bool         `json:"connections,omitempty"`
+	Deployment      bool         `json:"deployment,omitempty"`
+	Firewall        bool         `json:"firewall,omitempty"`
+	GatherLogs      bool         `json:"gatherLogs,omitempty"`
+	K8sVersion      bool         `json:"k8sVersion,omitempty"`
+	KubeProxyMode   bool         `json:"kubeProxyMode,omitempty"`
```
skitt: This could be a list of strings instead, which would avoiding tying the CR to diagnostic implementation details.

submariner-operator/diagnose-refactor-ocm.md
```diff
@@ -0,0 +1,105 @@
...
+	FirewallOptions FirewallOptions `json:"firewallOptions,omitempty"`
```
skitt: A generic configmap would allow greater flexibility, _e.g._ if other sub-commands end up requiring parameters.

skitt (changes requested):

tpantelis (commented):
Just a thought...instead of using a controller and CRD to trigger collection and communicate the results, could we have an executable that is launched on demand via a k8s Job that writes the results to stdout which the caller can scrape?

submariner-operator/diagnose-refactor-ocm.md
```diff
@@ -0,0 +1,105 @@
+# Refactor subctl diagonose for use with OCM
+
+## Summary
+
+Currently only way to run diagnostics on a Submariner installation is through `subctl` CLI. This requires user or admin
+to run it manually from a host. There is no way to run it through UI like OCM or deployment tools like helm charts etc.
+
+## Proposal
+
+This enhancement proposes decouple the diagnose implementation from `subctl` into a Debug controller running
```
Jaanki:
```suggestion
This enhancement proposes decouple of the `diagnose` implementation from `subctl` into a Debug controller running
```

or 

```suggestion
This enhancement proposes decoupling the `diagnose` implementation from `subctl` into a Debug controller running
```

vthapar:
> Just a thought...instead of using a controller and CRD to trigger collection and communicate the results, could we have an executable that is launched on demand via a k8s Job that writes the results to stdout which the caller can scrape?

I am not sure if I get what you mean by K8s Job. Do you mean just launch a pod with a binary?

We will still need some code to launch the executable and track stdout while it is running in realtime. Such code will need to be duplicated in both subctl and submariner-addon and any other means we may want to run diagnose. Controller+CRD is to prvide a single interface to do it irrespective of where one wants to run it from. If we want to scrape from stdout, it will require realtime tracking of output. Using a CR to store results makes it easier to do it from a different cluster, query results at a later time once run is done and also will allow OCM to consume the results.

skitt:
K8s jobs are [described here](https://kubernetes.io/docs/concepts/workloads/controllers/job/). You don’t need to scrape the output in real time; a job will run until it’s complete, and then you can retrieve the pod logs as usual.

tpantelis:
> K8s jobs are [described here](https://kubernetes.io/docs/concepts/workloads/controllers/job/). You don’t need to scrape the output in real time; a job will run until it’s complete, and then you can retrieve the pod logs as usual.

Right. The results would be stored in the job's pod log rather than a CR.

submariner-operator/diagnose-refactor-ocm.md
```diff
@@ -0,0 +1,142 @@
...
+## Proposal
+
+This enhancement proposes to decouple the diagnose implementation from `subctl` into a Debug controller running
+in the `submariner-operator` namespace. A new CRD `SubmarinerDebug` will be created. User or Admin can create this CR to
+trigger `diagnose` run in that cluster and results will be stored in the corresponding Status fields of SubmarinerDebug
+CR created.
+
+## Design Details
+
+* A new Image `submariner-debug` will be created with a binary `submariner-diagnose` which will contain `diagnose`
```
skitt: I would avoid naming anything “debug” (the image, the CR etc.) — from experience, this tends to produce adverse reactions from people managing deployments! (You never want to see anything “debug”-related in staging or production.)

submariner-operator/diagnose-refactor-ocm.md
```diff
@@ -0,0 +1,142 @@
...
+implementatio logic.
```
Jaanki:
```suggestion
implementation logic.
```

submariner-operator/diagnose-refactor-ocm.md
```diff
@@ -0,0 +1,142 @@
...
+* A new CR, `SubmarinerDebugStatus` will be created to capture and store results of diagnose.
+* `diagnose` will be run via a `SubmarinerDebug` Job using the newly created `submariner-debug` image with
+  appropriate parameters.
+* For subcommands like `firewall` which require additional parameters, a `ConfigMap` will be used.
+* `SubmarinerDebug` will create the `SubmarinerDebugStatus` CR with results of diagnostics run.
+* Multiple SubmarinerDebug Jobs can be created and run at same time. Name of each ConfigMap and SubmarinerDebugStatus CR
```
Jaanki:
```suggestion
* Multiple SubmarinerDebug Jobs can be created and run at the same time. Name of each ConfigMap and SubmarinerDebugStatus CR
```

submariner-operator/diagnose-refactor-ocm.md
```diff
@@ -0,0 +1,142 @@
...
+  will be same as the respective Job. Note that some of diagnostic subcommands might have issues with multiple concurrent
```
Jaanki:
```suggestion
  will be same as the respective Job. Note that some of the diagnostic subcommands might have issues with multiple concurrent
```

submariner-operator/diagnose-refactor-ocm.md
```diff
@@ -0,0 +1,142 @@
...
+This enhancement proposes to decouple the diagnose implementation from `subctl` into a Debug controller running
+in the `submariner-operator` namespace. A new CRD `SubmarinerDebug` will be created. User or Admin can create this CR to
+trigger `diagnose` run in that cluster and results will be stored in the corresponding Status fields of SubmarinerDebug
+CR created.
+
+## Design Details
+
+* A new Image `submariner-debug` will be created with a binary `submariner-diagnose` which will contain `diagnose`
+implementatio logic.
+* A new CR, `SubmarinerDebugStatus` will be created to capture and store results of diagnose.
```
Jaanki: L12 says output will be stored in status field of `SubmarinerDebug` CRD and this line says a new CR, `SubmarinerDebugStatus` will be created to store the output.

submariner-operator/diagnose-refactor-ocm.md
```diff
@@ -0,0 +1,142 @@
...
+* Instead of a new binary and refactoring `subctl diagnose`, reuse subctl itself.
+* Deploy SubmarinerDebug Job to run `subctl diagnose` with relevant parameters passed at command line
+* Read the `subctl diagnose` output from SubmarinerDebug Pod logs and create relevant SubmarinerDebugStatus CR for OCM
+ to consume.
+* Pros: No need to refactor subctl and we reuse subctl itself
+* Cons: Need to parse the textual output of `subctl diagnose` and convert it into a CR format that can be easily consumed
+ by OCM, need to provide credentials in files on pod so can be used by `subctl`
+
+### Prometheus Metrics
+* submariner_diagnose_errors Total count of error detected, label: cluster-name: counter, will only goes up, +1 for each
```
Jaanki:
```suggestion
* submariner_diagnose_errors Total count of error detected, label: cluster-name: counter, will only go up, +1 for each
```

Jaanki (changes requested):

submariner-operator/diagnose-refactor-ocm.md
vthapar: Missed changing L12. This was older method when we were planning to create ConfigCR in operator itself. New plan is to run it through a job, so config CR not needed in operator anymore.

Jaanki (approved):

submariner-operator/diagnose-refactor-ocm.md
```diff
@@ -0,0 +1,137 @@
...
+This enhancement proposes a means to run `subctl diagnose` through a `SubmarinerDiagnose` Job. User or Admin can deploy
+this Job manually. In case of OCM, a new SubmarinerDiagnose Controller will be added to SubmarinerAddon which will deploy
+this Job, track its completion and fetch results once done.
+
+## Design Details
+
+* A new option `--output json` will be added to `subctl diagnose` to capture output in a json format instead of current
+ user readable text format.
+* `subctl diagnose` will also update some Prometheus metrices whenever it is run.
+* A `SubmarinerDiagnoseJob` Job will be created which will run `subctl diagnose` with relevant flags as `command`. This
```
maayanf24: L10 refers to `SubmarinerDiagnose` Job, and in this line: `SubmarinerDiagnoseJob` Job

submariner-operator/diagnose-refactor-ocm.md
```diff
@@ -0,0 +1,137 @@
...
+ job will run once to completion. This Job will always run with new option of `--output json`.
+* A new `SubmarinerDiagnose` CRD will be added to SubmarinerAddon. OCM UI will create this CR with relevant values
```
maayanf24, line 18: Can a job has the same name as a CRD? `SubmarinerDiagnose`

submariner-operator/diagnose-refactor-ocm.md
```diff
@@ -0,0 +1,137 @@
...
+ to trigger running diagnose from OCM UI.`
+* A new `SubmarnerDiagnose` Controller will be added to OCM SubmarinerAddon. This controller will spawn the SubmarinerDiagnose
+ Job when SubmarinerDiagnose CR is created, track completion.
+* When SubmarinerDiagnose Job is completed, SubmarinerDiagnoseControler will fetch the logs from diagnose Pod, extract the
```
maayanf24: L23 - `SubmarnerDiagnose` Controlle, here: `SubmarinerDiagnoseControler`

submariner-operator/diagnose-refactor-ocm.md
vthapar: > Can a job has the same name as a CRD? `SubmarinerDiagnose`

Yes. Though open to suggestions for name here. going by convention, Job name should be `submariner-diagnose`. Maybe `submariner-diagnose-<randomized>` if we want to support multiple jobs running at a time. Note that most names are just placeholders for discussion and appreciate inputs on them.

submariner-operator/diagnose-refactor-ocm.md
```diff
@@ -0,0 +1,152 @@
...
+* A `submariner-diagnose` Job will be created which will run `subctl diagnose` with relevant flags as `command`. This
+ job will run once to completion. This Job will always run with new option of `--output json`.
+* A new `SubmarinerDiagnose` CRD will be added to SubmarinerAddon. OCM UI will create this CR with relevant values
+ to trigger running diagnose from OCM UI.`
+* A new `SubmarnerDiagnoseController` Controller will be added to OCM SubmarinerAddon. This controller will spawn the SubmarinerDiagnose
+ Job when SubmarinerDiagnose CR is created, track completion.
+* Only one active `submariner-diagnose` Job will be run in a cluster by SubmarinerAddon at a given time.
+* When `submariner-diagnose` Job is completed, SubmarinerDiagnoseController will fetch the logs from diagnose Pod, extract the
+ results in json format and populate the relevant Status fields in SubmarinerDiagnose CR.
+* Multiple SubmarinerDiagnose Jobs can be created and run at same time. Note that some of the diagnostic subcommands might have
```
tpantelis: This seems to contradict 25 above?

vthapar: > This seems to contradict 25 above?

1. User can always run multiple `submariner-diagnose` Jobs manually, so we can't really force it.
2. 25 talks more about only one Active Job being available. I'll add detail to 28 to clarify that we will only allow one from addon.

submariner-operator/diagnose-refactor-ocm.md
skitt: K8s can handle the name generation for us (`GenerateName` instead of `Name`). Having a fixed name makes everything simpler for now though!

skitt (approved):

maayanf24 (approved):

submariner-operator/diagnose-refactor-ocm.md
```diff
@@ -0,0 +1,151 @@
...
+
+* A new option `--output json` will be added to `subctl diagnose` to capture output in a json format instead of current
+ user readable text format.
+* `subctl diagnose` will also update some Prometheus metrics whenever it is run.
+* A `submariner-diagnose` Job will be created which will run `subctl diagnose` with relevant flags as `command`. This
+ job will run once to completion. This Job will always run with new option of `--output json`.
+* A new `SubmarinerDiagnose` CRD will be added to SubmarinerAddon. OCM UI will create this CR with relevant values
+ to trigger running diagnose from OCM UI.`
+* A new `SubmarnerDiagnoseController` Controller will be added to OCM SubmarinerAddon. This controller will spawn the SubmarinerDiagnose
+ Job when SubmarinerDiagnose CR is created, track completion.
```
tpantelis, line 21: With https://github.com/submariner-io/submariner-operator/pull/1790, diagnose can be invoked programmatically instead of the additional overhead of creating a `Job` with a new image. The output can be captured via a `reported.Interface` instance.

submariner-operator/diagnose-refactor-ocm.md
```diff
@@ -0,0 +1,151 @@
...
+ This support will be added in later release.
+
+### Alternate Design
+
+* Instead of `--output json` option to provide results in json format for SubmarinerAddon to consume, it will just retain
+ existing text output.
+* UI will provide a link to logs of `submariner-diagnose` Pod and show results as is to the user.
+* Advantage of this design is that newer features can be added to `subctl diagnose` without requiring corresponding changes
+ in UI to consume those results.
+* This will also simplify the code in SubmarinerAddon as it will no longer need to extract and parse diagnose results.
```
tpantelis: I like this approach better, ie just store the raw output as a string as is and present to the user. Perhaps output in HTML.

tpantelis (dismissed):

submariner-operator/diagnose-refactor-ocm.md
tpantelis: How will these objects be deleted?

submariner-operator/diagnose-refactor-ocm.md
vthapar: > I would avoid naming anything “debug” (the image, the CR etc.) — from experience, this tends to produce adverse reactions from people managing deployments! (You never want to see anything “debug”-related in staging or production.)

Agree. Will rename all these to `diagnose`

submariner-operator/diagnose-refactor-ocm.md
vthapar: UI code can't use Go APIs. So we need something to proxy UI requests. To use API we will need to add implementation in submariner-addon, which was the first iteration draft we had. UI can easily show the pod logs directly as it can use any K8s restful API.

submariner-operator/diagnose-refactor-ocm.md
vthapar: Just waiting for confirmation from UI team that they can show log output. Unless they say they can't we plan to go with this design.

submariner-operator/diagnose-refactor-ocm.md
tpantelis: ok so we don't actually need the `SubmarnerDiagnoseController` as this states. But if we were to use the `SubmarnerDiagnoseController` it could use the **submariner-operator** API instead of creating a `Job`.

skitt: Yes, the idea here is to avoid tying the UI to `Job`; instead, the UI only knows how to ask the operator to start a diagnostic, and we can change the operator to use something other than a `Job` if appropriate later on without changing the UI.

submariner-operator/diagnose-refactor-ocm.md
mkolesnik: Is there an update?

submariner-operator/diagnose-refactor-ocm.md
```diff
@@ -0,0 +1,151 @@
...
+* [Initial scope] For initial implementation, SubmarinerDiagnose will always run with `all` but without Firewall Options.
+ This is to reduce complexity on OCM UI side to pass credentials for remote cluster which are required for Firewall option.
+ This support will be added in later release.
+
+### Alternate Design
+
+* Instead of `--output json` option to provide results in json format for SubmarinerAddon to consume, it will just retain
+ existing text output.
+* UI will provide a link to logs of `submariner-diagnose` Pod and show results as is to the user.
+* Advantage of this design is that newer features can be added to `subctl diagnose` without requiring corresponding changes
```
mkolesnik: I think this approach is for sure simpler but probably less well integrated in the UI?

Although if the diagnose command reports prometheus metrics I guess that provides some friendlier integration.

submariner-operator/diagnose-refactor-ocm.md
```diff
@@ -0,0 +1,151 @@
...
+  SubmarinerAddon changes are in-progress.
+* [Initial scope] For initial implementation, SubmarinerDiagnose will always run with `all` but without Firewall Options.
+ This is to reduce complexity on OCM UI side to pass credentials for remote cluster which are required for Firewall option.
+ This support will be added in later release.
+
+### Alternate Design
+
+* Instead of `--output json` option to provide results in json format for SubmarinerAddon to consume, it will just retain
+ existing text output.
+* UI will provide a link to logs of `submariner-diagnose` Pod and show results as is to the user.
```
mkolesnik: This is a bit narrow scope right now, I think if we have this route perhaps it would be nice to let users execute any `subctl` command arbitrarily from the UI, or is it a crazy idea? :)

submariner-operator/diagnose-refactor-ocm.md
```diff
@@ -0,0 +1,151 @@
...
+CNIType string `json:"cniType,omitempty"`
+KubeProxyMode bool `json:"kubeProxyMode,omitempty"`
+ConnectionsStatus ConnectionsStatus `json:"connectionsStatus,omitempty"`
+FirewallStatus FirewallStatus `json:"firewallStatus,omitempty"`
+// DeploymentStatus already captured in SubmarinerStatus, no need to duplicate information.
+// DeploymentStatus SubmarinerStatus `json:"DeploymentStatus,omitempty"`
+}
+
+type ConnectionsStatus struct {
+// TBD
```
mkolesnik: Still TBD?

submariner-operator/diagnose-refactor-ocm.md
vthapar: Checked with UI team and they are okay with this. They prefer this as having link to logs also saves tracking progress of diagnose run as they can just stream log.

They don't have promehteus metrics integrated into UI yet and is one of the topics in upcoming call.

submariner-operator/diagnose-refactor-ocm.md
vthapar: That indeed is where we want to end up with. Option to run any command for "power users" alongwith easy to run options for rest. We're going with this narrow scope for now primarily coz it is easier/faster to do in UI.

submariner-operator/diagnose-refactor-ocm.md
vthapar: Had discussions with UI team, so we're dropping this for now. We'll have to revisit this when UI supports topoolgy view. At that time we will define these in an easy to consume format for the UI.

submariner-operator/diagnose-refactor-ocm.md
mkolesnik: Yes I understand why it's easier to start with a narrow scope just want to make sure if it is indeed the direction we don't make any design right now that later on will be hard or impossible to change without breaking backward compatibility, that's all.

skitt (approved):

mkolesnik (approved):

submariner-operator/diagnose-refactor-ocm.md
```diff
@@ -0,0 +1,152 @@
...
+kind: Job
+metadata:
+  name: submariner-diagnose-xyz123
+spec:
+  template:
+    spec:
+      containers:
+      - name: submariner-diagnose-xyz123
+        image: submariner-diagnose
+        command: ["subctl",  "diagnose", "all", "--in-cluster"]
```
sridhargaddam, line 122: Is "--in-cluster" some new flag? Cos' `subctl diagnose all` itself runs within the local cluster alone.

PR #61, opened 2021-12-21
Design proposal for alternative service control plane

elevran:
Proposal for centralized service control plane, with more flexible service naming and importing.

Submariner current service management follows the MCS model and works well in an environment where clusters have a high level of trust and sharing (e.g., clusters used as the runtime infrastructure and developers can deploy to any cluster for availability, redundancy, or geographic proximity).

Another possible set-up, is of clusters under different administrative domains (e.g., separated
by team). In those environments, service naming and sharing may need to be controlled 
differently, to allow:

1. Independent service naming (i.e., allow use of different names in different clusters).
1. Selective imports (i.e., import services into a subset of clusters in the ClusterSet).
1. Centralized control over exporting and importing (i.e., defined in Broker, not
  workload clusters).

This proposal outline a possible implementation of an controller that runs in the broker cluster and whose configuration
model is overlaid on top of the MCS CRDs.

Signed-off-by: Etai Lev Ran <etai@il.ibm.com>


submariner/centralized-control-plane.md
```diff
@@ -0,0 +1,236 @@
...
+- Service are exported and imported using the same namespace and name; and
+- Service exports are managed from workload clusters.
+
+The above works well in an environment where clusters are used by a single
+ administrative domain and services are commonly shared across all clusters.
+ For example, when a company runs clusters as the runtime infrastructure and developers can deploy
+ to any cluster for availability, redundancy, or geographic proximity. Another
+ common set-up is clusters under different administrative domains (e.g., separated
+ by team). In those environments, service naming and sharing may be controlled
+ differently.
```
dfarrell07: > Another common set-up is clusters under different administrative domains (e.g., separated
 by team). In those environments, service naming and sharing may be controlled
 differently.

Can you provide some more detail about this? I'd like to make sure I/we understand this use-case and the differences you allude to.

submariner/centralized-control-plane.md
```diff
@@ -0,0 +1,236 @@
...
+ We would like to propose a different approach for Service management, that allows:
+
+1. Independent service naming (i.e., allow use of different names in different clusters).
+1. Selective imports (i.e., import services into a subset of clusters in the ClusterSet).
```
dfarrell07, line 50: A cluster is already required to create a `ServiceImport` to select to import a service, does that not meet this requirement?

submariner/centralized-control-plane.md
```diff
@@ -0,0 +1,236 @@
...
+1. Centralized control over exporting and importing (i.e., defined in Broker, not
```
dfarrell07, line 55: This seems like the key goal to me, to be able to manage the imports/exports in a cluster set from central automation. But can that not be built on top of the MCS API, instead of as an alternative? If it's difficult to make automation to do this on top of the MCS API, then that seems like something we should discuss with SIG MC and try to resolve as a broader community.

submariner/centralized-control-plane.md
```diff
@@ -0,0 +1,236 @@
...
+The above works well in an environment where clusters are used by a single
+ administrative domain and services are commonly shared across all clusters.
+ For example, when a company runs clusters as the runtime infrastructure and developers can deploy
+ to any cluster for availability, redundancy, or geographic proximity. Another
+ common set-up is clusters under different administrative domains (e.g., separated
+ by team). In those environments, service naming and sharing may be controlled
+ differently.
+ We would like to propose a different approach for Service management, that allows:
+
+1. Independent service naming (i.e., allow use of different names in different clusters).
```
dfarrell07, line 44: Can you add some details, maybe a simple user story, to help me understand how naming a service differently in different members of a cluster set would be used?

dfarrell07 (commented):
Thanks for the proposal! Questions inline. I wasn't around to learn about Axon recently, sorry if they are quite basic or have already been discussed.

submariner/centralized-control-plane.md
elevran: my understanding is that, in response to a `ServiceExport`, the current service management automatically creates a `ServiceImport` for each and every cluster in the `ClusterSet`.
So while service importing *may* be made selective, it isn't under the current implementation.

Did I misunderstand the current model?

submariner/centralized-control-plane.md
elevran: I'll try and clarify.

You're correct that the MCS API is a building block (i.e., a mechanism) and as such can be used by a different policy engines to drive imports and exports from a central location. This might be "import everywhere" as in Submariner, or a more selective TBD policy.

What's missing, IMHO, are:
1. name independence across clusters;
1. the mechanism to propagate Exports and Import from/to workload clusters. Some synchronization/distribution mechanism is required (e.g., Submariner's Lighthouse is an example). The mechanism is implementation specific and currently not in scope for MCS API; and
1. Policy objects (e.g., how to declaratively define which exports to import where and with what local workload clusters attributes). These are similarly implementation specific and possibly out of MCS scope in its current form.

The first (name independence) seems to me a fundamental assumption on the MCS API. This makes `ServiceExport` and `ServiceImport` less useful as the abstraction in the central shared location. For example, since all imports are created in the Submariner broker namespace of the ClusterSet, Submariner resorts to tagging the destination namespace and name on the `ServiceImport` using labels and annotations. It might have been clearer to explicitly set those in `ServiceImport.Spec` if MCS provided that. Similarly, the assumption on name sameness has implication on `ServiceExport` (e.g., two completely independent cluster, owned and managed by different teams, must agree on using unique namespaces and/or names, lest their exports conflate when shared).

While I think all three (naming, synchronization and policies) are possible enhancements to the MCS API, I also think it would be worthwhile to experiment with different approaches before attempting to reach consensus there.

submariner/centralized-control-plane.md
elevran: Sure. Based on work we've done in Istio (see for example [this](https://istio.io/latest/blog/2019/isolated-clusters/)), we identified two common usage models for clusters:
- **shared infrastructure** - clusters are mostly identically configured (e.g., namespaces, RBAC, etc.). Each team is allocated the same set of namespaces across all clusters and can deploy to any cluster. The boundary between teams is the namespace.
- **federated mode** - clusters are used as the boundary between teams (e.g., they may even be billed to different departments). Each team can only deploy to a subset of the clusters (i.e., the one they "own") and is granted full reign over that cluster. They are free to choose whatever names they desire. Thus, the federated mode cannot assume name sameness. The teams may still have a shared clusters where the "integrated application" runs, pulling in services from the per team clusters.

The federated mode needs to support the following use cases (partial list...):
- I'd like to export my internal services only to my other clusters, but still consume shared services (e.g., centralized monitoring and alerting) provided by other teams' clusters.
- I'd like to decouple the local and global names used for services (e.g., two teams can have a "authz" service running in "frontend" namespace, but each is unique)
- when importing a service I'd like to control its "name" and "location" (e.g., import into a specific namespace with a specific name). I can't assume that the same namespaces exist across all clusters and importing with the name name may conflict with an existing service.

elevran: @dfarrell07 - once clearer, I can also add this to the enhancement proposal so better context is available to the reviewer. I'd like to first use the comments thread here to ensure I'm making a coherent statement and a painting a clear enough picture.
Makes sense?

submariner/centralized-control-plane.md
elevran: I tried to explain some of this above, describing "federated clusters" in response to an earlier comment.

submariner/centralized-control-plane.md
dfarrell07: Yes this is the implementation, but it seems the spec is actually more strict in this wording than I expected:

> When the mcs-controller sees a ServiceExport, a ServiceImport will be introduced in each importing cluster to represent the imported service.

> All clusters containing the service's namespace will import the service.

https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#importing-services

submariner/centralized-control-plane.md
dfarrell07: Thanks for these details, they do help clarify things! And yes it'll be great to add them to the doc eventually, thanks!

> While I think all three (naming, synchronization and policies) are possible enhancements to the MCS API, I also think it would be worthwhile to experiment with different approaches before attempting to reach consensus there.

Yes I think we'll have to show successful experimentation before reaching consensus with the SIG, but I think it'll be less work and a better result if we build that consensus in conjunction with the implementation. I shadow many of the SIG meetings, but @skitt has worked with the SIG a lot and might be interested in helping start that discussion.

submariner/centralized-control-plane.md
elevran: folded these comments into the doc

submariner/centralized-control-plane.md
elevran: added comments into proposal body for more complete context.

submariner/centralized-control-plane.md
```diff
@@ -0,0 +1,298 @@
...
+- Exported services are automatically reflected into all clusters in the ClusterSet;
+- Service are exported and imported using the same namespace and name; and
+- Service exports are managed from workload clusters.
+
+The above works well in an environment where clusters are used by a single
+ administrative domain and services are commonly shared across all clusters.
+ For example, when a company runs clusters as the runtime infrastructure and developers can
+ deploy to any cluster for availability, redundancy, or geographic proximity. However, it is
+ less suited in a deployment model where clusters are owned and operated by different teams,
+ each having independent control and configuration (see next section for elaboration).
```
mkolesnik: One has to ask themself, is MCS API and Submariner even suitable for such an environment?

MCS and Submariner are rather aligned on simplicity of operations for the user, and hence the "clusterset" model is a simple world where the aforementioned principles apply.

If the desire is simply to share *some* services between clusters that are not expected to be operated by the same people and have a "flat network" type of trust model, perhaps we shouldn't be applying MCS/Submariner at all since a better approach might be exposing said services via `Ingress` services?

submariner/centralized-control-plane.md
```diff
@@ -0,0 +1,298 @@
...
+
+Based on work we've done in Istio (see for example [this](https://istio.io/latest/blog/2019/isolated-clusters/)),
+ we identified two common usage models for clusters:
+
+- **Shared Infrastructure**: clusters are (mostly) identically configured (e.g., namespaces,
+  RBAC, etc.). Each team is allocated the same set of namespaces across all clusters and can
+  deploy to any cluster. The boundary between teams is the namespace. This is the model assumed
+  by MCS and Submariner.
+- **Federated**: clusters are used as the boundary between teams (e.g., they may even be
+  billed to different departments). Each team can only deploy to a subset of the clusters
```
mkolesnik, line 96: Is it expected to have full L3 connectivity (e.g. each pod can reach each other pod/service across the cluster-set) in such a scenario?

(I'm asking about the networking aspect on ly in terms of routing, assume that no ACLs exist to block the traffic)

submariner/centralized-control-plane.md
elevran: the MCS API is relatively abstract and not opinionated about how to implement the solution. We can still use it to program the Submariner data plane. The proposal is a superset of the functionality currently provided (i.e., publish to all without name changes - proposal is extending MCS)
Ingress is a less optimal as it becomes the destination/source of traffic instead of the originators (unless we start doing SSL without termination or passthrough....).

submariner/centralized-control-plane.md
elevran: without ACLs the **exposed** service's pods can be reached. Same as Submariner - you can access a service that has an export/import and is not blocked.

submariner/centralized-control-plane.md
```diff
@@ -0,0 +1,298 @@
...
+- `mcs:ServiceExport` and `mcs:ServiceImport` are references for objects in the same namespace
+  and thus can not be used directly for independent naming. A workaround (barring changes to
+  the MCS specification), is to replicate the equivalent `axon` objects to the workload clusters
+  and create the MCS objects locally in each. A better (short term?) alternative would be to use
+  the current Submariner workaround which uses predefined labels and annotations to communicate
+  this information.
+- Full reconciliation is required but not detailed above. For example, `ServiceBinding` status
+  may change over time, as Cluster or Service objects might be deleted, etc.
+- Since we don't propose to leverage any of the Lighthouse `ServiceExport` functionality,
+  we could create a `GlobalIngressIP` object instead of creating `ServiceExport` objects. This
```
sridhargaddam, line 316: In the current implementation, `GlobalIngressIP` is created by Globalnet controller when a Service is exported and Lighthouse is a consumer (i.e., to know the associated globalIP of the service) of this resource. A user is never expected to create the `GlobalIngressIP` object as its owned by Globalnet Controller. 

Based on what is agreed as part of this enhancement proposal, we will have to check the impact on Globalnet as I expect Globalnet to create the `GlobalIngressIP` in response to some resource (in place of serviceExport) and not the other way round.

submariner/centralized-control-plane.md
```diff
@@ -0,0 +1,298 @@
...
+  `ServiceID` and `ObjectRef` _syntax_ - note that we don't know at this point whether
+  the `ObjectRef` refers to a valid Service object). The `mcs:ServiceExport` is created in
+  accordance with the `ClusterID` and `ObjectRef`. We may wish to differentiate between a
+  pending and validated binding (i.e., before the `ObjectRef` is determined valid and after).
+  These may be captured in the `Status.Conditions` or by adding a `PendingClusters` array
+  to the `Service.Status` structure. A workload cluster agent (either a modified Lighthouse
+  component or an entirely new workload cluster controller), retrieves the `axon:ServiceBinding`
+  object and creates the corresponding `mcs:ServiceExport` object in the correct namespace.
+  The workload cluster agent can also update the binding status into the Broker cluster.
+1. `axon:ServiceImport` object trigger the creation of an `mcs:ServiceImport` in the
```
sridhargaddam, line 283: What would be the behaviour if the service with the same name exists in the importing workload cluster? 
When clients/applications access the service, will the DNS get resolved to local service or the imported service?

elevran: good question. I can see several options:
- declare it an error and not create the local endpoint if it exists
- merge the local and global endpoints into a single service
- use a different DNS domain for local and global services (e.g., `*.svc.cluster.local` and `*.svc.global`) to clearly delineate at the client request and avoid the overlap.
(or we make this configurable on the `axon:ServiceImport`/`axon:Service` level to select the desired behavior).

Thoughts?

submariner/centralized-control-plane.md
elevran: @sridhargaddam - thanks for clarifying. I wasn't clear, but the idea is for another controller to allocate the IP... We definitely need to better understand the impact on GlobalNet and ensure we don't break its expected behavior

markmc:
AIUI, this proposal to add a new virtual service and service binding abstraction is mostly about trying to address some limitations imposed by the MCS ClusterSet namespace sameness assumption.

(I must admit I’m not yet clear on how it addresses the case of cluster owners joining a cluster set, but without trusting the owners of other clusters in that set - e.g. how we’d separate limit the permissions of these tenants-within-a-cluster-set on the broker side to their own resources)

The current model, addressing the “shared infrastructure” case:

1. Infrastructure joins clusters A, B, and C into a cluster set on the broker
2. Service owner defines a Service `foo` in namespace `bar` on cluster A
3. Service owner defines a ServiceExport for `foo.bar` on cluster A
4. A ServiceImport for `foo.bar` is created on clusters B and C
5. Service consumers on cluster B and C can access `foo.bar.svc.clusterset.local`

One thing that’s nice about this model is that it’s clear what happens if the service owner wants to withdraw the service - they delete the ServiceExport and the ServiceImports get deleted from the other clusters.

This proposed model for the “federated” case:

1. The owners of clusters A, B, and C join a cluster set on the broker, but in a “no namespace sameness” model
2. Cluster A owner defines a Service foo in namespace bar on cluster A
3. Cluster A owner defines (on the broker) a virtual Service `A-bar-foo` and a ServiceBinding linking `A-bar-foo` to `A:foo.bar`
4. Nothing happens on cluster B or C
5. Cluster B owner defines (again on the broker) a ServiceImport making `A-bar-foo` available as `fu.baa` on cluster B
6. A ServiceImport `fu` in namespace `ba` is created on cluster B
7. Service consumers on cluster B and C can access `fu.baa.svc.clusterset.local`

The service withdrawal case is less clear in this model - cluster A owner can delete the virtual service and service binding, but ServiceImport defined by cluster B owner will dangle.

Could a “namespace linking” approach somewhat like Skupper’s model work? The cluster B owner who wants to import services from a namespace on cluster A just designates a specific namespace on cluster B where any exported services from cluster A should land:

1. The owners of clusters A, B, and C join a cluster set on the broker, but in a “no namespace sameness” mode
2. Cluster A owner defines a Service `foo` in namespace bar on cluster A
3. Cluster A owner defines a ServiceExport for `foo.bar` on cluster A
4. Nothing happens on cluster B or C
5. Cluster B owner creates a NamespaceLink on the broker linking namespace `baa` in cluster B to namespace `bar` in cluster A
6. A ServiceImport for `foo.baa` is created on cluster B
7. Service consumers on cluster B can access `foo.baa.svc.clusterset.local`

The service withdrawal case is as simple as the shared infrastructure case - simply delete the ServiceExport, and any ServiceImports will disappear.

HTH!

donaldh:
Following up from @markmc 's comment https://github.com/submariner-io/enhancements/pull/61#issuecomment-1022444782 the idea of following the skupper model could be easier understand from an admin perspective. I'm sure we could define the semantics of dangling imports, etc. but there is a concern that namespace overrides could become hard to understand at scale.

Join-by-namespace with namespace mapping can address two problems:

1. Selective import – only import from explicitly joined namespaces
2. A degree of namespace decoupling, allowing for operational teams to make their own namespace choices.

> The cluster B owner who wants to import services from a namespace on cluster A just designates a specific namespace on cluster B where any exported services from cluster A should land

My concern with mapping any exported services from cluster A into a designated namespace in cluster B is that you could get service name collisions. I think it would be preferable to selectively map export namespaces to import namespaces. It could be thought of as a selective join.

elevran:
Thanks for the review and suggestions @markmc.
I'll try to address your comments, please let me know if it's clear.

> AIUI, this proposal to add a new virtual service and service binding abstraction is mostly about trying to address some limitations imposed by the MCS ClusterSet namespace sameness assumption.
> 

This is indeed one goal of the revised model. The other is to allow selective importing of services, where MCS mandates that they're automatically imported into all clusters (having the import namespace). The proposal seeks to make this behavior
policy driven (e.g., import-to-all, skupper-like, etc. are example policies. Implementation is left to the "MCS" controller).

> (I must admit I’m not yet clear on how it addresses the case of cluster owners joining a cluster set, but without trusting the owners of other clusters in that set - e.g. how we’d separate limit the permissions of these tenants-within-a-cluster-set on the broker side to their own resources)

This needs to be defined (though I'm not fully clear on how MCS/Submariner limit access to only the relevant objects in the broker cluster...). We are assuming configuration will come from a central management cluster (akin to the Broker, but with active controllers facilitating the sharing). Permission should be set there but you're correct that it is currently not specified.

> 
> The current model, addressing the “shared infrastructure” case:
> 
> 1. Infrastructure joins clusters A, B, and C into a cluster set on the broker
> 2. Service owner defines a Service `foo` in namespace `bar` on cluster A
> 3. Service owner defines a ServiceExport for `foo.bar` on cluster A
> 4. A ServiceImport for `foo.bar` is created on clusters B and C
> 5. Service consumers on cluster B and C can access `foo.bar.svc.clusterset.local`
> 
> One thing that’s nice about this model is that it’s clear what happens if the service owner wants to withdraw the service - they delete the ServiceExport and the ServiceImports get deleted from the other clusters.
>
> This proposed model for the “federated” case:
> 
> 1. The owners of clusters A, B, and C join a cluster set on the broker, but in a “no namespace sameness” model
> 2. Cluster A owner defines a Service foo in namespace bar on cluster A
> 3. Cluster A owner defines (on the broker) a virtual Service `A-bar-foo` and a ServiceBinding linking `A-bar-foo` to `A:foo.bar`
> 4. Nothing happens on cluster B or C
> 5. Cluster B owner defines (again on the broker) a ServiceImport making `A-bar-foo` available as `fu.baa` on cluster B
> 6. A ServiceImport `fu` in namespace `ba` is created on cluster B
> 7. Service consumers on cluster B and C can access `fu.baa.svc.clusterset.local`
> 
> The service withdrawal case is less clear in this model - cluster A owner can delete the virtual service and service binding, but ServiceImport defined by cluster B owner will dangle.

In both cases you need to "reference count" the imports so that you don't leave dangling service or withdraw a service when you shouldn't (e.g., consider the case of both clusters A and B exporting the service in MCS and C importing it. The service should continue to exist in C after A removes the export, but only B's should be accessible. MCS uses the cluster name in the `ServiceImport.Status.Clusters` for that). The controller in the management cluster should address book-keeping removing the ServiceImport only when there are not more ServiceBindings.

> 
> Could a “namespace linking” approach somewhat like Skupper’s model work? The cluster B owner who wants to import services from a namespace on cluster A just designates a specific namespace on cluster B where any exported services from cluster A should land:
> 
> 1. The owners of clusters A, B, and C join a cluster set on the broker, but in a “no namespace sameness” mode
> 2. Cluster A owner defines a Service `foo` in namespace bar on cluster A
> 3. Cluster A owner defines a ServiceExport for `foo.bar` on cluster A
> 4. Nothing happens on cluster B or C
> 5. Cluster B owner creates a NamespaceLink on the broker linking namespace `baa` in cluster B to namespace `bar` in cluster A
> 6. A ServiceImport for `foo.baa` is created on cluster B
> 7. Service consumers on cluster B can access `foo.baa.svc.clusterset.local`
> 
> The service withdrawal case is as simple as the shared infrastructure case - simply delete the ServiceExport, and any ServiceImports will disappear.
> 
> HTH!

That's a good idea. We haven't specified import policies, but Skupper like namespace linking could definitely be an option. This too should address the book-keeping of having multiple exports mapped to a single import.

elevran:
> Following up from @markmc 's comment #61 (comment) the idea of following the skupper model could be easier
> understand from an admin perspective. I'm sure we could define the semantics of dangling imports, etc. but there is
> a concern that namespace overrides could become hard to understand at scale.

Regarding dangling import - while we can (and should) consider the ways to avoid that (e.g., book-keeping mentioned in previous comment), I'd like to point out that this is consistent with k8s Service behavior. AFAIU, a Service can be defined and is resolved even without Pods (could be due an error in selector or if the service is deployed before the backing Pods).

I agree that it could be difficult to understand the full mapping at scale, it is worthwhile to consider that we propose a level of indirection between exports and imports through an axon:Service. This means that we can clearly split the mapping in two:
- which clusters export the Service? and 
- which clusters import the Service?

Seems that visualization of those relations would not be as complicated.

While we've alluded to Import and Export policies in the proposal, we should perhaps elaborate further to show how we can use them to simplify the UX to mimic the Submariner/MCS or Skupper behaviors. Perhaps also show how to migrate a Submariner deployment to Axon CP. I'll add these notes into the proposal for clarity.

markmc:
To put my question a bit more directly - could the "namespace linking" approach address a substantial subset of the requirements here, with less complexity?

(And to be super clear, I ask this because the proposal strikes me as quite a significant jump in complexity. And potentially a simpler approach could progress more easily)

To take the example in the diagram:

- The `C2:app:OrderMgr` and `C4:prod:Database` `ServiceExports` in cluster 2 and 4 would be unchanged
- Instead of `axon:Service` and `axon:ServiceBinding` on the config cluster, you'd have e.g. `ServiceNamespaces` denoting that `C2:app` and `C4:prod` are set up to export services
- Instead of `axon:ServiceImport` on the config cluster, you'd have e.g. `ServiceNamespaceJoins` denoting that any services exported from `C2:app` and `C4:prod` should be imported into `C1:app`

To compare this model with the goals:

- Independent service naming - the importing user would have control over how the imported service is named, by having control over the destination namespace
- Selective imports - the importing user would have control over what clusters/namespaces to import from
- Centralized control - I'm not sure I really understand this as a goal (as opposed to e.g. creating separate administrative domains), but it appears it would achieve it

Then again, I'm not a submariner developer, so YMMV!

donaldh:
I am inclined to agree that the proposal is a jump in complexity. I think that is in part due to the fact that we have potentially conflated two things:

1. Finer grained control of multi-cluster services
2. A central control point for a multi-cluster orchestrator

I think the reason for combining the problems is an attempt at creating a single definition and single policy control point for a service that happens to be federated across multiple clusters. It's almost like we want a ServiceClass, somewhat like a StorageClass or a GatewayClass, so that we can say `ClusterA/app:svc` and `ClusterB/prod:svc` are both instances of the same ServiceClass, without implicitly matching by namespace:name. Perhaps this could be achieved by using labels to define an alias or service identity for associating service instances?

The core goals for extensions to MCS are:

* Selective imports
* Independent naming
* Separate admin domains

This leads me to think that we could use cluster level policy at the importing cluster to define:

* import all (default?)
* import namespace `foo`
* import from namespace `foo` to namespace `bar`
* no import

These would all enable automatic service import using the desired policy behaviour. Alternatively directly create a ServiceImport `app:svc` that imports from `foo:svc` for fine grained control.

submariner/centralized-control-plane.md
```diff
@@ -0,0 +1,386 @@
...
+We would like to support Multi-Cluster Services in the following additional use cases (see
+also Context section below.):
+
+- **Separate administrative domains**: a deployment model where clusters are owned and operated
+  by different teams, each having independent control and configuration.
+- **Global load balancing**: allow a global load balancer to use services in different clusters.
+- **Non flat networks**: exported services are accessible via an ingress point, such as Ingress,
+  NodePort, LoadBalancer, etc., and not directly using IP addresses of backing Pods.
+- **Centralized management plane**: service sharing is managed from a central location, in a
+  controlled manner (i.e., only identified services in a catalog may be shared).
```
mkolesnik, line 35: These are all outside the scope of Submariner itself, each point for it's own reasons.

Since the section details "We would like to support Multi-Cluster Services in the following additional use cases", this implies that the desired capabilities should be part of Submariner and in it's scope.

Hence it seems to me this proposal as a whole should live outside of submariner, and specific enhancements which fall in Submariner's scope.

submariner/centralized-control-plane.md
```diff
@@ -0,0 +1,386 @@
...
+  version of testing in the same or different NS, using the same or different name, etc.
+
+### High Level Changes and Relation to MCS API
+
+1. Independent service naming (i.e., allow use of different names in different clusters).
+  This could be done without changes to MCS specification by setting the target Service name
+  and namespace as labels or annotations on the `ServiceImport` object. Alternately, we could
+  extend the objects' `Spec` to explicitly support a different target service name and namespace.
+  Backward compatibility is provided by defaulting to the object namespace and name if the
+  `Spec` does not define a different destination.
```
mkolesnik, line 49: This should be raised as a proposal to MCS to extend the CRD specification.

submariner/centralized-control-plane.md
```diff
@@ -0,0 +1,386 @@
...
+  applications running on VMs or bare metals, and not naturally part of a Kubernetes Service).
+- **TBD**: migration of services, brownfield with existing names, attempting to coordinate
+  use of namespaces across all clusters has overhead, merging infra after M&A, release of a new
+  version of testing in the same or different NS, using the same or different name, etc.
+
+### High Level Changes and Relation to MCS API
+
+1. Independent service naming (i.e., allow use of different names in different clusters).
+  This could be done without changes to MCS specification by setting the target Service name
+  and namespace as labels or annotations on the `ServiceImport` object. Alternately, we could
```
mkolesnik, line 46: If you implement a separate service discovery solution in lieu of Lighthouse, you can have service targeting however you want and theoretically aren't bound to MCS API anyhow.

submariner/centralized-control-plane.md
```diff
@@ -0,0 +1,386 @@
...
+  what local workload clusters attributes). These are similarly implementation specific.
+
+While all three (naming, synchronization and policies) could be considered as possible
+ enhancements to the MCS API, it would be worthwhile to experiment with different approaches,
+ such as the one defined here, before attempting to reach community consensus on MCS API changes.
+
+The design proposal attempts to achieve the above goals minimal changes to workload
+ clusters, especially with respect to data plane configuration. Ideally, the control
+ plane changes proposed will are extensions to MCS or implemented as a layer above
+ MCS and confined to the Broker cluster only.
```
mkolesnik, line 84: It is not intended to deviate from MCS API in Lighthouse, but if you implement your own service discovery layer you can define your own behaviors that will fit your needs. This shouldn't be part of submariner though.

submariner/centralized-control-plane.md
mkolesnik: Then on the L3 connectivity aspect there aren't any expected changes IIUC, and all the discussion is regarding how service discovery is handled.

It's just that the way this section is described is not clear about this point, and seems like the expectation is isolation even on the L3 level.

submariner/centralized-control-plane.md
```diff
@@ -0,0 +1,386 @@
...
+
+## Detailed Proposal
+
+### Goals
+
+- Enable independent service name and namespace in workload clusters.
+- Enable a Service to be imported into a subset of the workload clusters only.
+- Enable centralized control over definition of Service exports and imports.
+- Enable more decoupled binding (i.e., explicit and not only based on names)
+  between existing network endpoints and the service being shared.
```
mkolesnik, line 124: As stated previously, all of these can be achieved by writing a different service discovery layer other than Lighthouse.

This shouldn't be part of Submariner core though, but if you feel that parts of Lighthouse can be broken into micro-services (eg DNS server component) such that they can be reused in such a solution, we can discuss that in the scope of Submariner, to make sure work is not being replicated needlessly.

mkolesnik (changes requested):

elevran:
Thank you all for the detailed reviews.
It does not seem this proposal is aligned with Submariner's scope and therefore it is withdrawn.
Will try to tease out subtasks and smaller changes in separate proposals which may be better aligned.

PR #69, opened 2022-03-01
Multiple active gateway enhancement proposal

anfredette:
Submariner currently only allows a single gateway to be active at
any one time in a given cluster.  This enhancement proposes that
there be an option to allow more than one gateway to be active in
a given cluster to enhance both performance and resiliency. This
feature is sometimes referred to as “active-active gateways”;
however, the goal is to allow two or more active gateways to be used.

Signed-off-by: Andre Fredette <afredette@redhat.com>


submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,100 @@
+# Multiple Active Gateways
+
+## Summary
+
+NOTE: The design details are still a work in progress, so this is currently a draft proposal.
+
+Submariner currently only allows a single gateway to be active at any one time in a given cluster.
+This enhancement proposes that there be an option to allow more than one gateway to be active in
+a given cluster to enhance both performance and resiliency. This feature is sometimes referred to as
+“active-active gateways”; however, the goal is to allow two or more actove gateways to be used.
```
astoycos: /s/actove/active

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Proposal
+
+The proposal is to optionally allow multiple gateways to be active at the same time in a given cluster.
+
+When only a single active gateway is used, all inter-cluster traffic must flow through that one gateway,
+which creates a performance bottleneck. Enabling more than one node to be used as a gateway can improve
```
astoycos, line 19: Maybe go in a little more about how this is a performance bottleneck specifically with encryption engines? I.e All encryption is locked to one core etc.

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,100 @@
...
+![SingleGateway](./images/single-gateway.png)
+
+### Multiple Active Gateway Design
+
+The proposal for supporting multiple active gateway nodes is shown in the following diagram and described below.
+
+![MultiGateway](./images/multi-gateway.png)
+
+* Each non-gateway node will establish a VXLAN tunnel to each gateway node.
+* The non-gateway nodes will load-balance flows between the active gateway nodes.
```
astoycos, line 49: Following the POC It will be good to have an "implementation" section going into the details of these bits

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,100 @@
...
+
+![SingleGateway](./images/single-gateway.png)
+
+### Multiple Active Gateway Design
+
+The proposal for supporting multiple active gateway nodes is shown in the following diagram and described below.
+
+![MultiGateway](./images/multi-gateway.png)
+
+* Each non-gateway node will establish a VXLAN tunnel to each gateway node.
```
astoycos, line 48: Do we have any understanding of the performance impact with the configuration of so many more tunnels per clusterset?

submariner/multiple-active-gateways.md
anfredette: Agreed. I was mainly trying to get a high-level description documented here. We should add more details once we have a few more things figured out.

submariner/multiple-active-gateways.md
anfredette: This is a good point and something that should be covered.  We don't have hard numbers.  However, the goal and expectation are that this will improve forwarding performance. Initial setup could take a little longer for each new endpoint (peer cluster), but it shouldn't be too big of a hit.  However, this could increase the total number of tunnels for a clusterset significantly.  For N nodes, we currently have N(N-1) tunnels. If there are M gateways on each node, we multiply that by 2M^2.  So, I'm thinking that for large deployments, we'd only want to use multiple gateways in a set of high speed "core" clusters, and then connect "leaf" clusters in with fewer connections to a subset of core clusters.  However, this will require that the gateways have "transit node" functionality and that we use dynamic routing, so that's for a future enhancement :-).

submariner/multiple-active-gateways.md
anfredette: Ok.

anfredette:
Thanks for your feedback @astoycos!

submariner/multiple-active-gateways.md
anfredette: So, in summary, I do have some scalability concerns for large deployments if we don't also make other enhancements. We'll have to analyze the impact.

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,100 @@
...
+* The non-gateway nodes will load-balance flows between the active gateway nodes.
+* Each gateway node will establish a tunnel to each peer gateway node.
+* The gateway nodes will load-balance flows between the active peer gateway nodes.
+* Load balancing will be performed using Linux resilient nexthop groups.
```
sridhargaddam: Currently the Active Submariner Gateway pod is responsible for couple of things like
1. Creating local endpoint and cluster objects 
2. Syncing the endpoints/cluster objects with the Broker Cluster
3. Healthchecks with remote cluster gateway nodes
4. Performing NAT discovery to identify if remote cluster is reachable over private or public-ip in the endpoint.
5. Gateway object syncer.
6. Supporting metrics

Since some of the above operations have to be done by only one component, can you please add notes on how the above operations will be handled in the new Active/Active architecture.

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,100 @@
+# Multiple Active Gateways
+
+## Summary
+
+NOTE: The design details are still a work in progress, so this is currently a draft proposal.
+
+Submariner currently only allows a single gateway to be active at any one time in a given cluster.
+This enhancement proposes that there be an option to allow more than one gateway to be active in
```
sridhargaddam, line 10: In this new design, will we be using the same submariner gateway label to mark a node[/s] as Gateway nodes?

And all the labelled Gateway nodes will automatically be part of Active/Active gateways?

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Proposal
+
+The proposal is to optionally allow multiple gateways to be active at the same time in a given cluster.
+
+When only a single active gateway is used, all inter-cluster traffic must flow through that one gateway,
+which creates a performance bottleneck. Enabling more than one node to be used as a gateway can improve
+the performance linearly in the number of gateways used.
+
+Additionally, when the active gateway fails in the current implementation, work must be done to switch
```
sridhargaddam, line 22: Can you please describe how we will figure of when the active gateway fails?

PS: In the current design, we use the endpoint objects to identify the active gateway nodes. When a new node becomes the active gateway node, it deletes the old/stale endpoint.

nyechiel (commented):
I have a general comment about this proposal: based on user's feedback so far, it's not clear if the gateway model (where we assume that dedicated nodes or certain worker nodes have public IPs available) is really the future of Submariner. This model is not very cloud-native, cumbersome to setup, and is not compatible/realistic with many of the managed Kubernetes offerings which are becoming more popular. For this reason, we started some early exploration work around using a `LoadBalancer` type service instaed (see also https://submariner.io/getting-started/architecture/gateway-engine/). I am wondering if we shouldn't complete the investigation of the `LoadBalancer` mode first, before further investing into HA of the gateway model.

sridhargaddam:
> I have a general comment about this proposal: based on user's feedback so far, it's not clear if the gateway model (where we assume that dedicated nodes or certain worker nodes have public IPs available) is really the future of Submariner. This model is not very cloud-native, cumbersome to setup, and is not compatible/realistic with many of the managed Kubernetes offerings which are becoming more popular. For this reason, we started some early exploration work around using a `LoadBalancer` type service instaed (see also https://submariner.io/getting-started/architecture/gateway-engine/). I am wondering if we shouldn't complete the investigation of the `LoadBalancer` mode first, before further investing into HA of the gateway model.

+1 
This is inline with what I mentioned earlier on couple of occasions that LB mode is going to be the future.

submariner/multiple-active-gateways.md
anfredette: @astoycos and @Billy99 are working on a POC of this now and we'll update this proposal when we work it out.

submariner/multiple-active-gateways.md
anfredette: That is the plan.  We think it's simpler than having some nodes be active gateways, while others are backups.  However, we think enabling the later could be useful.  Some options could include explicitly labeling active/backup, or have a setting that allows us to configure N active gateways, and then have the first N go active, and any additional gateways would be passive unless and until one of the active gateways fails.

submariner/multiple-active-gateways.md
anfredette: I don't have an answer right now, but we'd know this needs to be worked out.  We can add it to TODO list.

mkolesnik:
+1 to @nyechiel's assessment.

Additionally, LB mode would have many of the problems of multiple gateways (since we'd probably want to run more than one GW pod).
I'm suggesting we focus on LB mode first, and that would also pave the way to an easier implementation of multiple dedicated gateways, should such a feature be desired.

anfredette:
> I am wondering if we shouldn't complete the investigation of the `LoadBalancer` mode first, before further investing into HA of the gateway model.

We definitely agree that LoadBalander mode is essential and probably should be a requirement for the first release.  However, we discussed this with @sridhargaddam and we believe that all of the infrastructure work we need to do for non-Loadbalancer mode will also be required for LoadBalancer mode, and the non-LoadBalancer mode development environment is much more straightforward. So, we are choosing to work on non-LoadBalancer mode first.  Once we get that fully working, we can turn our attention to LoadBalancer mode.

mkolesnik:
> We definitely agree that LoadBalander mode is essential and probably should be a requirement for the first release. However, we discussed this with @sridhargaddam and we believe that all of the infrastructure work we need to do for non-Loadbalancer mode will also be required for LoadBalancer mode, and the non-LoadBalancer mode development environment is much more straightforward. So, we are choosing to work on non-LoadBalancer mode first. Once we get that fully working, we can turn our attention to LoadBalancer mode.

What is the benefit to the users from active active dedicated ha gateways?

We already see that LB mode is necessary for deployments where you cant control the infrastructure (any hosted K8s) and also as @nyechiel mentioned the "dedicated gw" mode is not K8s native and might even be problematic in many cases. All the while, LB mode should provide the same benefits while being much more native to the K8s ecosystem and easier to set up and maintain by the users.

anfredette:
> What is the benefit to the users from active active dedicated ha gateways?

The term "LoadBalancer" in LoadBalancer mode is currently a bit of a misnomer.  The code still only supports a single active gateway sitting behind that LB.  The LB is just used as a mechanism to get an external address in certain cloud deployments.  With this change, we can support multiple active gateways behind the LB.  I attempted to explain the benefits of multiple active gateways in the initial proposal, but basically, they are increased forwarding performance and resiliency.

submariner/multiple-active-gateways.md
yboaron: It seems that the painful points in the current solution are performance bottleneck and slow failover time.

While the proposed solution addresses the performance bottleneck, it complicates the datapath handling both intra-cluster and inter-cluster. 

Are there other alternatives (e.g: setting more resources somehow to vpn pod, etc) to mitigate the performance bottleneck?

Do we know how critical is the failover timing? the maximum time for GW failover? 
Is it possible to reduce GW failover timing with Active/SB architecture?  (Maybe we can use VIP to represent the active GW? )

mkolesnik:
> The term "LoadBalancer" in LoadBalancer mode is currently a bit of a misnomer. The code still only supports a single active gateway sitting behind that LB.

Yes but current HA is supported so users can still expect a decent level of service.

> The LB is just used as a mechanism to get an external address in certain cloud deployments. 

Yes but that would be of great value in many use cases where its not possible to deploy Submariner otherwise, or where other considerations such as cost and/or security would have users rather use the native infrastructure LB rather than going with dedicated gateways with external IPs.

The value add of a working LB mode is quite high.

> With this change, we can support multiple active gateways behind the LB. I attempted to explain the benefits of multiple active gateways in the initial proposal, but basically, they are increased forwarding performance and resiliency.

We're talking about extreme scenarios when a flaiover occurs, right?

Currently the HA has a disruption of a few seconds AFAIK so I'd say the value add in this area is very low and is more of a stretch goal.

As for performance I agree there's a better use case here, but it can also be mitigated to some extent by making sure Submariner is deployed on more suitable hardware with higher networking bandwidth and/or stronger processors.

So the value add here is medium IMHO.

Hence the initial assessment of "LB mode is higher priority" still holds. Once that mode is stable we can work on active-active and integrate it into the new model.

submariner/multiple-active-gateways.md
mkolesnik: This seems like something we should design for as it could get quite complicated.

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,100 @@
...
+and new routes, which takes time. With the proposed design, It is expected that the fail-over times can
+be significantly improved if additional gateways are operational. Furthermore, it is expected that
+traffic flowing through operational gateways will continue to flow while the traffic from the failed
+gateway is redirected to the operational gateways. Both properties reduce the impact of a failure.
+
+## Design Details
+
+### Single Gateway
+
+In the current single gateway Submariner implementation, intra-cluster VXLAN tunnels are established by
```
mkolesnik, line 33: This is only true for `iptables` based CNIs, other CNIs don't necessarily need these internal tunnels.

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,100 @@
...
+### Multiple Active Gateway Design
+
+The proposal for supporting multiple active gateway nodes is shown in the following diagram and described below.
+
+![MultiGateway](./images/multi-gateway.png)
+
+* Each non-gateway node will establish a VXLAN tunnel to each gateway node.
+* The non-gateway nodes will load-balance flows between the active gateway nodes.
+* Each gateway node will establish a tunnel to each peer gateway node.
+* The gateway nodes will load-balance flows between the active peer gateway nodes.
```
mkolesnik, line 51: What does this mean?

submariner/multiple-active-gateways.md
mkolesnik: A link to some documentation, or even a brief explanation, of "Linux resilient nexthop groups" will help understand what's expected to be happening

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,100 @@
...
+![MultiGateway](./images/multi-gateway-flow.png)
+
+1. Node 1-1
+    1. Intercept packet destined for global service
+    2. Load balances flow to gateway 1-1
+    3. Hash ensures remaining packets in the connection also go to gateway 1-1
+2. Gateway 1-1
+    1. Determines packet is destined for Cluster 2 based on dest Globalnet IP
+    2. Load balances flow to Gateway 2-2
+    3. Perform SNAT from local source/port to one of the Globalnet address/ports for gateway 1-1
```
mkolesnik: This needs to conform with Globalnet 2.0

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,100 @@
...
+* Load balancing will be performed using Linux resilient nexthop groups.
+
+### Multiple Active Gateway Packet Flow
+
+An example packet flow is shown in the following diagram and described below.
+
+![MultiGateway](./images/multi-gateway-flow.png)
+
+1. Node 1-1
+    1. Intercept packet destined for global service
```
mkolesnik: Is there an implicit assumption that globalnet is deployed? We need to support deployments both with and without it

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,100 @@
...
+    2. Load balances flow to gateway 1-1
+    3. Hash ensures remaining packets in the connection also go to gateway 1-1
+2. Gateway 1-1
+    1. Determines packet is destined for Cluster 2 based on dest Globalnet IP
+    2. Load balances flow to Gateway 2-2
+    3. Perform SNAT from local source/port to one of the Globalnet address/ports for gateway 1-1
+    4. Hash ensures that remaining packets in the flow also go to Gateway 2-2
+3. Gateway 2-2
+    1. The packet is handed off to the local service load balancer, which performs necessary NAT
```
mkolesnik: Since you mentioned globalnet, it should also perform NAT here for the destination, before its further processed

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,100 @@
...
+operations and forwards the packet to chosen endpoint on Node 2-1.
+4. Node 2-1
+    1. Send response back to Gateway 2-2 (through service load balancer)
+    2. Note: it should be okay if the return path uses a different gateway, but I think the cluster
```
mkolesnik: How do you make sure in this case it routes back through the same gateway? Or is it automatic?

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,100 @@
...
+load balancer will return it to the same node in order to maintain the NAT state.
+5. Gateway 2-2
+    1. Forward back to Gateway 1-1 over the appropriate tunnel
+    2. Note: It is important that Gateway 2-2 forwards the packet directly to Gateway 1-1 and not
```
mkolesnik: Again its not clear hos you determine to send it back to the same gateway?

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,100 @@
...
+via the nexthop load balancer which may send it to Gateway 1-2. This can be accomplished by using host
+routes for the gateway IP addresses that are higher priority than the routes for the cluster Globalnet
+CIDRs which would use the nexthop load balancing.
+6. Gateway 1-1
+    1. Perform DNAT from the global address back to the local address
+    2. Send to the pod on Node 1-1 via local CNI.
+
+## Notes and Assumptions
+
+* This design assumes that Globalnet is being used. It may be possible to make it work without Globalnet if
```
mkolesnik: It's essential this works without globalnet.

Moreover, globalnet is an optional layer on top of core Submariner so I'd say to get it working without globalnet is higher priority.

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,100 @@
...
+* The initial POC will use standard external IP addresses for gateways, but load balancer mode is
+a high priority for the final implementation.
+
+* The initial design and POC is focused on non-ovn-kubernetes CNIs, but ovn-kubernetes must also be supported.
+
+* There will be a config option to enable multiple active gateways, and when enabled, all nodes labeled with
+“submariner.io/gateway=true” will become active. It is for future study whether we should support an option
+where we support more than one active gateway as well as additional backup inactive gateways.
+
+## Plan
```
mkolesnik: I don't see any alternatives, nor do I see any detailed design (except a theoretical example packet flow)

mkolesnik (changes requested):

anfredette:
FYI, we'll be pushing an update to the proposal in the next day or two which hopefully will address many of the comments/questions.  Thanks, Andre

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,660 @@
...
+drawing, new routes need to be added to remote
+cluster gateways (GW-4 in the drawing) that include the `ClusterGlobalEgressIPs` IP Addresses from each sending
+node so that traffic is returned to the sending gateway (GW-1 in drawing) so that conntrack state of the packet
+is maintained and the proper DNATing is performed (see Step 5-c). To enable these routes to be written, the
+`ClusterGlobalEgressIPs` IP Addresses must be known on all gateways.
+
+##### Option 1: Add ClusterGlobalEgressIP IP Addresses to Submariner EndPoint
+
+The Submariner Endpoint CRD Object is currently already shared across clusters. A new field will be added
+to the existing `submariners` CRD object and used to store and disseminate the IP Addresses to the remote
```
Billy99:
```suggestion
to the existing Submariner `endpoint` CRD object and used to store and disseminate the IP Addresses to the remote
```

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,660 @@
...
+
+   a. Traffic is DNAT'ed (reverse the SNAT in 3-b) back to one of the Original GW node’s `clusterglobalEgressIP`.
+
+    `SRC: 10.0.128.5 DST: 242.254.1.1`
+
+   b. Traffic is SNAT'ed (reverse the DNAT in 3-a) back to external IP of the k8s service mirrored by submariner.
+
+    `SRC: 242.254.2.254  DST: 242.254.1.1`
+
+   c. Routed back to the Original gateway using specific routes have been added for ClusterGlobalEgressIPs that take
```
Billy99:
```suggestion
   c. Routed back to the Original gateway using specific routes that have been added for ClusterGlobalEgressIPs that take
```

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,660 @@
...
+
+    `SRC: 242.254.2.254  DST: 10.1.128.4`
+
+## Data Model Changes
+
+### Enabling Multiple Active Gateways Feature
+
+Multiple Active Gateways is an optional feature that can be enabled when a cluster is joined to the broker.
+The user can indicate whether they wish the feature to be enabled or not using a new flag added to the
+subctl join` command.
```
Billy99:
```suggestion
`subctl join` command.
```

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,660 @@
...
+
+Both the Gateway and Globalnet pods are launched as a DaemonSet, where pods are scheduled on any node with the
+label `submariner.io/gateway=true`. As mentioned above, the Gateway pods run based on the leadership semaphore.
+The Globalnet pods on the other hand run in either a Gateway (running on the active Gateway) or Non-Gateway
+(running on the standby Gateway) mode. In both modes, the Gateway Monitor Controller watches all Submariner
+Endpoints and for remote Endpoints, updates the IP Table rules to mark packets destined for the remote cluster.
+
+In the Gateway mode case, a lot more is done. The IP Pool is created and all the other controllers are started.
+The IP Pool is used to reserve IP Addresses used by Globalnet (ClusterGlobalEgressIP, GlobalEgressIP, and
+GlobalIngressIP). Whenever IP Addresses are reserved or released by one of the controllers, associated IP Table
```
Billy99:
```suggestion
GlobalIngressIP). Whenever IP Addresses are reserved or released by one of the controllers, associated datapath
```

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,660 @@
...
+requesting IP allocation are GlobalEgressIP and GlobalIngressIP, which are per cluster.
+
+The proposed solution is to move the Globalnet pod from being a Daemonset that mirrors the Gateway Daemonset
+to a Kubernetes Deployment with a ReplicaSet of 1. With only one instance of the Globalnet pod, there will
+only be one IP Allocator and one instance of each of the controllers that are requesting the IPs.
+
+Since the Globalnet pod is not running on every node that needs local datapath updates, the Globalnet pod cannot
+update the datapath rules as it currently does. The second part of the proposal is to move datapath rule management
+out of the Globalnet pod and into the Gateway pod, which runs on each Gateway node. The controllers running in
+Globalnet pod will still manage IP allocation and the CRD object management, just remove the IP Table updates
```
Billy99:
```suggestion
Globalnet pod will still manage IP allocation and the CRD object management, just remove the local datapath updates
```

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,660 @@
...
+from the following controllers in the Globalnet pod:
+
+* Gateway Monitor Controller
+* Node Controller
+* Cluster Global Egress IP Controller
+* Global Ingress IP Controller
+* Global Egress IP Controller
+
+Five new controllers with the same name will be added to the Gateway pod, which watch their respective CRD
+objects and modify the IP Tables as needed on their host. An alternative to this approach could be to add all of the
```
Billy99, line 393:
```suggestion
objects and modify the local datapath as needed on their host. An alternative to this approach could be to add all of the
```

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,660 @@
...
+
+   NOTE: Weave performs connection tracking on flows and only allows responses to TCP sessions with state.  The state for
+the given flow exists on **GW-2**, but not on **GW-1**. Because of this, it is necessary to add the following rule to allow
+the response packets through if they happen to be returned to a gateway other than the one that the outbound flow uses:
+
+   ```shell
+    -A FORWARD -d 10.1.0.0/16 -i vxlan-tunnel -j ACCEPT
+    ```
+
+   An alternative approach is to SNAT packets using egress IPs in a way similar to what is done for Globalnet with
```
Billy99:
```suggestion
**NOTE:** An alternative approach is to SNAT packets using egress IPs in a way similar to what is done for Globalnet with
```

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,660 @@
...
+    :
+    // The list of allocated ClusterGlobaEgresslIPs.
+    AllocatedIPs  []string          `json:"allocated_ips,omitempty"`
+}
+```
+
+The advantage of using Submariner Endpoint is that it is already shared across clusters and instanced properly
+for sharing the desired data.
+
+*NOTE:* This is how the POC was implemented.
```
Billy99:
```suggestion
**NOTE:** This is how the POC was implemented.
```

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,660 @@
...
+
+* Performance testing on POC code.
+* Support non-kubeproxy route agent handlers.
+* Support Loadbalancer mode.
+* Support detection and handling of gateway failure and recovery.
+
+## Links
+
+* Source for images:
+  [Submariner-MAG-Drawings](https://docs.google.com/presentation/d/19W-0YJB0wDiEWDhs7MoLFgKEm5GkGs2g47jXMUnIrv0/edit#)
```
Billy99:
```suggestion
  [Submariner-MAG-Drawings](https://docs.google.com/presentation/d/1PsxbBbzt9A9opO4bVxkJcLnzD6zAQTlKtK2_aq5SINs/edit#)
```
The previous link was the "Axon Data Plane Overview" slides. I added the link before I knew which we were going to use.

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,660 @@
...
+well. Here, all the desired routes are calculated for every relevant remote endpoint event and the existing routes are simply
+overridden with the total correct state.  This is slightly different from the re-written Route Agent reconciler, but could
+be a bit less resource-intensive.  Ultimately we as a community should agree on the architecture and move all of the route
+reconciliation logic to a common library.
+
+### Iptables Architecture Changes
+
+#### Route Agent Iptables Architecture Changes
+
+For the MAG functionality, two main IPtables changes were needed on every GW node.  The first was to ensure that K8’s service
```
Billy99:
```suggestion
For the Multiple Active Gateway functionality, two main IPtables changes were needed on every GW node.  The first was to ensure that K8’s service
```

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,660 @@
...
+Currently, there is one instance of the `clusterglobalegressips` CRD object per cluster. This object is a set of
+globally unique IP Addresses (across all clusters managed by Submariner) one of which is used on the Gateway node
+as the IP Address to SNAT the source address on egress packets. Currently, the one instance of the
+`clusterglobalegressips` CRD object uses the default name of `cluster-egress.submariner.io`.
+
+With Multiple Active Gateways, there will need to be a `clusterglobalegressips` CRD object per Gateway, each of
+which is a globally unique set of IPs. To adjust for multiple instances, the node name is prepended to the existing
+constant instance name.
+
+*(TBD - Do we need to keep non-MAG name of object as is and only prepend node name if MAG enbable? POC always
```
Billy99:
```suggestion
*(TBD - Do we need to keep non-Multiple Active Gateway naming of the object as is and only prepend node name if Multiple Active Gateway is enabled for upgrade reasons? POC always
```

anfredette:
Thanks for the edits @Billy99!

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,660 @@
...
+Currently, there is one instance of the `clusterglobalegressips` CRD object per cluster. This object is a set of
+globally unique IP Addresses (across all clusters managed by Submariner) one of which is used on the Gateway node
+as the IP Address to SNAT the source address on egress packets. Currently, the one instance of the
+`clusterglobalegressips` CRD object uses the default name of `cluster-egress.submariner.io`.
+
+With Multiple Active Gateways, there will need to be a `clusterglobalegressips` CRD object per Gateway, each of
+which is a globally unique set of IPs. To adjust for multiple instances, the node name is prepended to the existing
+constant instance name.
+
+*(TBD - Do we need to keep non-Multiple Active Gateway naming of the object as is and only prepend node name if Multiple Active Gateway is enabled for upgrade reasons? POC always
```
anfredette:
```suggestion
*(TBD - Do we need to keep non-Multiple Active Gateway naming of the object as is and only prepend node 
name if Multiple Active Gateway is enabled for upgrade reasons? POC always
```

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,661 @@
...
+Currently, there is one instance of the `clusterglobalegressips` CRD object per cluster. This object is a set of
+globally unique IP Addresses (across all clusters managed by Submariner) one of which is used on the Gateway node
+as the IP Address to SNAT the source address on egress packets. Currently, the one instance of the
+`clusterglobalegressips` CRD object uses the default name of `cluster-egress.submariner.io`.
+
+With Multiple Active Gateways, there will need to be a `clusterglobalegressips` CRD object per Gateway, each of
+which is a globally unique set of IPs. To adjust for multiple instances, the node name is prepended to the existing
+constant instance name.
+
+*(TBD - Do we need to keep non-Multiple Active Gateway naming of the object as is and only prepend node
```
anfredette:
```suggestion
*(TBD - Do we need to keep non-Multiple Active Gateway naming of the object as is and only prepend node
```

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,661 @@
...
+globally unique IP Addresses (across all clusters managed by Submariner) one of which is used on the Gateway node
+as the IP Address to SNAT the source address on egress packets. Currently, the one instance of the
+`clusterglobalegressips` CRD object uses the default name of `cluster-egress.submariner.io`.
+
+With Multiple Active Gateways, there will need to be a `clusterglobalegressips` CRD object per Gateway, each of
+which is a globally unique set of IPs. To adjust for multiple instances, the node name is prepended to the existing
+constant instance name.
+
+*(TBD - Do we need to keep non-Multiple Active Gateway naming of the object as is and only prepend node 
+name if Multiple Active Gateway is enabled for upgrade reasons? POC always
```
anfredette, line 326:
```suggestion
name if Multiple Active Gateway is enabled for upgrade reasons? POC always
```

submariner/multiple-active-gateways.md
anfredette: Improving the recovery time isn't the main objective.  So, I think of it as more of a bonus.  I'm sure there is room for improvement in the current design/implementation without going to MAG.

submariner/multiple-active-gateways.md
anfredette: This has been described in the updated proposal.

submariner/multiple-active-gateways.md
anfredette: For which CNIs does Submariner not use a tunnel for inter-cluster traffic between non-gateway and gateway nodes?

submariner/multiple-active-gateways.md
anfredette: This has been expanded on in the updated proposal.

submariner/multiple-active-gateways.md
anfredette: Done.

submariner/multiple-active-gateways.md
anfredette: The details have been included in the updated proposal.

submariner/multiple-active-gateways.md
anfredette: The details have been included in the updated proposal.

submariner/multiple-active-gateways.md
anfredette: The design and POC now support non-Globalnet.

submariner/multiple-active-gateways.md
anfredette: The details have been included in the updated proposal.

submariner/multiple-active-gateways.md
anfredette: The details have been included in the updated proposal.

submariner/multiple-active-gateways.md
anfredette: It does.

submariner/multiple-active-gateways.md
anfredette: If there is a decision to move forward with this enhancement, then we can work out the details.

anfredette:
Thanks everyone for the review and comments/suggestions.  We've updated the proposal to reflect our current thoughts on the design, and the POC that has been implemented.  We're currently doing some performance testing and will update the pr when we get some results.

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,663 @@
+# Multiple Active Gateways
+
+## Summary
```
Billy99, line 7:
```suggestion
## Status

Deferred

## Summary
```

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,663 @@
...
+The flag will need to be passed to the `submariner-gateway` and `submariner-globalnet` pods when they are
+started. Within each pod, the flag will be propagated to multiple controllers.
+
+### Non-Globalnet
+
+No data-model changes specific to the non-Globalnet deployment.
+
+### Globalnet
+
+#### Globalnet ClusterGlobalEgressIps Instancing
```
Billy99:
```suggestion
#### Globalnet ClusterGlobalEgressIps, GlobalEgressIps and GlobalIngressIps Instancing

**NOTE:** In the PoC, only the `clusterglobalegressips` CRD Object was updated as described below. However,  `globalegressips` and `globalingressips` CRD Objects will need similar changes.

```

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,663 @@
...
+* Cluster Global Egress IP Controller
+* Global Ingress IP Controller
+* Global Egress IP Controller
+
+Five new controllers with the same name will be added to the Gateway pod, which watch their respective CRD
+objects and modify the IP Tables as needed on their host. An alternative to this approach could be to add all of the
+datapath functionality needed for globalnet to the RouteAgent, however this would result in a much higher level of
+complexity for the RA, and a bigger re-write.
+
+#### Expose ClusterGlobalEgressIPs to Remote Clusters
```
Billy99:
```suggestion
#### Expose ClusterGlobalEgressIPs, GlobalEgressIPs and GlobalIngressIPs to Remote Clusters

**NOTE:** In the PoC, only the `clusterglobalegressips` CRD Object was considered as described below. However,  `globalegressips` and `globalingressips` CRD Objects will need similar changes.

```

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,663 @@
...
+* The multipath routes require an interface, so this approach won't work with IPsec Tunnel Mode.  IPsec Transport mode,
+  proposed in [IPSec transport mode support](https://github.com/submariner-io/enhancements/pull/103) will be required for
+  IPsec encryption.
+
+## TODO
+
+* Performance testing on POC code.
+* Support non-kubeproxy route agent handlers.
+* Support Loadbalancer mode.
+* Support detection and handling of gateway failure and recovery.
```
Billy99, line 742:
```suggestion
* Support detection and handling of gateway failure and recovery.
* In the PoC, only the new ClusterGlobalEgressIP Controller was implemented in the Gateway Pod (See #multiple-active-gateway-changes). The other controllers need to be ported from Globalnet.
```

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,663 @@
...
+
+* The initial POC will use standard external IP addresses for gateways, but load balancer mode is
+  a high priority for the final implementation.
+* The initial design and POC is focused on CNIs that use the kubeproxy route agent handler that establishes VXLAN
+  tunnels between worker nodes and gateways. Other options need to be investigated and supported.
+* The multipath routes require an interface, so this approach won't work with IPsec Tunnel Mode.  IPsec Transport mode,
+  proposed in [IPSec transport mode support](https://github.com/submariner-io/enhancements/pull/103) will be required for
+  IPsec encryption.
+
+## TODO
```
Billy99, line 737: ADD this section. You might get some linting errors. Make sure last section views correctly. There are multi-line bullets.

## PoC

To prove out the architecture described in this document, a PoC was implemented and can be found in:
* https://github.com/submariner-io/submariner/tree/feature-multi-active-gw 
* https://github.com/submariner-io/submariner-operator/tree/feature-multi-active-gw 

Additional details regarding the PoC can be found in the
[README.md](https://github.com/submariner-io/submariner/blob/feature-multi-active-gw/README.md) 
on the feature branch.


The PoC implements Multiple Active Gateways with both Globalnet and non-Globalnet deployments.
Use the following commands to deploy in a set of KIND clusters. The globalnet flags are optional.
Because there are now additional Gateways, additional nodes need the `submariner.io/gateway=true` label.

Using `make deploy`:

```
make deploy using=lighthouse,vxlan,globalnet DEPLOY_ARGS='--deploytool_submariner_args="--multi-active-gateway=true"'

kubectl config use-context cluster1
kubectl label nodes cluster1-worker2 submariner.io/gateway=true

kubectl config use-context cluster2
kubectl label nodes cluster1-worker2 submariner.io/gateway=true
```

Manually deploying:

```
make clusters

kubectl config use-context cluster1
kubectl label nodes cluster1-worker submariner.io/gateway=true
kubectl label nodes cluster1-worker2 submariner.io/gateway=true

subctl deploy-broker --kubeconfig output/kubeconfigs/kind-config-cluster1 --repository localhost:5000 --version local --globalnet

subctl join --kubeconfig output/kubeconfigs/kind-config-cluster1 broker-info.subm --clusterid cluster1 --natt=false --repository localhost:5000 --version local --cable-driver vxlan --multi-active-gateway=true

kubectl config use-context cluster2
kubectl label nodes cluster1-worker submariner.io/gateway=true
kubectl label nodes cluster1-worker2 submariner.io/gateway=true

subctl join --kubeconfig output/kubeconfigs/kind-config-cluster2 broker-info.subm --clusterid cluster2 --natt=false --repository localhost:5000 --version local --cable-driver vxlan --multi-active-gateway=true
 ```

This is a PoC and there were some shortcuts taken and there are some limitations.
* All nodes must be in Multiple Active Gateway mode.
  This won’t be required in the final implementation.
* Headless services are broken.
  Every Globalnet Controller was not rewritten until architecture was agreed upon.
* Only the VxLAN Cabledriver is supported. 
* Some of the `subctl` commands like `subctl diagnose` were not modified to take into account MAG.
  Unit Tests were not written for the new feature, some existing tests were modified to fix compilation
  issues or get existing tests to pass.
* The new `multi-active-gateway` flag was added to both implementations of `subctl` (old and new).
  In the time since the PoC was completed, `subctl` has been moved to its own repo.
  So this PoC code will need to be ported to the new repo.
  **Caution when rebasing to the feature-multi-active-gw branch to devel.**

File: submariner/multiple-active-gateways.md [hunk 36]
```diff
@@ -0,0 +1,663 @@
...
+
+## Notes and Assumptions
+
+* The initial POC will use standard external IP addresses for gateways, but load balancer mode is
+  a high priority for the final implementation.
+* The initial design and POC is focused on CNIs that use the kubeproxy route agent handler that establishes VXLAN
+  tunnels between worker nodes and gateways. Other options need to be investigated and supported.
+* The multipath routes require an interface, so this approach won't work with IPsec Tunnel Mode.  IPsec Transport mode,
+  proposed in [IPSec transport mode support](https://github.com/submariner-io/enhancements/pull/103) will be required for
+  IPsec encryption.
```
Billy99, line 662:
```suggestion
  IPsec encryption.
* During the review of the Enhancement Proposal, there was concern raised regarding the SNAT that occurs in Step 3-b (See
   #multiple-active-gateway-packet-flow-without-globalnet and #multiple-active-gateway-packet-flow-with-globalnet).
   The concern is that up to this point, Submariner has preserved the source IP all the way to the destination pod.
   This may be needed in the future for policy enforcement. The SNAT is in place so the packet returns to the Gateway
   the packet came from within this cluster so that connection tracking can reverse the DNAT that occurred when the
   Service IP was resolved to a Pod IP. At this time, we do not have a work around. 
```

Billy99 (commented):
Added updates based on the Tech Exchange. Also added some PoC details.

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,680 @@
...
+* During the review of the Enhancement Proposal, there was concern raised regarding the SNAT that occurs in Step 3-b (See
+   #multiple-active-gateway-packet-flow-without-globalnet and #multiple-active-gateway-packet-flow-with-globalnet).
```
Billy99: My bad:
```suggestion
   [Multiple Active Gateway Packet Flow without Globalnet](#multiple-active-gateway-packet-flow-without-globalnet) and
   [Multiple Active Gateway Packet Flow with Globalnet](#multiple-active-gateway-packet-flow-with-globalnet)).
```

submariner/multiple-active-gateways.md
```diff
@@ -0,0 +1,680 @@
...
+   the packet came from within this cluster so that connection tracking can reverse the DNAT that occurred when the
+   Service IP was resolved to a Pod IP. At this time, we do not have a work around. 
+
+## TODO
+
+* Performance testing on POC code.
+* Support non-kubeproxy route agent handlers.
+* Support Loadbalancer mode.
+* Support detection and handling of gateway failure and recovery.
+* In the PoC, only the new ClusterGlobalEgressIP Controller was implemented in the Gateway Pod (See #multiple-active-gateway-changes). The other controllers need to be ported from Globalnet.
```
Billy99:
```suggestion
* In the PoC, only the new ClusterGlobalEgressIP Controller was implemented in the Gateway Pod (See [Multiple Active Gateway Changes](#multiple-active-gateway-changes)). The other controllers need to be ported from Globalnet.
```

Billy99 (approved):

astoycos (approved):

sridhargaddam (approved):
I see that the open-items/limitations discussed during the tech-exchange session are added as  TODO/Notes in the PR. LGTM, thank you.

submariner/multiple-active-gateways.md
mkolesnik: I see this listed under TODO as "* Support non-kubeproxy route agent handlers."

mkolesnik (approved):
I see concerns are listed as TODO, so we can merge as "Deferred", with the intention that should this be picked up in the future the entire EP will have to be re-reviewed.

anfredette:
@sridhargaddam, I think this is good to merge now, isn't it?  Thanks, Andre

sridhargaddam:
> @sridhargaddam, I think this is good to merge now, isn't it? Thanks, Andre

Indeed, thank you @anfredette

anfredette:
Thanks, @sridhargaddam!

PR #72, opened 2022-03-08
Proposal: finer grained connectivity and multiple cable driver configuration

maryamtahhan:
An enhancement proposal that outlines finer grained connectivity and multiple cable driver configuration based on https://github.com/submariner-io/enhancements/pull/4 


submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,415 @@
...
+- The same cable driver implementation with differing configurations.
+
+For other use cases, it makes sense to offer finer-grained connectivity options. One such use case is a client-server
+scenario where multiple client clusters need to access a server cluster but the user does not want the client clusters
+to connect to one another. Finer-grained connectivity could also improve overall scalability where full mesh is not
+required.
+
+## Proposal
+
+This enhancement proposes a mechanism to allow users to specify their intent as to which clusters to connect and how to
```
yboaron: This proposal allows defining a list of 'neighbors' clusters for each cluster and how to connect (vxlan,ipsec, etc)  to each 'neighbor'.

It isn't clear to me how the entire clusters connectivity is derived from this information. 

for example, in the diagram you attached, do you expect cluster#2 will be able to connect  cluster#6 though they aren't 'neighbors'?

maryamtahhan: > It isn't clear to me how the entire clusters connectivity is derived from this information.

More details of how this is accomplished is given on L221 if this is not clear I can certainly clarify further.

based on the diagrams and the policy definitions for the example diagram (specified further below) cluster#2 and cluster#6 shouldn't connect.

the idea behind the proposal is to allow for the specification of connectivity policies between clusters in a clusterset based on label selectors (so clusters with label x can connect to clusters in label y but are not allowed to connect to clusters with label z). These policies also define the topology for the cluster connection. The policies can also be prioritized in order to establish how the clusters should be interconnected.

yboaron: Got it, thanks for the clarification @maryamtahhan .

I can see the added value in allowing flexibility in SM topology I think the current proposal allows high flexibility in topology configuration, which I am not sure all of that is required in the context of Submariner.

Maybe it would be useful to add a section with the requirements we would like to cover in the context of SM topology config flexibility, what do you think?

This section may help to clarify things like:
- Do we want to support different cable types between connected clusters (like in your example with IpSec and VxLAN)? 
- Do we want to support multiple cable drivers of the same type (e.g: unencrypted) on topology?  
- Do we want to support directional granularity?  IIUC, the current proposal allows to define for example A-->B will connect via VxLan and reverse direction (B --> A) will use IpSec
- How does partially connected topology comply with MCS service export API?
- etc

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,415 @@
...
+
+## Summary
+
+Submariner currently assumes full mesh connectivity by default and provisions VPN connections between all participating
+clusters. In addition, Submariner currently offers one of three cable drivers: VXLAN, IPSec or wireguard with limited
+configuration options. Only one of these cable drivers is enabled across all the clusters in a cluster set, that is all
+clusters that join a broker must use the same cable driver. The goal of this proposal is to outline an extended model
+that will support:
+
+- The selection of which clusters to connect and in what fashion (full mesh or connection to subset of clusters in
```
mkolesnik: Why is this necessary?

A clusterset is, by our definition, "one reachable space". If we're starting to segment that, then why even use submariner?

I'm talking about L3 connectivity, not ACLs which could block or allow certain types of communication.

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,415 @@
...
+cluster set. For example, a cluster set could contain three clusters, two that are on-premise and the other
+in the cloud. Connections to the cloud cluster will need to use encryption while the on-premise clusters can use
+unencrypted connections, as they’re connected by a private link, to avoid the performance overhead of encryption.
+This could be achieved using either:
+
+- Different cable driver implementations or
+- The same cable driver implementation with differing configurations.
+
+For other use cases, it makes sense to offer finer-grained connectivity options. One such use case is a client-server
+scenario where multiple client clusters need to access a server cluster but the user does not want the client clusters
```
mkolesnik: Doesn't sound like Submariner is a good fit for this use case.

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,415 @@
+# Support fine-grained cluster connectivity and multiple cable drivers
+
+## Summary
+
+Submariner currently assumes full mesh connectivity by default and provisions VPN connections between all participating
```
mkolesnik: I suggest for sake of simplicity to stick with full mesh for this discussion as other topologies have been already discussed before and pose little benefit but come at great costs.

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,415 @@
...
+this enhancement.
+
+This enhancement will propose a single API for finer-grained connectivity and multiple cable driver support.
+
+### Finer-grained connectivity and multiple cable driver API
+
+This following CRD is proposed:
+
+```Go
+type ClusterConnectionPolicies struct {
```
mkolesnik, line 60:
```suggestion
type ClusterConnectionPolicy struct {
```

Shouldn't it be in the singular?

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,415 @@
...
+    metav1.TypeMeta `json:",inline"`
+
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+
+    Spec ClusterConnectionPolicy `json:"spec"`
```
mkolesnik:
```suggestion
    Spec ClusterConnectionPolicySpec `json:"spec"`
```

?

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,415 @@
...
+
+    // RightClusterSelector identifies the Cluster resources representing the clusters on the other end of a connection.
+    // An empty selector indicates wildcard, ie matches any cluster.
+    // Not used in combination with ClusterSelector.
+    RightClusterSelector metav1.LabelSelector `json:"rightClusterSelector,omitempty"`
+
+    // ClusterSelector identifies the Cluster resources representing a subset of clusters.
+    // Not used in combination with LeftClusterSelector or RightClusterSelector.
+    // An empty selector indicates wildcard, ie matches any cluster.
+    ClusterSelector metav1.LabelSelector `json:"clusterSelector,omitempty"`
```
mkolesnik: Why is there this option?

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,415 @@
...
+
+    // Topology identifies the connection topology for the cable ends: full-mesh/point-to-point/client-server.
+    Topology `json:"topology,omitempty"`
+
+   // Priority identifies the priority for the connection policy. Higher priorities are applied first.
+    Priority  *int  `json:"priority,omitempty"`
+
+    // ConnectionPolicy identifies the default connection policy if no policies are found: to connect or no-connect.
```
mkolesnik, line 82: This doesn't comply with the definition of "cluster set"

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,415 @@
...
+    ConnectionPolicy *string `json:"connectionPolicy,omitempty"`
+
+    Spec CableDriverSpec `json:"spec"`
```
mkolesnik: It'll be just named "spec"? A bit lacking of information IMO

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,415 @@
...
+
+If any matching policy is found, the clusters are either:
+
+- connected (based on the connectionPolicy) according to the specified topology in the policy. Or
+- not connected (based on the connectionPolicy).
+
+If more than one matching policy is found, the priority field is used to determine which policy to use. If policies of equal priority are found
+it's assumed that these policies are to be applied in parallel.
+
+If no matching policy is found, the default cable driver and default connection policy are used. This policy can be auto-created at broker
```
mkolesnik: The default cable driver should be the one defined on the `Submariner` CR

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,415 @@
...
+a topology and default cable are not needed.
+
+> **_NOTE:_**  the `ConnectionPolicy` field can also be used to specify policies for which clusters not to connect.
+> **_NOTE:_** In the case of multiple labels being specified, the selector will need to match all the labels. Equality-based requirements will
+be supported. For e.g. labels such as `env: !production` will match any clusters that aren't labelled with `env: production`.
+
+Defining separate resources for each policy and defining a priority or precedence field, allows Submariner to prioritize the
+connectivity and cable type between a pair of clusters.
+
+The `ClusterConnectionPolicies` resources would be created and maintained on the broker cluster and synced to each participating cluster.
```
mkolesnik: Policies should be created locally as any other custom resource, and utilize the broker mechanism to federate them

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,415 @@
...
+A controller to reconcile updates to policies will need to be implemented in Submariner.
+
+### Labelling clusters
+
+Clusters are defined as Custom Resources and can be labeled in the CR definition, or after creation using kubectl. A new parameter to label
```
mkolesnik: Currently clusters are created by the gateway binary AFAIK, are you proposing to change that?

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,415 @@
...
+
+| Flag                       | Description
+|:---------------------------|:-------------------------------------------------------------------------------|
+| `--kubeconfig` `<string>`  | Absolute path(s) to the kubeconfig file(s) (default `$HOME/.kube/config`)
+| `--kubecontext` `<string>` | Kubeconfig context to use
+| `--name` `<string>`        | The name of the policy to delete
+
+##### `export`
+
+`subctl connection-policy export [flags]`
```
mkolesnik: What does it mean to "export" it?

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,415 @@
...
+    location: on-premise
+    env: production
+  rightClusterSelector:
+    location: cloud
+    env: production
+  connectionPolicy: no-connect
+  priority: 3
+```
+
+- We only want certain clusters to access a database server in another cluster and no other clusters to connect to one another. We
```
mkolesnik: This use case matches "Multi Cluster Network Policies" and should be covered by that effort, not by limiting connectivity as a whole

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,415 @@
...
+namespace).
+
+The proposed approach for additional submariner cable driver configuration parameters is:
+
+- To continue to use the existing environment variables injected into the Gateway Pod (with the exception of PSKs/secrets). These environment
+variables will come from the `submariner.io_submariners.yaml`.
+- To extend `submariner.io_submariners.yaml` with some new global parameters for cable driver configuration in a cluster set.
+- To use a volume mounted ConfigMap for more detailed cable driver configurations (as these may vary).
+- To use mounted secrets for parameters that need to be protected, e.g. the Pre-Shared Key (PSK) and to abandon the use of environment variables
+for secrets.
```
mkolesnik: This seems like a plethora of ideas that are pre-cursors or perhaps tangential to this EP, I suggest to tackle them separately instead of lumping them in here.

This EP should focus on the policy aspects not on extending or refactoring cable driver customization options.

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,415 @@
...
+
+#### Multiple cable driver configuration
+
+The proposed global configuration parameter extensions to `submariner.io_submariners.yaml` to configure multiple cable drivers are listed
+in the following table:
+
+<!-- markdownlint-disable line-length -->
+| Parameter name              | Description
+|:----------------------------|:----------------------------------------------------------------------------|
+| cableDrivers                | An array of strings listing the drivers to configure as separate cables.
```
mkolesnik, line 385: Global config should remain a single driver, policies are an optional addition on top.

mkolesnik (changes requested):

submariner/multiple-cable-drivers.md
maryamtahhan: If we are just using full mesh then the rest of this proposal is null and void, no? as we are already operating in the desired mode

submariner/multiple-cable-drivers.md
maryamtahhan: So what you are saying is Submariner will ONLY EVER cater for full mesh connectivity between all clusters participating in a cluster set. You either look to connect everything and take the hit of redundant paths and their costs or find some other soln...

submariner/multiple-cable-drivers.md
maryamtahhan: The intention was that this holds a collection of ClusterConnectionPolicySpec - but I will update the name to a singular and make sure that the ClusterConnectionPolicySpec is a slice of ClusterConnectionPolicySpecs...

submariner/multiple-cable-drivers.md
maryamtahhan: Yes - I will update

submariner/multiple-cable-drivers.md
maryamtahhan: Just an option, if a bunch of clusters share a label, e.g. in the attached example it's usage is shown on L262... to connect all the clusters labelled cloud in a full mesh fashion...

submariner/multiple-cable-drivers.md
maryamtahhan: Perhaps my understanding of the MCS definition of a ClusterSet is in conflict with the Submariner definition. Where is the Submariner definition described?

submariner/multiple-cable-drivers.md
maryamtahhan: @yboaron you bring up good questions and points. I will attempt to address them all over the coming days.

submariner/multiple-cable-drivers.md
maryamtahhan: Yes - I can rename

submariner/multiple-cable-drivers.md
maryamtahhan: Yes

submariner/multiple-cable-drivers.md
maryamtahhan: That originally was the intention - I will update the description to match the intention, the other thing we will need to look out for is actually making sure to create the policies before cable driver startup to avoid having to restart cables.

But I need to think about this a little more however - managing the initial definitions through the operator could actually be better to avoid cable restarts...

submariner/multiple-cable-drivers.md
maryamtahhan: no not planning on changing that... Just saying that the labels can be applied at creation time (on the GW) or later through Kubectl... but I guess the latter would pose an issue from a cable startup POV - as any changes to the cluster labels later on may require a cable restart based on the defined cluster connectivity policy...

submariner/multiple-cable-drivers.md
maryamtahhan: The intention was to create the connectivity policy locally on the cluster and then export it in a similar way we export services (use a federator to propagate it to the Broker)...

submariner/multiple-cable-drivers.md
maryamtahhan: I guess in one sense it is but the GWs will have no awareness of what they can/can't connect to as with the multicluster network policy everything is handled at an external level - so there could be wasted resources on the GW side... I would need to look at this more closely

submariner/multiple-cable-drivers.md
maryamtahhan: Fair enough

submariner/multiple-cable-drivers.md
maryamtahhan: The point is that you can connect with different cables to different clusters. Not everything is only vxlan OR libreswan tunnels...

submariner/multiple-cable-drivers.md
yboaron: Thanks

submariner/multiple-cable-drivers.md
mkolesnik: You can see it here https://submariner.io/getting-started/architecture/

submariner/multiple-cable-drivers.md
mkolesnik: First, secure tunnels are point to point by design, there should be no middlemen eavesdropping on the contents, so I'm not sure how you expect to provide a secure tunnel with a hub cluster.

Second, I think where clusters on "the edge" don't need to talk to each other, call them clients or whatever, then this isn't what Submariner is trying to solve.

submariner/multiple-cable-drivers.md
mkolesnik: There is value in allowing different types of tunneling (or encryption) to happen within parts of the clusterset, I don't see how that's null and void.

submariner/multiple-cable-drivers.md
mkolesnik: That's not my point, rather that the default config should be global and you could optionally specify policies on top to have a different connectivity type for specific areas in the whole clusterset

submariner/multiple-cable-drivers.md
maryamtahhan: That's the same definition as the MCS one. I guess I'm not seeing the part that's explicit re-cluster connectivity --> or maybe I'm misunderstanding the meaning of "the ClusterSet is one reachable space" does that mean that every cluster should be connected to every other cluster?

Or perhaps my wording is misleading... The intention is that clusters are aware of all other clusters in a cluster set. They may not necessarily be connected to them directly as in the attached example.

Are you saying - for the scenario in the example: 1,2,3 and 4 should be a ClusterSet while 5 and 6 should be a different ClusterSet?

submariner/multiple-cable-drivers.md
maryamtahhan: > First, secure tunnels are point to point by design, there should be no middlemen eavesdropping on the contents, so I'm not sure how you expect to provide a secure tunnel with a hub cluster.

I think we are talking past each other here... and I'm a not sure I understand the relevance of the comment TBH

> Second, I think where clusters on "the edge" don't need to talk to each other, call them clients or whatever, then this isn't what Submariner is trying to solve.

What is/are the use cases Submariner is trying to resolve? 

I don't think this is applicable to edge only... Suppose there's a service exposed in a cluster we'll call cluster-eu and that service is consumed by applications that reside in clusters in the US we will call cluster-us-1 and cluster-us-2... now the user knows that cluster-us-1 and cluster-us-2 only ever consume that service from cluster-eu (a really simple example an enterprise CI service exposed to multiple dev teams with dev environments in clusters in different geos)... why would they want to connect cluster-us-1 and cluster-us-2? There's an explicit cost of maintaining a connection they don't need (BW, CPU cycles all to process cable health checks...).

Also another use-case beyond server-client topology would be in the situation you need to take down a cluster (that's already part of a ClusterSet) for maintenance or upgrade or some-other operation, how do you specify to the other clusters in the ClusterSet to temporarily not connect to that cluster? Or is the expectation to uninstall/unjoin Submariner on the cluster then re-join when ready?

submariner/multiple-cable-drivers.md
maryamtahhan: So should this be 2 proposals?

submariner/multiple-cable-drivers.md
maryamtahhan: These params are extensions to the existing global configuration for 1 or more cable driver(s)... 

So the single cable driver configuration that exists in `submariner.io_submariners.yaml`  would still exist and that cable driver would be assumed to be the default as specified on L401. Beyond the policies - these params are configuration for the cable drivers in a cluster

submariner/multiple-cable-drivers.md
tpantelis: > > First, secure tunnels are point to point by design, there should be no middlemen eavesdropping on the contents, so I'm not sure how you expect to provide a secure tunnel with a hub cluster.
> 
> I think we are talking past each other here... and I'm a not sure I understand the relevance of the comment TBH
> 

yeah I think there's a disconnect here. Back 3 years ago when we were first starting with this project, we had discussions/proposal re: gateway topologies, eg hub-n-spoke where all gateway nodes connect to a hub gateway node on a hub cluster which would route traffic accordingly. I **think** this is what @mkolesnik is referring to and interpreting from your proposal. But, from my interpretation, this is not what you're proposing, ie you're merely proposing policies which determine whether or not submariner should establish a connection to a remote cluster `Endpoint`. Unless I'm missing something...

submariner/multiple-cable-drivers.md
tpantelis: I was also a little confused by this option. Can't that be accomplished by:

```
leftClusterSelector:
    location: cloud
rightClusterSelector:
    location: cloud
```

submariner/multiple-cable-drivers.md
tpantelis: Connectivity policies need to be automatically exported/synced to the broker, right? Each cluster needs to observe the same connectivity policies.

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,415 @@
...
+    // Not used in combination with ClusterSelector.
+    RightClusterSelector metav1.LabelSelector `json:"rightClusterSelector,omitempty"`
+
+    // ClusterSelector identifies the Cluster resources representing a subset of clusters.
+    // Not used in combination with LeftClusterSelector or RightClusterSelector.
+    // An empty selector indicates wildcard, ie matches any cluster.
+    ClusterSelector metav1.LabelSelector `json:"clusterSelector,omitempty"`
+
+    // Topology identifies the connection topology for the cable ends: full-mesh/point-to-point/client-server.
+    Topology `json:"topology,omitempty"`
```
tpantelis: What is the significance of this, ie how would submariner use it? The current tunnel controller watches for remote `Endpoint` resources and establishes a connection via the configured cable driver. It seems the intent of this proposal is to extend the controller such that it consults the connectivity policies to determine if the connection should be established. If so, I'm not clear on how seemingly higher-level concepts such as "point-to-point" and "client-server" are relevant.

submariner/multiple-cable-drivers.md
maryamtahhan: Yes - that is correct. I guess we can get rid of the export option and just assume that policies will always be automatically exported on creation

submariner/multiple-cable-drivers.md
maryamtahhan: it was to represent the intention of the logical topology for a connection - my thinking was that we could use it to reduce the number of policies specified. It's not strictly needed as the intention can be reflected via individual policies. 

For example, if someone specified a server-client topology with the left and right selectors, we could assume that the right selector criteria represents all clients and they don't need to be connected to one another. Rather than have a second policy that states not to connect label x clusters together.

But like I said - not strictly needed and can be removed.

submariner/multiple-cable-drivers.md
maryamtahhan: It can. which is what I used in the example on L242. The intention was to allow someone a shorthand option at expressing the same thing. It can be removed

submariner/multiple-cable-drivers.md
maryamtahhan: Yeah, @tpantelis that is the intention of the proposal - simple policies that say connect cluster a to cluster b, or don't connect cluster c to cluster a...

The expression of topology I was using relates to the logical topology for the connection between clusters - which may no longer be full-mesh (when using connectivity policies) as a cluster may or may not establish a connection to another cluster. I was just trying to add language for a representation for the resulting connection graph.

Anyway - I can see the reference to topologies is causing a little heartburn - so I'm happy to abandon the reference to it.

submariner/multiple-cable-drivers.md
tpantelis: OK - I see. That would come into play depending on what the user wants wrt everything else not in that server-client sub-topology. That would depend on what the default/fallback policy is. If `no-connect` then nothing further is needed with or without `topology`. If `connect` then:

```
spec:
  leftClusterSelector:
    database-role: client
  rightClusterSelector:
    database-role: client
  connectionPolicy: no-connect
```

would be needed in lieu of the `topology` field.

maryamtahhan: yeah that's was the intention

tpantelis: The `topology` field or something equivalent might still make sense. It depends on how flexible we want to make it. In the [original proposal](https://github.com/submariner-io/enhancements/pull/4), there wasn't a `no-connect` policy, ie the out-of-box mode was full mesh which switched off when at least one policy was defined. So the fallback was implicitly `no-connect`, assuming that users with larger environments would want to restrict cluster connectivity to a smaller, manageable set with the "opt-in" approach (for client-server and scalability).  Your proposal expands that to include the "opt-out" approach to do the opposite. We didn't have a use case for that and didn't consider it. Do you have a real-world use case for it?

tpantelis: Let's consider the client-server use case. I have a `server` cluster along with `client1` and `client2` and another `cluster3`.  So I add a policy to connect the clients with the server but disallow `client1` <-> `client2` connectivity.  With your  proposal the default/fallback policy is `connect`. So the cluster connectivity would be:

  `client1` <-> `server`
  `client2` <-> `server`
  `cluster3` <-> `server`
  `cluster3` <-> `client1`
  `cluster3` <-> `client2`

The policy only disallows `client1` <-> `client2`. The 2 clients can connect to `cluster3`. Also `cluster3` is able to connect to the `server`.  This doesn't make sense to me - it's not really what I want. It **does** make sense if the default policy is `no-connect`, then you get the desired connectivity of:

  `client1` <-> `server`
  `client2` <-> `server`

This is why the original proposal took the strictly opt-in approach with policies. Otherwise with the `connect` fallback policy, you may not be really restricting much of anything or not what you intended.

submariner/multiple-cable-drivers.md
tpantelis: > 
> Are you saying - for the scenario in the example: 1,2,3 and 4 should be a ClusterSet while 5 and 6 should be a different ClusterSet?

But that wouldn't allow 2 and 5 to connect via IPSec as in your diagram, unless a cluster is able to be a member of multiple cluster sets which isn't allowed by the MCS spec. This also wouldn't achieve the client-server use case where you don't need/want the client clusters to access each other, but they all still need to be part of the same cluster set. Like you said, a cluster is added to a set b/c it needs to access at least one cluster in the set but it doesn't necessarily need to access **every** cluster in the set.

submariner/multiple-cable-drivers.md
tpantelis: Again @mkolesnik might be referring to the gateway topologies which this proposal isn't referring to.

However this proposal does combine 2 concepts which may or may not be disparate, I think depending on one's POV or use case: 1) should two clusters establish a connection and 2) how should they be connected (ie which cable driver and config to use). The absolute original proposal also combined them using the same CRD, as you've done here. But after much discussion, it was decided that the concepts and known associated use cases were different and thus it made sense to defined separate CRDs, hence the [second proposal](https://github.com/submariner-io/enhancements/pull/4). I think it might make sense to continue down that path. It would allow us to implement one w/o the other. I think you're really interested in 2) at this point anyway. The semantics of the cable driver connection config policy seem simpler than that of the connectivity policy and I think more thought and discussion and perhaps more real-world uses cases are needed for the latter.

submariner/multiple-cable-drivers.md
maryamtahhan: @tpantelis the intention in the proposal was that the default policy is configurable. It can be connect or it can be no connect. Where none is specified the default policy will be to connect (which is keeping in line with what Submariner has done historically)...
There is some onus on the user to use with care. But this comes through examples and guides to help them reach a desirable configuration

submariner/multiple-cable-drivers.md
```diff
@@ -0,0 +1,427 @@
...
+    Name string `json:"name"`
+
+    // Options specifies key/value configuration parameters for the cable driver.
+    CableOptions *v1.ConfigMap `json:"options,omitempty"`
+}
+```
+
+The `ClusterConnectionPolicies` CR defines the criteria for selecting which cluster pairs/subsets can connect and
+associated optional configuration options to use for each cluster connection. To resolve the connectivity policy
+for a remote cluster's Endpoint, Submariner would:
```
maryamtahhan, line 99: I've been thinking about the order of processing here and the complexity added by having a priority... 

From a practical POV of what would be preferable from an implementation - is to remove the priority altogether. There's an onus on the user to put together a set of policies that describe the desirable end state for the cluster connections.

If priorities are gone then it's also important to have a default policy which will be called `default` in order to distinguish it from everything else.

So in order to process these policies (looking at this purely from a connectivity POV and excluding the cable driver for the minute), the GW could build a connection map[string]string , where the key is a cluster identifier or potentially even an endpoint identifier and the value is connect or no-connect. On the first pass the default policy is applied to all values in the map. Establishing a connectivity list that's either a permit-list or a deny-list based on the default policy. 

The next thing the GW should do is parse the rest of the policies - where it finds a policy whose left selectors match it, it establishes the list of clusters that match the right selectors and updates the connectivity map with what's specified in the policy (this info could be pushed as a ConfigMap to the local K8s API so that it's available for other GWs - or just let the GWs do the relevant configuration when they come up). 

When the time comes to actually install the cable - the connectivity map is consulted... 

or something along those lines...

submariner/multiple-cable-drivers.md
maryamtahhan: No worries at all. Continuing down a path that the community agreed made sense. 

Can I close this proposal?

tpantelis: Sure however you want to do it.

maryamtahhan:
Thanks all for the reviews. 

Closing this PR as the community sees this EP as two separate work items and there's an EP already in progress that breaks the first part of this work down. As regards the second item more discussion, design and real life use cases are needed.

submariner/multiple-cable-drivers.md
tpantelis: yeah - the default policy may not make sense depending on what the user wants to do. We could make it simpler by not having a connect or not connect choice as in the original proposal. Again the thinking/assumption is that a user would want to pare down the connectivity in which case the `no-connect` default policy makes sense - this was based on the 2 use cases that we thought of, client-server and scalability. As mentioned before, your proposal extends it to facilitate a case where the user only wants to disallow connectivity between a small number of clusters in which case a `connect` default policy makes that easier. But I don't know if there's a real use case for that.

submariner/multiple-cable-drivers.md
tpantelis: For cable driver connection policies, we would need priorities in some manner to avoid ambiguity should a cluster match multiple policies. For connectivity, if we go with a simpler approach where a policy only species what to connect then it's a simple binary decision so priorities aren't needed, ie we need only search for one policy that matches.  But if we add the `connect`/`no-connect` option, then I think we do need priorities to resolve a case where a cluster matches both a `connect` and a `no-connect` policy.

PR #98, opened 2022-03-29
cable-driver: IP in IP

maryamtahhan:
Enhancement proposals for Cable Drivers: IPinIP support


submariner/cable-driver-protocols-enhancement.md
```diff
@@ -0,0 +1,155 @@
+# Cable Driver Protocol Support Enhancements
+
+## Summary
+
+Submariner today supports three cable drivers: VXLAN, IPSec (tunnel mode) and WireGuard. The
+goal of this proposal is to outline additional protocols and enhancements for the Submariner
+Cable Driver. These include the additional support of IP-in-IP for unencrypted connections
+between clusters and the support of IPSec transport mode.
```
sridhargaddam: Please add some notes about why we would need the new cable drivers and how it is going to help.

submariner/cable-driver-protocols-enhancement.md
```diff
@@ -0,0 +1,155 @@
...
+
+> **_NOTE:_** A single IP Tunnelling interface will be added, and tunnels managed using
+`# ip route add encap ip <args>` in order to avoid a point to point connection per tunnel.
+
+#### Design Details
+
+* A new interface "ipip-tunnel" will be added.
+* The IP tunnelling interface will be one-to-many, that is a single interface will be used to
+connect to all the joined clusters.
+* For the "ipip-tunnel" IP, the prefix "242" will be used, followed by the rest of the blocks
```
sridhargaddam: Lets avoid 242 prefix as its the default [Globalnet CIDR](https://github.com/submariner-io/submariner-operator/blob/16df4bcd7bc55cb87d6d80adbb7d5a726749a715/pkg/broker/globalcidr_cm.go#L39) and the same ip could be allocated by Globalnet controller as Egress/IngressIP.

VxLAN currently uses 241 prefix, so we can use 243.

submariner/cable-driver-protocols-enhancement.md
```diff
@@ -0,0 +1,155 @@
...
+from the endpoint private IP address. For example, if the private IP address is 10.2.96.1 the
+"ipip-tunnel" IP would be 242.2.96.1.
+* The routes will be added to forward all the Service and Pod CIDR traffic to the respective
+remote Tunnel IPs.
+* Table 100 shall be used to add the routes, which is otherwise used for host network traffic
+routes. We do not require any extra rules to handle host network traffic for the IPTun driver.
```
sridhargaddam:
```suggestion
* Table 100 shall be used to add the routes for inter-cluster communication.
```

Table 100 is actually used by [VxLAN cable driver](https://github.com/submariner-io/submariner/blob/8c4145c64e1391048ffad9e7553ad47be0e26ff9/pkg/cable/vxlan/vxlan.go#L48) to program inter-cluster routes. We use Table 150 to support [hostNetworking use-cases](https://github.com/submariner-io/submariner/blob/8c4145c64e1391048ffad9e7553ad47be0e26ff9/pkg/routeagent_driver/constants/constants.go#L51) from the route-agent driver.

I think we can continue to use Table 100 even if its used by VxLAN cable driver.

sridhargaddam (changes requested):
Thanks for pushing the enhancement PR @maryamtahhan. 

Added some initial comments, will take a close look at it later.

submariner/cable-driver-protocols-enhancement.md
maryamtahhan: I will update it

submariner/cable-driver-protocols-enhancement.md
```diff
@@ -0,0 +1,179 @@
...
+## Summary
+
+Submariner today supports three cable drivers: VXLAN, IPSec (tunnel mode) and WireGuard. The
+goal of this proposal is to outline additional protocols and enhancements for the Submariner
+Cable Driver. These include the additional support of IP-in-IP for unencrypted connections
+between clusters and the support of IPSec transport mode.
+
+For IP in IP, the benefits fall into the following categories:
+
+* **Performance**: IP in IP is has a lower overhead than VXLAN which could lead to performance
```
nyechiel, line 12: Do we have any numbers of the performance impact?

maryamtahhan: I don't have an appropriate setup to run such a benchmark... But here's an academic paper that has: https://repository.dl.itc.u-tokyo.ac.jp/record/51086/file_preview/A34115.pdf 

and some screen shots from it... 

Screenshot 2022-03-31 at 09 02 27
Screenshot 2022-03-31 at 09 02 00
Screenshot 2022-03-31 at 09 02 17

dfarrell07: Those numbers are compelling to me @maryamtahhan, thanks! Do you think we should add them to the EP, for the record?

dfarrell07:
If we implement IP-in-IP, noting the significant perf improvements, is there any reason to keep VXLAN?

maryamtahhan:
> If we implement IP-in-IP, noting the significant perf improvements, is there any reason to keep VXLAN?

Yes I think VXLAN should be kept, both are typical supported encapsulations and some customers may prefer one or the other.

VXLAN is also still the inter-node in a cluster encapsulation used in Submariner and it would be the more appropriate protocol to encap all sorts of traffic between the nodes in the same cluster (as it takes the whole ethernet frame in it's payload)

Also - I *think* IP-in-IP supports the encapsulation of unicast traffic only... I haven't played around with it with more complex traffic profiles.

submariner/cable-driver-protocols-enhancement.md
maryamtahhan: I will add them in

submariner/cable-driver-protocols-enhancement.md
```diff
@@ -0,0 +1,179 @@
+# Cable Driver Protocol Support Enhancements
+
+## Summary
+
+Submariner today supports three cable drivers: VXLAN, IPSec (tunnel mode) and WireGuard. The
+goal of this proposal is to outline additional protocols and enhancements for the Submariner
+Cable Driver. These include the additional support of IP-in-IP for unencrypted connections
+between clusters and the support of IPSec transport mode.
+
+For IP in IP, the benefits fall into the following categories:
```
yboaron, line 10: Will it be possible to run IPinIP tunnel behind NAT?

yboaron, line 10: I'm not familiar with cloud-prepare bits, I assume we need to run some specific firewall config to allow IPinIP traffic, right?

submariner/cable-driver-protocols-enhancement.md
```diff
@@ -0,0 +1,179 @@
...
+
+* **Performance**: IP in IP is has a lower overhead than VXLAN which could lead to performance
+improvements in terms of packet overhead (bandwidth consumption) as well as the cost of processing
+packets.
+* **Interoperability**:IP in IP can be used to transport packets between domains when the protocol
+in those domains (e.g. IPv6) is not supported by intermediary networks (e.g. IPv4).
+
+For Overlay-over-IPSec in transport mode with further configuration, the benefits fall into the
```
yboaron: Not familiar with SM IPSec implementation, 
could you please elaborate more about this option? and how it differs from the current IPSec implementation?

yboaron: OK, I tcpdumpped SM ipsec traffic.

So, do you mean that for VxLAN over ipsec for example, the packet format will be something like: 

[L2]-[external-IP][UDP][VxLan Header] [org_L2][org_ip header][ESP of org IP payload]   ?   or I got confused with the layers :-) 

While current SM ipsec packet format (tunnel mode)  is : 
[L2]-[external-IP] [ESP of org IP payload including org_ip header]

Could you please elaborate on the performance improvement you mentioned?  is it related to HW offload?

submariner/cable-driver-protocols-enhancement.md
maryamtahhan: So it depends on the cloud env - for GCP and for AWS it looks like some simple configuration... For Azure IP in IP traffic is blocked... 

It looks like it's also supported in OpenShift but in concert with the edge loadbalancer from F5... But I'm guessing from a configuration POV it shouldn't pose a problem

The goal here though is to enable a low overhead encapsulation that can be used together with IPSec in transport mode - so the IPinIP traffic would be encaped with the ESP header[Outer IP | ESP Header | Payload (inner IP | UDP or TCP | inner payload )] and this should not pose any additional issues for various environments as it's transparent till arrival at the GW...

submariner/cable-driver-protocols-enhancement.md
maryamtahhan: Yeah - the performance improvements are mostly related to HW offload... 

reducing the packet overhead and thus the amount of cycles needed to process the packet also helps from a CPU and bw consumption POV... 

There's a nice illustration of Tunnel mode vs transport mode packets here https://yurmagccie.files.wordpress.com/2019/01/ipsec-tunnel-transport.jpg

maryamtahhan:
Any objections to breaking this up into 2 EPs? one for IP in IP and another for IPSec transport mode?

skitt:
> Any objections to breaking this up into 2 EPs? one for IP in IP and another for IPSec transport mode?

None from me!

submariner/cable-driver-protocols-enhancement.md
```diff
@@ -0,0 +1,189 @@
...
+  * ![Throughput rate Gbps](./images/tx-tput-gbps.png)
+  * ![CPU cycles](./images/cpu-cycles.png)
+
+* **Interoperability**:IP in IP can be used to transport packets between domains when the protocol
+in those domains (e.g. IPv6) is not supported by intermediary networks (e.g. IPv4).
+
+For Overlay-over-IPSec in transport mode with further configuration, the benefits fall into the
+following categories:
+
+* **Performance**: Taking advantage of and overlay protocol wrapped in offloaded IPSec in transport mode can
```
dfarrell07:
```suggestion
* **Performance**: Taking advantage of an overlay protocol wrapped in offloaded IPSec in transport mode can
```

submariner/cable-driver-protocols-enhancement.md
```diff
@@ -0,0 +1,189 @@
...
+
+* **Performance**: IP in IP is has a lower overhead than VXLAN which could lead to performance
+improvements in terms of packet overhead (bandwidth consumption) as well as the cost of processing
+packets. Some of the performance improvements are shown below [1]:
+
+  * ![Throughput rate vs Kernel](./images/tx-tput-rate.png)
+  * ![Throughput rate Gbps](./images/tx-tput-gbps.png)
+  * ![CPU cycles](./images/cpu-cycles.png)
+
+* **Interoperability**:IP in IP can be used to transport packets between domains when the protocol
```
dfarrell07:
```suggestion
* **Interoperability**: IP in IP can be used to transport packets between domains when the protocol
```

submariner/cable-driver-protocols-enhancement.md
```diff
@@ -0,0 +1,189 @@
...
+ip route add 100.2.0.0/16 encap ip id 100 dst 172.18.0.21 via 243.18.0.21 dev ipip0 table 100 metric 100 src 10.1.96.0
+ip route add 10.3.0.0/16 encap ip id 100 dst 172.18.0.8 via 243.18.0.8 dev ipip0 table 100 metric 100 src 10.1.96.0
+ip route add 100.3.0.0/16 encap ip id 100 dst 172.18.0.8 via 243.18.0.8 dev ipip0 table 100 metric 100 src 10.1.96.0
+ip route add 10.4.0.0/16 encap ip id 100 dst 172.18.0.4 via 243.18.0.4 dev ipip0 table 100 metric 100 src 10.1.96.0
+ip route add 100.4.0.0/16 encap ip id 100 dst 172.18.0.4 via 243.18.0.4 dev ipip0 table 100 metric 100 src 10.1.96.0
+```
+
+and the resulting routes on the gateway node:
+
+```bash
```
dfarrell07:
```suggestion
```text
```

This is actually text I think. As-is, the shell prompt is being syntax highlighted as bash (command is a comment). Same applies to a few other places below.

submariner/cable-driver-protocols-enhancement.md
maryamtahhan: I've fixed this up in https://github.com/submariner-io/enhancements/pull/103

submariner/cable-driver-protocols-enhancement.md
maryamtahhan: Fixed to console in https://github.com/submariner-io/enhancements/pull/103

submariner/cable-driver-protocols-enhancement.md
maryamtahhan: From what I see - most routers are able to deal with it... 
Also ATM it's supported in GCP and AWS

submariner/cable-driver-protocols-enhancement.md
```diff
@@ -0,0 +1,76 @@
...
+* The IP tunnelling interface will be one-to-many, that is a single interface will be used to
+connect to all the joined clusters.
+* For the "ipip-tunnel" IP, the prefix "243" will be used, followed by the rest of the blocks
+from the endpoint private IP address. For example, if the private IP address is 10.2.96.1 the
+"ipip-tunnel" IP would be 243.2.96.1.
+* The routes will be added to forward all the Service and Pod CIDR traffic to the respective
+remote Tunnel IPs.
+* Table 150 shall be used to add the routes for inter-cluster communication.
+
+Example of routeing rules added by the IPTun driver on a gateway node to other nodes
```
maryamtahhan:
```suggestion
Example of routing rules added by the IPTun driver on a gateway node to other nodes
```

mkolesnik:
> > If we implement IP-in-IP, noting the significant perf improvements, is there any reason to keep VXLAN?
> 
> Yes I think VXLAN should be kept, both are typical supported encapsulations and some customers may prefer one or the other.

I don't see why both need to be actively maintained, on the broader level.
If we have one "unencrypted" type of cable driver we should use that and the other can be experimental (no or limited testing, best effort support)

> 
> VXLAN is also still the inter-node in a cluster encapsulation used in Submariner and it would be the more appropriate protocol to encap all sorts of traffic between the nodes in the same cluster (as it takes the whole ethernet frame in it's payload)

That's an implementation detail for iptables based CNIs and isn't necessarily true in all deployments, the fact that it's there has no bearing on the solution as a whole.

submariner/cable-driver-protocols-enhancement.md
yboaron: Thanks, 
I assume that IPinIP NAT requires 1:1 mapping, isn't it?

By the way, where did you get the IPinIP information for GCP, AWS? Official documentation?

maryamtahhan: Yeah - it looks like a 1:1 mapping to me.

There's some references:

- https://cloud.google.com/config-connector/docs/reference/resource-docs/compute/computefirewall?hl=en
- https://cloud.google.com/workflows/docs/reference/googleapis/compute/beta/Overview?hl=en
- https://cloud.google.com/vpc/docs/firewalls?hl=en
- https://projectcalico.docs.tigera.io/reference/public-cloud/gce

For AWS 

- https://projectcalico.docs.tigera.io/reference/public-cloud/aws there was an AWS link that's now disappeared behind a new developer forum in AWS and I can no longer find the reference through the new developer forum :(

maryamtahhan: the only other API reference I can find now is https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_ec2/Protocol.html?highlight=ipip#aws_cdk.aws_ec2.Protocol.IPIP

submariner/cable-driver-protocols-enhancement.md
mkolesnik: The numbers seems interesting but I wonder how it would behave in our product and if they will really be reproducible or not, since the study you linked used a very specific method which may or may not translate well into real world use cases (even if we don't consider HW offloading support).

> To measure the time required for these parts, we modified a Linux kernel. We added
timestamp fields to the packet buffer (struct sk buff), and inserted read time stamp
counter (RDTSC) instructions to the start and end of each part. In this experiment,
a packet generated by the traffic generator went through each part and transit times
were stored in the timestamp fields of the packet buffer. The dummy interface shown
in Figure 4.3 saved the timestamp fields and dropped the packet immediately when it
received a packet. By using this method, we measured the time required for each part
on the TX path with five tunneling protocols.

submariner/cable-driver-protocols-enhancement.md
```diff
@@ -0,0 +1,76 @@
...
+Submariner today supports three cable drivers: VXLAN, IPSec (tunnel mode) and WireGuard. The
+goal of this proposal is to outline additional enhancements needed for the Submariner Cable Driver
+to support of IP-in-IP for unencrypted connections between clusters or for use as an overlay with
+IPSec in transport mode for encrypted connections between clusters.
+
+For IP in IP, the benefits fall into the following categories:
+
+* **Performance**: IP in IP is has a lower overhead than VXLAN which could lead to performance
+improvements in terms of packet overhead (bandwidth consumption) as well as the cost of processing
+packets. Some of the performance improvements are shown below [1]:
```
mkolesnik, line 14:
```suggestion
packets. Some of the possible performance improvements are shown below [1]:
```

submariner/cable-driver-protocols-enhancement.md
```diff
@@ -0,0 +1,76 @@
...
+`# ip route add encap ip <args>` in order to avoid a point to point connection per tunnel.
+
+### Design Details
+
+* A new interface "ipip-tunnel" will be added.
+* The IP tunnelling interface will be one-to-many, that is a single interface will be used to
+connect to all the joined clusters.
+* For the "ipip-tunnel" IP, the prefix "243" will be used, followed by the rest of the blocks
+from the endpoint private IP address. For example, if the private IP address is 10.2.96.1 the
+"ipip-tunnel" IP would be 243.2.96.1.
```
mkolesnik, line 40: Is there some importance to this decision?

submariner/cable-driver-protocols-enhancement.md
```diff
@@ -0,0 +1,76 @@
...
+* The routes will be added to forward all the Service and Pod CIDR traffic to the respective
+remote Tunnel IPs.
+* Table 150 shall be used to add the routes for inter-cluster communication.
```
mkolesnik: Table 150 where exactly?

Also your examples list table 100 is it supposed to be 150? 
e.g. `ip route show table 100`

maryamtahhan: it was table 100 - which I updated based on @sridhargaddam comment to 150... I haven't updated the implementation but I can do so and update the screen shot of the example.

table 100 and table 150 are currently the routing tables used by vxlan to redirect traffic (intra-cluster and inter-cluster respectively). we are just re-using the same practice

submariner/cable-driver-protocols-enhancement.md
maryamtahhan: we adopted a similar practice to what vxlan is doing - to reuse a class E address

submariner/cable-driver-protocols-enhancement.md
yboaron: Ack,
I think it makes sense to add to the proposal section that describes the required changes in cloud-prepare to support IPinIP traffic.

submariner/cable-driver-protocols-enhancement.md
```diff
@@ -0,0 +1,76 @@
+# Cable Driver Enhancement - IP-in-IP
+
+## Summary
+
+Submariner today supports three cable drivers: VXLAN, IPSec (tunnel mode) and WireGuard. The
+goal of this proposal is to outline additional enhancements needed for the Submariner Cable Driver
+to support of IP-in-IP for unencrypted connections between clusters or for use as an overlay with
+IPSec in transport mode for encrypted connections between clusters.
```
yboaron, line 8: Is this EP only cover IPinIP as un-encrypted cable driver or also as an overlay with IPSec transport mode?

submariner/cable-driver-protocols-enhancement.md
```diff
@@ -0,0 +1,76 @@
...
+
+For IP in IP, the benefits fall into the following categories:
+
```
yboaron, line 11: It makes sense also to note in the proposal if it's possible or not to run SM IPIP traffic with GW as LB service.

submariner/cable-driver-protocols-enhancement.md
yboaron: Do you think that 1:1 NATing is supported 'out of the box' in GCP,AWS, etc  or extra configuration is required?

sridhargaddam: Sorry, late to the review, but good question Yossi about NAT support with IP-in-IP tunnels. As @maryamtahhan mentioned it seems to work only when we have 1:1 NAT but not otherwise.

```
Important! The IPIP, GRE, EoIP tunnels are of the point-to-point type. Both tunnel participants must have external IP addresses (or be on the same network), and there must be no NAT address translation between them. These are the prerequisites for establishing a tunnel
```
Ref: https://help.keenetic.com/hc/en-us/articles/115002715029-Setting-up-IPIP-GRE-and-EoIP-tunnels

sridhargaddam: > Do you think that 1:1 NATing is supported 'out of the box' in GCP,AWS, etc or extra configuration is required?

The public-ip we assign to the Gateway nodes is like 1:1 NAT, hence it seems to work out of box. But the use-case may not work with Loadbalancer mode.

submariner/cable-driver-protocols-enhancement.md
```diff
@@ -0,0 +1,76 @@
...
+* **Performance**: IP in IP is has a lower overhead than VXLAN which could lead to performance
```
sridhargaddam, line 12:
```suggestion
* **Performance**: IP in IP has a lower overhead than VXLAN which could lead to performance
```

submariner/cable-driver-protocols-enhancement.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Proposal
+
+This enhancement proposes to support IP-in-IP for un-encrypted connection between clusters. A
+new IP tunnelling interface `ipip-tunnel` will be added to the active gateway node of each cluster
+and it will act as the Tunnel End Point. The traffic which is destined to other clusters will be
+forwarded via this interface to the appropriate clusters.
+
+> **_NOTE:_** A single IP Tunnelling interface will be added, and tunnels managed using
+`# ip route add encap ip <args>` in order to avoid a point to point connection per tunnel.
```
sridhargaddam, line 31: Good to know about this.

submariner/cable-driver-protocols-enhancement.md
```diff
@@ -0,0 +1,76 @@
...
+
+### Design Details
+
+* A new interface "ipip-tunnel" will be added.
```
sridhargaddam, line 35: Might be a good idea to include some reference to "submariner" (like sm-ipip-tunnel) in the interface name to convey that its managed by Submariner. I know that we are using "vxlan-tunnel" in case of VxLAN cable driver, but it was more of an oversight and we missed it.

submariner/cable-driver-protocols-enhancement.md
sridhargaddam: >Is there some importance to this decision?

Since we do not have a global IPAM to allocate unique ips to these interface, we took this approach.

submariner/cable-driver-protocols-enhancement.md
sridhargaddam: I have the same question. I see some potential draw-backs with IP-in-IP as a cable driver.
1. Unlike VxLAN cable driver, I'm not sure if we can use IP-in-IP cable driver along with Load-balancer mode.
2. IP-in-IP cable driver cannot work in environments where there is NAT traversal without explicit 1:1 NAT. 

The above points IMHO are important ones to be considered before adding a new cable driver in the code. 

So the question is, should IP-in-IP considered only as an internal overlay tunnel with IPsec transport mode? Even if its the case, can we just use VxLAN itself both as an independent Cable driver as well as overlay tunnel with IPsec transport mode?

sridhargaddam (changes requested):
Can you please clarify the questions raised by me and Yossi.

submariner/cable-driver-protocols-enhancement.md
aswinsuryan: The load balancer supports need to be checked as Sridhar mentioned since it is planned to be the default mode going forward.

maryamtahhan: Re the Load-balancer mode, I guess it depends on the meaning of Load-balancer mode, is it Submariner's Load-balancer mode where it's getting a public IP or is it Load-balancer mode from the POV of an L4 Load Balancer sitting infront of a number of Submariner GWs and distributing packets across them?

Re the working in environments where there is NAT traversal without explicit 1:1 NAT - this is indeed an issue.

Wrapped in IPSec transport mode these issues go away.

But yes, I take your point about VXLAN being usable as both an independent driver as well as an overlay over IPSec transport mode..

submariner/cable-driver-protocols-enhancement.md
maryamtahhan: it's pretty neat :)

submariner/cable-driver-protocols-enhancement.md
mkolesnik: In this case the address selection should be clearly documented, eg is address randomized or is it expected the user using the installer supply it explicitly, or is it taken from a range in the broker, and if so how do we make sure two endpoints dont get the same address

submariner/cable-driver-protocols-enhancement.md
sridhargaddam: > Re the Load-balancer mode, I guess it depends on the meaning of Load-balancer mode, is it Submariner's Load-balancer mode where it's getting a public IP or is it Load-balancer mode from the POV of an L4 Load Balancer sitting infront of a number of Submariner GWs and distributing packets across them?

When Submariner is configured with Loadbalancer mode, we create a standard K8s Service of type Loadbalancer and configure the ports with UDP/4500 (so that IPsec tunnels or vxLAN tunnels can be setup). So its a L4 load balancer.

> But yes, I take your point about VXLAN being usable as both an independent driver as well as an overlay over IPSec transport mode..

I'd like to hear the views of others in the team whether we should add an additional IP-in-IP cable-driver since VxLAN can be used for both use-cases. Basically, having additional driver also implies maintenance/support/CI overhead, so I'm thinking on those lines.

yboaron:
I think IPinIP encapsulation is more suitable than VxLAN for L3 connectivity applications like Submariner.

But checking it in the context of an application that should run behind K8S LBS and the 1: 1 NAT limitation, I don't see the added value in  IPinIP as an unencrypted cable driver.

yboaron:
This is a general note that is not directly related to this EP.

I'm a little confused with the bigger picture of SM datapath changes plan,
there are some ( I think 6) SM datapath related EPs that IMO are related to each other, I think it makes sense to have a high level SM datapath design meeting before discussing the details of each of these proposals.

This kind of discussion should address the following questions for example and set priorities:

- Do we want to refactor GW engine (I think currently GW pod implements 6 logical components)?
 
- Multiple active GW ?

- Do we want to support encrypted and uncrypted cable drivers at the same clusterset?

- Do we want to support multiple cable drivers per family (e.g: IpSEC tunnel and overlay + IpSEC transport for encrypted)?

- Do we want to expose cable driver configuration parameters (e.g: VxLAN configuration details)?

What do you think?

submariner/cable-driver-protocols-enhancement.md
maryamtahhan: For L4 loadbalancer this will be an issue for sure, if the balancer is round robin probably not as much of an issue...

maryamtahhan:
Closing this EP as this will not be part of the 0.13 release.
The reference implementation can be found for submariner [here](https://github.com/submariner-io/submariner/pull/1716)
and for [shipyard](https://github.com/submariner-io/shipyard/pull/748)

The documented manual configuration that formed the bases of the implementation above can be found [here](https://github.com/maryamtahhan/notes/blob/main/Submariner/markdown/SubmarinerIPinIP.md)

The EP was not accepted due to concerns around

- how IP in IP would work with Submariner Loadbalancer mode
- long term maintenance of another cable type

Issue #79, opened 2022-03-17
Rework kubeconfig/kubecontext handling in subctl


skitt:
The various `subctl` commands support a mixture of `KUBECONFIG` environment variables, `--kubeconfig` flags, `--kubecontext` flags, kubeconfig files etc.

We should make this consistent, ideally using `clientcmd`’s tools to handle kubeconfig. https://github.com/skitt/submariner-operator/pull/23 has an old attempt at this, using https://pkg.go.dev/k8s.io/client-go@v0.23.5/tools/clientcmd#BindOverrideFlags to provide access to all the features supported by `clientcmd` (see `kubectl options`).

For commands using a single context, _all_ the following should work, with kubeconfigs containing one or more clusters and/or contexts (this uses a `kube` prefix when setting up `clientcmd`):

```
# Use the default context in the kubeconfig
KUBECONFIG=/path/to/kubeconfig subctl foo
subctl foo --kubeconfig /path/to/kubeconfig
subctl --kubeconfig /path/to/kubeconfig foo

# Use the specified context from the kubeconfig
KUBECONFIG=/path/to/kubeconfig subctl foo --kubecontext bar
subctl foo --kubeconfig /path/to/kubeconfig --kubecontext bar
subctl --kubeconfig /path/to/kubeconfig --kubecontext bar foo
```

For commands using multiple contexts, the prefixes should make the purpose of each context clear, using appropriate prefixes when setting up `clientcmd`:

```
KUBECONFIG=/path/to/kubeconfig subctl benchmark latency --fromcontext foo --tocontext bar
subctl benchmark latency --kubeconfig /path/to/kubeconfig --fromcontext foo --tocontext bar
```

`--kubeconfig` is preserved as-is, since users can use a single file combining all their contexts. However such commands should also support separate kubeconfigs:

```
subctl benchmark latency --fromconfig /path/to/fromconfig --toconfig /path/to/toconfig
```

`clientcmd` _doesn’t_ handle `--kubeconfig` itself, that needs to be handled by `subctl`.

This enhancement should also include a test suite covering all the different possibilities. As far as possible, the 0.12 options should keep their existing behaviour, and be marked for deprecation; they will be removed after 0.13.

This would subsume submariner-io/subctl#49

dfarrell07:
[...]

PR #101, opened 2022-04-04
Improve kubeconfig handling

skitt:
This is the enhancement PR for #79.

Signed-off-by: Stephen Kitt <skitt@redhat.com>

### DIFF
--- a/submariner-operator/issue-79-kubeconfig-handling.md
+++ b/submariner-operator/issue-79-kubeconfig-handling.md
@@ -1,0 +1,132 @@
+## Epic Description
+
+The various `subctl` commands support a mixture of `KUBECONFIG` environment variables, `--kubeconfig` flags, `--kubecontext` flags,
+kubeconfig files etc.
+
+We should make this consistent, ideally using `clientcmd`’s tools to handle kubeconfig.
+[Operator PR #23](https://github.com/skitt/submariner-operator/pull/23) has an old attempt at this,
+using [`BindOverrideFlags()`](https://pkg.go.dev/k8s.io/client-go@v0.23.5/tools/clientcmd#BindOverrideFlags) to provide access
+to all the features supported by `clientcmd` (see `kubectl options`).
+
+On top of that, we should align our flags with `kubectl`’s;
+beyond helping users already familiar with `kubectl` flags,
+this will make it easier to turn `subctl` into a `kubectl` plugin,
+should we choose to do so.
+This would involve turning `--kubecontext` into `--context`;
+`--kubecontext` would be preserved as a deprecated alias for `--context` for one release,
+and removed in the following release.
+
+## Acceptance Criteria
+
+For commands using a single context, _all_ the following should work,
+with kubeconfigs containing one or more clusters and/or contexts:
+
+```bash
+# Use the default context in the kubeconfig
+KUBECONFIG=/path/to/kubeconfig subctl foo
+subctl foo --kubeconfig /path/to/kubeconfig
+subctl --kubeconfig /path/to/kubeconfig foo
+
+# Use the specified context from the kubeconfig
+KUBECONFIG=/path/to/kubeconfig subctl foo --context bar
+subctl foo --kubeconfig /path/to/kubeconfig --context bar
+subctl --kubeconfig /path/to/kubeconfig foo --context bar
+subctl --kubeconfig /path/to/kubeconfig --context bar foo
+```
+
+(and other variations on option order).
+
+For commands using multiple contexts,
+there should be one “main” set of flags, with no prefix,
+and another set with a prefix describing the purpose of the context:
+
+```bash
+KUBECONFIG=/path/to/kubeconfig subctl benchmark latency --context foo --tocontext bar
+subctl benchmark latency --kubeconfig /path/to/kubeconfig --context foo --tocontext bar
+```
+
+`--kubeconfig` is preserved as-is, since users can use a single file combining all their contexts.
+However such commands should also support separate kubeconfigs, to allow using kubeconfigs which
+contain conflicting configuration (_e.g._ kubeconfigs using the same context name and user name,
+as produced by the OpenShift installer):
+
+```bash
+subctl benchmark latency --kubeconfig /path/to/fromconfig --toconfig /path/to/toconfig
+```
+
+These can be combined:
+
+```bash
+subctl benchmark latency --kubeconfig /path/to/fromconfig --context foo --toconfig /path/to/toconfig --tocontext bar
+```
+
+If different kubeconfigs are specified, each context will only be loaded from the corresponding
+kubeconfig.
+
+(When presented with conflicting configurations, `kubectl` only uses the first relevant context,
+the other contexts are inaccessible unless the conflicts are resolved.
+For example, with OpenShift `kubeconfig` files and their `admin` context, only the `admin` context
+in the first file listed in `$KUBECONFIG` can be used.)
+
+For commands potentially using _all_ accessible contexts (`subctl gather`), the existing behaviour
+should be preserved:
+
+```bash
+# Use all accessible contexts
+KUBECONFIG=/path/to/kubeconfig subctl gather
+subctl gather --kubeconfig /path/to/kubeconfig
+subctl gather /path/to/kubeconfig
+
+# Use one named context
+KUBECONFIG=/path/to/kubeconfig subctl gather --context context1
+subctl gather --kubeconfig /path/to/kubeconfig --context context1
+subctl gather /path/to/kubeconfig --context context1
+
+# Use multiple named contexts
+KUBECONFIG=/path/to/kubeconfig subctl gather --contexts context1,context2
+subctl gather --kubeconfig /path/to/kubeconfig --contexts context1,context2
+subctl gather /path/to/kubeconfig --contexts context1,context2
+```
+
+`clientcmd` _doesn’t_ handle `--kubeconfig` itself, that needs to be handled by `subctl`.
+
+This enhancement should also include a test suite covering all the different possibilities.
+As far as possible, the options present in the previous release should keep their existing behaviour,
+and be marked for deprecation so that they can be removed in a future release.
+
+This would subsume [`--kubecontexts` support in `subctl diagnose`](https://github.com/submariner-io/submariner-operator/issues/1327).
+
+## Work Items
+
+* Create a test suite for the various `--kubeconfig` and `--kubecontext` variants
+* Delegate flag setup for `--kubecontext` and related flags to `k8s.io/client-go/tools/clientcmd`
+* Integrate clientcmd handling into `restconfig.Producer` (if it still exists)
+* Where appropriate, add `--kube` variants for command which need multiple contexts
+
+It will probably be necessary to isolate the `clientcmd` data structures in each command using them,
+to ensure that uninitialised data structures can’t be accidentally used.
+See [`kubectl apply`](https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/apply/apply.go) for
+an example of data structure use to isolate flags and options.
+
+This work will flip the way `clientcmd` settings are handled.
+Each command using `clientcmd` will need to maintain a `clientcmd.ConfigOverrides` instance as part
+of its state (_e.g._ in a structure describing its settings), and “hook up” the various flags to
+fields in that structure: `--kubeconfig` maps to `clientcmd.DefaultClientConfig.ExplicitPath`,
+the remaining flags are set up with `clientcmd.BindOverrideFlags` using an appropriate prefix
+(empty in most cases, so the flags end up being `--context` etc.).
+The plural form, `--contexts`, will need special handling.
+
+The following illustrates the initial setup:
+
+```go
+// addKubeContextFlag adds a "kubeconfig" flag and a single "context" flag that can be used once and only once
+func addKubeContextFlag(cmd *cobra.Command) {
+    loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
+    loadingRules.DefaultClientConfig = &clientcmd.DefaultClientConfig
+    overrides := clientcmd.ConfigOverrides{}
+    kflags := clientcmd.RecommendedConfigOverrideFlags("")
+    addKubeConfigFlag(cmd, &loadingRules.ExplicitPath)
+    clientcmd.BindOverrideFlags(&overrides, cmd.PersistentFlags(), kflags)
+    clientConfig = clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, &overrides)
+}
+```


submariner-operator/issue-79-kubeconfig-handling.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Acceptance Criteria
+
+For commands using a single context, _all_ the following should work,
+with kubeconfigs containing one or more clusters and/or contexts (this uses a `kube` prefix when setting up `clientcmd`):
+
+```bash
+# Use the default context in the kubeconfig
+KUBECONFIG=/path/to/kubeconfig subctl foo
+subctl foo --kubeconfig /path/to/kubeconfig
```
sridhargaddam, line 27: In the current implementation when I have a kubeconfig with multiple contexts and use `subctl gather ...` command, it executes the command for each context. What would be the behaviour for such commands according to this proposal?

submariner-operator/issue-79-kubeconfig-handling.md
```diff
@@ -0,0 +1,101 @@
...
+For commands using multiple contexts, the prefixes should make the purpose of each context clear,
+using appropriate prefixes when setting up `clientcmd`:
+
+```bash
+KUBECONFIG=/path/to/kubeconfig subctl benchmark latency --fromcontext foo --tocontext bar
+subctl benchmark latency --kubeconfig /path/to/kubeconfig --fromcontext foo --tocontext bar
+```
+
+`--kubeconfig` is preserved as-is, since users can use a single file combining all their contexts.
+However such commands should also support separate kubeconfigs:
```
mkolesnik: Why should they?

I find such UX confusing and error prone, e.g. which one would take precedence?

Given that `--kubeconfig` can support multiple configs separated by `:`, why isn't it sufficient?

submariner-operator/issue-79-kubeconfig-handling.md
```diff
@@ -0,0 +1,101 @@
...
+subctl foo --kubeconfig /path/to/kubeconfig
+subctl --kubeconfig /path/to/kubeconfig foo
+
+# Use the specified context from the kubeconfig
+KUBECONFIG=/path/to/kubeconfig subctl foo --kubecontext bar
+subctl foo --kubeconfig /path/to/kubeconfig --kubecontext bar
+subctl --kubeconfig /path/to/kubeconfig --kubecontext bar foo
+```
+
+For commands using multiple contexts, the prefixes should make the purpose of each context clear,
```
mkolesnik: Should they?

Given the `benchmark` example, why wouldnt we support multiple contexts and run benchmarks between the different combinations?

I don't see a particular reason to limit the commands using the flags, but rather support versatile usage according to the capability of the command.

submariner-operator/issue-79-kubeconfig-handling.md
```diff
@@ -0,0 +1,101 @@
...
+* [ ] Create a test suite for the various `--kubeconfig` and `--kubecontext` variants
+* [ ] Delegate flag setup for `--kubecontext` and related flags to `k8s.io/client-go/tools/clientcmd`
+* [ ] Integrate clientcmd handling into `restconfig.Producer`
+* [ ] Where appropriate, add `--kube` variants for command which need multiple contexts
+
+It will probably be necessary to isolate the `clientcmd` data structures in each command using them,
+to ensure that uninitialised data structures can’t be accidentally used.
+For example, a command using `--from` and `--to` instead of `--kube` shouldn’t end up with unused
+`--kube` flag variables.
+See [`kubectl apply`](https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/apply/apply.go) for
```
mkolesnik, line 108: Not sure what you're referring to, I don't see `clientcmd` used there.

submariner-operator/issue-79-kubeconfig-handling.md
skitt: You asked for this a year or so ago, I’ll try to find the reference.

submariner-operator/issue-79-kubeconfig-handling.md
skitt: See the updated proposal.

submariner-operator/issue-79-kubeconfig-handling.md
skitt: See the updated proposal.

Multiple configs separated by `:` isn’t sufficient for two reasons: `clientcmd` expects to build a consolidated config with all the kubeconfig entries; and the various kubeconfigs aren’t given explicit purposes. I know we removed support for this a while back, but given that the OpenShift installer isn’t fixing the way it builds kubeconfigs, I think we should restore it so that OpenShift kubeconfigs can be used directly.

submariner-operator/issue-79-kubeconfig-handling.md
skitt: This proposal doesn’t cover turning commands like `benchmark` into matrices. Currently there’s a “from” context and a “to” context, this proposal aims to make that explicit. If we turn `benchmark` into an _n_–_n_ command, we can revisit the context flag naming then.

The goal isn’t to limit commands using flags, but to ensure that the use each command makes of its contexts is explicit. The flag names serve the purposes of the commands, not the other way round; if a command’s behaviour changes, its flag names should also change to suit.

submariner-operator/issue-79-kubeconfig-handling.md
skitt: See the updated proposal.

submariner-operator/issue-79-kubeconfig-handling.md
mkolesnik: Arguably on `benchmark` it actually has two modes - either two context of a single context, for intra-cluster testing, so how would intra-cluster work if I specify just one of the contexts (e.g. just `--fromcontext`)?
Alternatively if I supply multiple configs in `--kubeconfig` and don't specify a context would it fail or not (currently it would if two configs are given for this scenario).

I feel the UX would be simpler if we use just `--kubeconfig` and `--kubecontexts`, or at least if we use same `--kubeconfig` and perhaps several `--[prefix]context` flags.

skitt: > I feel the UX would be simpler if we use just `--kubeconfig` and `--kubecontexts`, or at least if we use same `--kubeconfig` and perhaps several `--[prefix]context` flags.

My proposal uses the latter approach, see the example given a few lines further down.

As far as `--intra-cluster` handling goes, I don’t know what the best approach is; I think the existing UX here is sub-optimal. Perhaps we could use different context prefixes to choose between the modes: `--fromcontext`/`--tocontext` to use inter-cluster tests, `--kubecontext` on its own to use intra-cluster tests.

dfarrell07:
@mkolesnik @sridhargaddam Can you loop back to the comments here?

If not, let's go ahead and merge this, marking as deferred if we need to.

submariner-operator/issue-79-kubeconfig-handling.md
```diff
@@ -0,0 +1,122 @@
...
+KUBECONFIG=/path/to/kubeconfig subctl gather --kubecontexts context1,context2
+subctl gather --kubeconfig /path/to/kubeconfig --kubecontexts context1,context2
+subctl gather /path/to/kubeconfig --kubecontexts context1,context2
+```
+
+`clientcmd` _doesn’t_ handle `--kubeconfig` itself, that needs to be handled by `subctl`.
+
+This enhancement should also include a test suite covering all the different possibilities.
+As far as possible, the 0.12 options should keep their existing behaviour, and be marked for deprecation;
+they will be removed after 0.13.
```
tpantelis:
```suggestion
As far as possible, the prior release options should keep their existing behaviour, and be marked for deprecation for removal in a future release.
```

submariner-operator/issue-79-kubeconfig-handling.md
```diff
@@ -0,0 +1,122 @@
...
+* [ ] Troubleshooting (gather/diagnose) added
+* [ ] Documentation added
+* [ ] Release notes added
+
+## Work Items
+
+* [ ] Create a test suite for the various `--kubeconfig` and `--kubecontext` variants
+* [ ] Delegate flag setup for `--kubecontext` and related flags to `k8s.io/client-go/tools/clientcmd`
+* [ ] Integrate clientcmd handling into `restconfig.Producer`
+* [ ] Where appropriate, add `--kube` variants for command which need multiple contexts
```
tpantelis: I don't think these are needed in the EP as they are part of the epic issue template.

tpantelis (changes requested):

submariner-operator/issue-79-kubeconfig-handling.md
```diff
@@ -0,0 +1,122 @@
...
+specified, without fallback to a generic `--kubeconfig` flag or the `KUBECONFIG` environment
+variable.
+
+For commands potentially using _all_ accessible contexts (`subctl gather`), the existing behaviour
+should be preserved:
+
+```bash
+# Use all accessible contexts
+KUBECONFIG=/path/to/kubeconfig subctl gather
+subctl gather --kubeconfig /path/to/kubeconfig
```
sridhargaddam, line 77: If we have a kubeconfig with two contexts pointing to two different clusters, will the `summary.html` file generated as part of `gather output` be overridden (when gather is executed for the second context) or does it contain entries for both the clusters?

mkolesnik (commented):
Given that we recently discussed using subctl as a `kubectl` plugin, I have a simpler (IMHO) approach:

Adopt standard options `--kubeconfig` and `--context`, as they're used in `kubectl`, for all commands.

Single context commands are obviously fine with that approach and need nothing further.

For backwards compatibility we could retain the `kubecontext` flag for a while, but as a plugin I think it would be cleaner to adopt `context` as the main `kubectl` uses.

Dual context commands (e.g. `verify`) can add additional flag to specify other needed contexts, e,g, `subctl verify --context cluster1 --targetcontext cluster2`.
For `benchmark` we could use the same strategy, with logic that if a target context is specified then that is used, and otherwise the benchmark is intra-cluster (obviating that flag).

For multi-cluster commands (e.g. `gather`) we could introduce additional flags, but honestly I think from a UX perspective a user may want to gather from a single context, or from all, and not cherry pick the clusters, so we could have it gather from all available contexts unless an explicit one is requested.

skitt:
I like the idea of aligning with `kubectl`, I’ve reworked this PR to do that (as a separate commit, purposefully without DCO so it can’t be merged without squashing).

submariner-operator/issue-79-kubeconfig-handling.md
skitt: `summary.html` is already generated per cluster.

submariner-operator/issue-79-kubeconfig-handling.md
```diff
@@ -0,0 +1,113 @@
...
+
+```bash
+KUBECONFIG=/path/to/kubeconfig subctl benchmark latency --context foo --tocontext bar
+subctl benchmark latency --kubeconfig /path/to/kubeconfig --context foo --tocontext bar
+```
+
+`--kubeconfig` is preserved as-is, since users can use a single file combining all their contexts.
+However such commands should also support separate kubeconfigs, to allow using kubeconfigs which
+contain conflicting configuration (_e.g._ kubeconfigs using the same context name and user name,
+as produced by the OpenShift installer):
```
mkolesnik, line 51: BTW what does `kubectl` do in such case?

Does it just ignore the fact that you can use multiple conflicting configs, or does it produce an error?

mkolesnik (approved):
Generally LGTM.

Only thing you didnt explicitly address is if we're keeping `--kubecontext` flag and depracating it, or cutting  it all together?
Generally I see `subctl` as porcelain, as you can do all of it's functions directly against standard K8s API, but we never explicitly marked it as such.

skitt:
> Generally LGTM.
> 
> Only thing you didnt explicitly address is if we're keeping `--kubecontext` flag and depracating it, or cutting it all together? Generally I see `subctl` as porcelain, as you can do all of it's functions directly against standard K8s API, but we never explicitly marked it as such.

This was supposed to be described by https://github.com/submariner-io/enhancements/pull/101/files#diff-cc0b5da732814eba0e75977ad54b1aa16beba895ceaf8fb963159061e7c8f752R75-R76 but I can make it more explicit if necessary.

mkolesnik:
> This was supposed to be described by https://github.com/submariner-io/enhancements/pull/101/files#diff-cc0b5da732814eba0e75977ad54b1aa16beba895ceaf8fb963159061e7c8f752R75-R76 but I can make it more explicit if necessary.

Yes I saw that already but its a bit amorphous, but I'm fine with it so it's your call

submariner-operator/issue-79-kubeconfig-handling.md
skitt: It picks the first which matches whatever context is specified, and ignores the rest, with no error. This can be used to resolve the situation; _e.g._ with multiple `admin` contexts, `kubectl config rename-context admin foo` will rename the first `admin` context, and make the next one accessible using the `admin` name.

mkolesnik: We can go that route then if we want to stay closely aligned, though I think it's not critical either way (ie adding another kubeconfig param)

submariner-operator/issue-79-kubeconfig-handling.md
```diff
@@ -0,0 +1,120 @@
...
+```bash
+# Use the default context in the kubeconfig
+KUBECONFIG=/path/to/kubeconfig subctl foo
+subctl foo --kubeconfig /path/to/kubeconfig
+subctl --kubeconfig /path/to/kubeconfig foo
+
+# Use the specified context from the kubeconfig
+KUBECONFIG=/path/to/kubeconfig subctl foo --context bar
+subctl foo --kubeconfig /path/to/kubeconfig --context bar
+subctl --kubeconfig /path/to/kubeconfig --context bar foo
```
dfarrell07, line 34:
```suggestion
subctl --kubeconfig /path/to/kubeconfig foo --context bar
```

Should this work?

submariner-operator/issue-79-kubeconfig-handling.md
```diff
@@ -0,0 +1,120 @@
...
+subctl benchmark latency --kubeconfig /path/to/kubeconfig --context foo --tocontext bar
+```
+
+`--kubeconfig` is preserved as-is, since users can use a single file combining all their contexts.
+However such commands should also support separate kubeconfigs, to allow using kubeconfigs which
+contain conflicting configuration (_e.g._ kubeconfigs using the same context name and user name,
+as produced by the OpenShift installer):
+
+```bash
+subctl benchmark latency --kubeconfig /path/to/fromconfig --toconfig /path/to/toconfig
```
dfarrell07, line 54:
```suggestion
subctl benchmark latency --kubeconfig /path/to/fromconfig --toconfig /path/to/toconfig --context foo --tocontext bar
```

Would it be valid to provide both config and context of main/to flags?

submariner-operator/issue-79-kubeconfig-handling.md
skitt: Yes, it should.

submariner-operator/issue-79-kubeconfig-handling.md
skitt: It would indeed!

dfarrell07 (approved):
Thanks for adding those examples, LGTM!

tpantelis (approved):

PR #102, opened 2022-04-04
EP: IPSec Central CA

maryamtahhan:
This is a proposal for a Central Certificate Authority that can sign
Certificate Signing Requests for Submariner GWs that can be used
for the IPSec connections instead of the PSK. 


submariner/ipsec-cert-issuer.md
```diff
@@ -0,0 +1,114 @@
+# IPSec Certificate Issuer
+
+## Summary
+
+Submariners current security arrangements, namely using an identical pre-shared key in
+all deployments will fail security assessment in most environments which have to comply
+with a pre-existing security policy and rules on using VPNs in general and IPSEC in particular.
+
+In order to comply the following minimum enhancement that needs to be implemented is the use
```
maryamtahhan:
```suggestion
In order to comply the minimum enhancement that needs to be implemented is the use
```

submariner/ipsec-cert-issuer.md
```diff
@@ -0,0 +1,114 @@
...
+of X509 certificates per Gateway. This will require the introduction of a Certificate Authority
+(CA) that's capable of signing Certificate Signing Requests (CSRs) for Gateways in the
+`ClusterSet`. This CA will be serviced by a Syncer that will act as a proxy between the Certificate
+Requester and the CA itself.
+
+## Proposal
+
+This enhancement proposes to use a `ClusterSet` wide `Certificate Authority` and a `Syncer`
+that is co-located with the Broker to service Gateway. At a high level this requires:
```
maryamtahhan:
```suggestion
that are co-located with the Broker to service Gateway. At a high level this requires:
```

submariner/ipsec-cert-issuer.md
```diff
@@ -0,0 +1,114 @@
...
+
+    // The PEM-encoded x509 certificate signing request to be submitted to the
+    // CA for signing.
+    Request []byte
+
+    // Usages is the set of x509 usages that are requested for the certificate.
+    // "signing", "digital signature", "key agreement"...
+    // details can be found here: https://www.rfc-editor.org/rfc/rfc5280#section-4.2.1.3
+    // and https://github.com/cert-manager/cert-manager/blob/master/internal/apis/certmanager/types.go#L159
+    Usages [] string
```
maryamtahhan:
```suggestion
    Usages []string
```

submariner/ipsec-cert-issuer.md
```diff
@@ -0,0 +1,114 @@
...
+on the Broker node.
+* A CA_CSR_Syncer on the Broker Node.
+* A new CRD to encapsulate a Submariner CSR (towards this centralized CA), and the result (towards
+the Gateway). Gateways will fill the CRD with relevant information (a base64 encoded string of a PEM
+encoded certificate request) and publish the CR to the Kube API.
+* A GW_CSR_Syncer on the gateway Node.
+
+The general idea is that, when a Broker is installed, a CA and CA_CSR_Syncer are also installed on
+the Broker node. Then, when a Submariner gateway starts up it will launch a GW_CSR_Syncer, it will
+also create a CSR, base64 encode it and send it via a Submariner CSR CRD towards the Broker node CA.
```
skitt: Could we use general-purpose Kubernetes CSRs instead? _I.e._ have the client gateway directly submit a CSR to the broker cluster, instead of going through syncers.

skitt:
For extra bonus points, it would be great if this could be replaced with ACM’s certificate signing setup when running on ACM. (I’ve been meaning to look into how that’s used...)

submariner/ipsec-cert-issuer.md
maryamtahhan: Just so we are clear are you suggesting:

1. Using the Broker Node Kubernetes API/CP as the dedicate 
ClusterSet CA and getting rid of the proposed Submariner CSR CRD? so then the syncer on the local/GW node forwards/exports CSRs to the broker node (via federator) and the rest of the process from there on out is the same (sign, publish result, publish back to GW)? OR

2. Just using the Kubernetes CSRs but still using something else for the CA such as cert-manager? 

I guess for both options 1 and 2 I need to look into the following questions:

- How would we distinguish between local (Pod/svc) CSRs vs the GW CSRs to make sure that the GW CSRs are not signed locally... and
- does the K8s CSR facility do any sort of verification of svcs/pods in the cluster before signing a cert?

The reason why I suggested cert-manager was that you can really configure different CAs behind the cert-manager API I also thought it was what ACM uses and potentially for the GW verification we may want to have an independent CA to what's provided in the scope of a local cluster.

elevran: Using the `spec.signerName` may offer a way to distinguish whom should be signing the CSR.

submariner/ipsec-cert-issuer.md
```diff
@@ -0,0 +1,114 @@
+# IPSec Certificate Issuer
+
+## Summary
+
+Submariners current security arrangements, namely using an identical pre-shared key in
+all deployments will fail security assessment in most environments which have to comply
+with a pre-existing security policy and rules on using VPNs in general and IPSEC in particular.
+
+In order to comply the minimum enhancement that needs to be implemented is the use
+of X509 certificates per Gateway. This will require the introduction of a Certificate Authority
```
elevran: I most likely missed it, but am unclear on how would you establish trust chain with the CA? Its certificate bundle should be present on each GW so it needs to be sync'ed.

submariner/ipsec-cert-issuer.md
elevran: Some CSR verification in https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/#kubernetes-signers.

skitt: Right, that’s pretty much what I’m wondering. We probably still want the cert-manager API, as used in ACM, but I don’t think we need to add our own syncing mechanism. I know the addons for example have CSRs, signed using cert-manager without any other syncing mechanism, but I haven’t looked into the details — in particular, whether spoke-to-hub CSRs are supported.

submariner/ipsec-cert-issuer.md
maryamtahhan: The CAs cert can be synced to the clusters in the same way the self signed broker cert is synced today (via a volume mounted secret). On GW startup this cert could be read from the volume and then loaded into /etc/pki/tls/certs/ca-bundle.crt
I will add more details on this

maryamtahhan:
So the OCM CSR issuing process is described [here](https://github.com/open-cluster-management-io/api/blob/main/docs/clusterjoinprocess.md) and implementation (from managed cluster POV) is [here](https://github.com/open-cluster-management-io/registration/blob/892a6ea78295bfa3dc012b890f80017cf9172e8f/pkg/spoke/managedcluster/registration.go) and the controller on the hub is [here](https://github.com/open-cluster-management-io/registration/blob/e6435f32a3ed877b974f3299bd26db1101162f7b/pkg/hub/csr/controller.go)

My over simplified understanding there's a certificate controller on the hub managing certificate approvals/issuing/renewals and verifying the CSR requestor. At managed cluster registration time, the agent on the managed cluster generates a CSR on behalf of the cluster towards the controller in the hub. The controller in the Hub approves the managed cluster and issues the relevant secret which is then propagated to the managed cluster as a secret.
For renewing the cert - the agent on the managed cluster initiates the renewal request on behalf of the managed cluster... 

It also looks like cert manager can be used alongside K8s certificates API as a signer however it is `experimental` https://cert-manager.io/docs/usage/kube-csr/... so probably best to avoid... 

If we are using the K8s certificates API directly with the K8s signer - we would still need a controller on the Broker node that validates/signs the certs... and still need an agent on the clusters to sync the requests to the broker and the results to the GW... 

All that being said, the discussion comment above and the design in OCM reusing the K8s Certificates API should be possible:

This means that on the broker node, submariner needs to:
- Create a controller that will manage certificate requests and validation of csr issuers. 
- Create the central CA cert and propagate to all cluster gateways as a mounted secret similar to the Broker CA.crt

On the clusters nodes:

- At GW startup, the GW will need to install the central CA cert, and it will still need to generate a private key and a CSR.
- The GW will need to periodically check if the cert is due to expire soon and issue a renewal... 
- Using the existing mechanisms in Submariner today, a syncer will still be required from what I can see to sync the CSRs at Gateway startup time to the Broker API and sync back the result (signed cert chain)... Unless we have some agent on the GW node that directly interacts with the broker API on our behalf... This of course introduces the issue of having CSRs synced to all clusters in the clusterset and introduces the need on the broker to redirect the response to a particular cluster via selected imports? 
If not a syncer - then modify admiral to introduce some API to allow us to submit CSRs to the Broker API and retrieve the results directly...

Thoughts?

maryamtahhan:
> * This of course introduces the issue of having CSRs synced to all clusters in the clusterset and introduces the need on the broker to redirect the response to a particular cluster via selected imports?

SourceLabelSelector and SourceFieldSelector on the Syncer with labels and annoations could resolve this issue... Thanks @donaldh for the suggestion.

submariner/ipsec-cert-issuer.md
```diff
@@ -0,0 +1,114 @@
...
+(CA) that's capable of signing Certificate Signing Requests (CSRs) for Gateways in the
+`ClusterSet`. This CA will be serviced by a Syncer that will act as a proxy between the Certificate
+Requester and the CA itself.
+
+## Proposal
+
+This enhancement proposes to use a `ClusterSet` wide `Certificate Authority` and a `Syncer`
+that are co-located with the Broker to service Gateway. At a high level this requires:
```
mkolesnik: Does it have to be co-located with the broker?

submariner/ipsec-cert-issuer.md
```diff
@@ -0,0 +1,114 @@
+# IPSec Certificate Issuer
+
+## Summary
+
+Submariners current security arrangements, namely using an identical pre-shared key in
+all deployments will fail security assessment in most environments which have to comply
+with a pre-existing security policy and rules on using VPNs in general and IPSEC in particular.
+
+In order to comply the minimum enhancement that needs to be implemented is the use
```
mkolesnik: Can we also discuss alternatives?

I see CA is not always preferred to PSK [1] so I think we can also consider the alternative security model and pros and cons of each.

[1] https://community.cisco.com/t5/vpn/ipsec-authentication-using-x509-certificates/m-p/3215190/highlight/true#M119210

skitt:
I was thinking of the CSR approval mechanism used in the OCM Submariner addon; this relies on the OCM addon framework, see https://github.com/open-cluster-management-io/addon-framework/blob/main/pkg/addonmanager/controllers/certificate/csrapprove.go

So obviously this can’t be used as-is in Submariner, but I was wondering if there was anything useful to learn from that.

Ultimately I don’t think I have anything to add to https://github.com/submariner-io/enhancements/pull/102#issuecomment-1092827953. Given that Submariner already has syncing mechanisms in place, we might as well use those instead of exposing a new signing service.

submariner/ipsec-cert-issuer.md
skitt: The alternative is to ensure we use one PSK per connected cluster. This would require inventing our own ways of authenticating clients and revoking PSKs; CAs provide standardised ways of handling all this.

maryamtahhan:
> I was thinking of the CSR approval mechanism used in the OCM Submariner addon; this relies on the OCM addon framework, see https://github.com/open-cluster-management-io/addon-framework/blob/main/pkg/addonmanager/controllers/certificate/csrapprove.go
> 
> So obviously this can’t be used as-is in Submariner, but I was wondering if there was anything useful to learn from that.

I will have a look. @donaldh also suggested potentially looking at creating intermediate CAs locally --- I will take a look at both and circle back

submariner/ipsec-cert-issuer.md
mkolesnik: I think it would be useful to capture this in the EP

submariner/ipsec-cert-issuer.md
```diff
@@ -0,0 +1,114 @@
+# IPSec Certificate Issuer
+
+## Summary
+
+Submariners current security arrangements, namely using an identical pre-shared key in
```
dfarrell07:
```suggestion
Submariner's current security arrangements, namely using an identical pre-shared key in
```

submariner/ipsec-cert-issuer.md
```diff
@@ -0,0 +1,114 @@
...
+all deployments will fail security assessment in most environments which have to comply
```
dfarrell07:
```suggestion
all deployments, will fail security assessment in most environments which have to comply
```

submariner/ipsec-cert-issuer.md
dfarrell07:
```suggestion
In order to comply, the minimum enhancement that needs to be implemented is the use
```
IMHO

submariner/ipsec-cert-issuer.md
```diff
@@ -0,0 +1,114 @@
...
+
+In order to comply the minimum enhancement that needs to be implemented is the use
+of X509 certificates per Gateway. This will require the introduction of a Certificate Authority
+(CA) that's capable of signing Certificate Signing Requests (CSRs) for Gateways in the
+`ClusterSet`. This CA will be serviced by a Syncer that will act as a proxy between the Certificate
+Requester and the CA itself.
+
+## Proposal
+
+This enhancement proposes to use a `ClusterSet` wide `Certificate Authority` and a `Syncer`
```
dfarrell07:
```suggestion
This enhancement proposes to use a `ClusterSet`-wide `Certificate Authority` and a `Syncer`
```

submariner/ipsec-cert-issuer.md
```diff
@@ -0,0 +1,114 @@
...
+Monitor the result of the CA CSR and publish the result back to the original Gateway via the Broker
+(by updating the original CSR CRD). The details of the design are shown in the upcoming sections.
+
+## Design Details
+
+### CA Deployment
+
+The following diagram shows the various component needed for the central CA.
+
+![CA Deployment](http://www.plantuml.com/plantuml/proxy?cache=no&src=https://raw.githubusercontent.com/maryamtahhan/enhancements/feat_central_cert_manager/submariner/images/CA-deploy.puml)
```
dfarrell07: After this is merged, we should remember to update these source URLs to reference the UMLs that will be merged as a part of this PR.

submariner/ipsec-cert-issuer.md
```diff
@@ -0,0 +1,114 @@
...
+## Proposal
+
+This enhancement proposes to use a `ClusterSet` wide `Certificate Authority` and a `Syncer`
+that are co-located with the Broker to service Gateway. At a high level this requires:
+
+* A Certificate Authority that can be used across the ClusterSet. Ideally this authority resides
+on the Broker node.
+* A CA_CSR_Syncer on the Broker Node.
+* A new CRD to encapsulate a Submariner CSR (towards this centralized CA), and the result (towards
+the Gateway). Gateways will fill the CRD with relevant information (a base64 encoded string of a PEM
```
dfarrell07:
```suggestion
the Gateway). Gateways will fill the CRD with relevant information (a base64-encoded string of a PEM-
```

maryamtahhan:
> > I was thinking of the CSR approval mechanism used in the OCM Submariner addon; this relies on the OCM addon framework, see https://github.com/open-cluster-management-io/addon-framework/blob/main/pkg/addonmanager/controllers/certificate/csrapprove.go
> > So obviously this can’t be used as-is in Submariner, but I was wondering if there was anything useful to learn from that.
> 
> I will have a look. @donaldh also suggested potentially looking at creating intermediate CAs locally --- I will take a look at both and circle back

I had a look at the addon csr signing controller - it seems to work in a similar way to the central csr signing controller in OCM in that the addon manager is serving as a custom CSR signer controller for renewing CSRs for managed clusters...
I'm not sure that it adds anything more for us, so going back to the suggestions
https://github.com/submariner-io/enhancements/pull/102#issuecomment-1092827953 I will push the relevant updates to the proposal

submariner/ipsec-cert-issuer.md
dfarrell07:
```suggestion
all clusters in a ClusterSet will fail security assessment in most environments which have to comply
```

Is this equivalent to what you mean? If so, it might be more clear to be a bit more explicit.

submariner/ipsec-cert-issuer.md
```diff
@@ -0,0 +1,114 @@
...
+    Usages [] string
+}
+
+// SubmarinerCSRStatus defines the status of CertificateRequest and
+// resulting signed certificate.
+type SubmarinerCSRStatus struct {
+    // Status of the request (`Ready`, `InvalidRequest`, `Approved`, `Denied`).
+    Status string
+
+    // The PEM encoded x509 certificate resulting from the certificate
```
dfarrell07:
```suggestion
    // The PEM-encoded X.509 certificate resulting from the certificate
```

submariner/ipsec-cert-issuer.md
```diff
@@ -0,0 +1,114 @@
...
+    // signing request. Empty if the Status is InvalidRequest/Denied.
+    Certificate []byte
+
+    // The PEM encoded x509 certificate of the Certificate Authority.
```
dfarrell07:
```suggestion
    // The PEM-encoded X.509 certificate of the Certificate Authority.
```

submariner/ipsec-cert-issuer.md
dfarrell07:
```suggestion
of X.509 certificates per Gateway. This will require the introduction of a Certificate Authority
```

Seems to be the consistent and [official](https://datatracker.ietf.org/doc/html/rfc5280) usage.

submariner/ipsec-cert-issuer.md
```diff
@@ -0,0 +1,114 @@
...
+    // Result of CSR.
+    CSRStatusSpec SubmarinerCSRStatus `json:"spec"`
+}
+
+// SubmarinerCSRSpec defines the CSR for the Submariner Gateway
+type SubmarinerCSRSpec struct {
+    // The requested 'duration' (i.e. lifetime) of the Certificate.
+    Duration *metav1.Duration
+
+    // The PEM-encoded x509 certificate signing request to be submitted to the
```
dfarrell07:
```suggestion
    // The PEM-encoded X.509 certificate signing request to be submitted to the
```

submariner/ipsec-cert-issuer.md
```diff
@@ -0,0 +1,114 @@
...
+    // CA for signing.
+    Request []byte
+
+    // Usages is the set of x509 usages that are requested for the certificate.
```
dfarrell07:
```suggestion
    // Usages is the set of X.509 usages that are requested for the certificate.
```

submariner/ipsec-cert-issuer.md
```diff
@@ -0,0 +1,114 @@
...
+    Certificate []byte
+
+    // The PEM encoded x509 certificate of the Certificate Authority.
+    CA []byte
+}
+```
+
+### CA CSR CRD
+
+The Proposed CA project to leverage is [cert-manager](https://cert-manager.io/docs/) Here is an
```
dfarrell07:
```suggestion
The Proposed CA project to leverage is [cert-manager](https://cert-manager.io/docs/). Here is an
```

dfarrell07 (commented):
Thanks, very exciting work! Some small suggestions inline.

maryamtahhan:
For posterity :) I've updated the EP based on the discussions we had and resolved any open issues.

I can no longer implement this EP (due to a shift in priorities), should I close this EP?

skitt (approved):
This has obviously had a lot of thought already put into it; the process described is similar to that used in other projects for similar use-cases, so it is implementable.

I reckon this is safe to merge as-is in “deferred” status.

tpantelis (approved):

PR #103, opened 2022-04-08
cable-driver: Overlay over IPSec transport mode support

maryamtahhan:
Proposal for IPSec transport mode support in Submariner


submariner/IPSec-transport-mode.md
```diff
@@ -0,0 +1,123 @@
...
+    ceIPSecMode: "transport",
+    ceIPSecOverlay: "vxlan",
+    ceIPSecPfsGroup: ["modp2048"]
+    ceIPSecEsp: ["aes_gcm-null"]
+```
+
+On the invocation of the `NewLibreswan()` function the configuration params
+are read and and establish if the request is for a tunnel or transport based
+connection. If the connection is transport based - then the `libreswan` driver
+will need to also drive the VXLAN or IPTun Cable Driver as well as the setup, and
```
maryamtahhan:
```suggestion
will need to also drive the VXLAN Cable Driver as well as the setup, and
```

submariner/IPSec-transport-mode.md
```diff
@@ -0,0 +1,123 @@
...
+teardown of the IPSec connections. The existing VXLAN/IPTun Configuration will be the
```
maryamtahhan:
```suggestion
teardown of the IPSec connections. The existing VXLAN Configuration will be the
```

submariner/IPSec-transport-mode.md
```diff
@@ -0,0 +1,123 @@
...
+[root@cluster2-worker]# ipsec whack --psk --encrypt --name submariner-cable-cluster1-172-18-0-11-0 --host 172.18.0.9 --clientproto udp/vxlan --to --host 172.18.0.11 --clientproto udp 
+[root@cluster2-worker]# ipsec whack --psk --encrypt --name submariner-cable-cluster1-172-18-0-11-1 --host 172.18.0.9 --clientproto udp --to --host 172.18.0.11 --clientproto udp/vxlan
+[root@cluster2-worker]# ipsec whack --route --name submariner-cable-cluster1-172-18-0-11-0
+[root@cluster2-worker]# ipsec whack --route --name submariner-cable-cluster1-172-18-0-11-1
+[root@cluster2-worker]# ipsec whack --initiate --asynchronous --name submariner-cable-cluster1-172-18-0-11-0
+[root@cluster2-worker]# ipsec whack --initiate --asynchronous --name submariner-cable-cluster1-172-18-0-11-1
+```
+<!-- markdownlint-enable line-length -->
+
+An example configuration for an IPinIP over IPSec Transport connection between two Submariner
```
maryamtahhan: Remove reference to IPTun

submariner/IPSec-transport-mode.md
```diff
@@ -0,0 +1,106 @@
+
+# Cable Driver Enhancement - IPSec Transport Mode
+
+## Summary
+
+Submariner today supports three cable drivers: VXLAN, IPSec (tunnel mode) and WireGuard. The
+goal of this proposal is to outline the enhancements needed for the Submariner
+Cable Driver to enable IPSec in transport mode.
```
mkolesnik: I'm no expert in IPSec, but from [this article](https://www.twingate.com/blog/ipsec-tunnel-mode/) it seems that tunnel mode is what's recommended for gateway-to-gateway communications, while transport mode is recommended for client-to-gateway.

> Transport mode is not without its flaws. It has poor compatibility with security gateways, as well as greater difficulty in implementing traversal NATs. For this reason, transport mode can’t be used in protected gateway-to-gateway configurations.

Maybe someone more knowledgeable than me in IPSec can chime in on this.

submariner/IPSec-transport-mode.md
```diff
@@ -0,0 +1,106 @@
...
+
+* **Performance**: Taking advantage of an overlay protocol wrapped in offloaded IPSec in transport mode can
+yield significant performance improvements. The existing IPSEC configuration performance on a pair of modern
+servers ~ 2Gbit maximum. Switching to IP GRE Tunnel protected by an offloaded transport mode IPSEC external
+“wrap” - 25Gbit. A difference of 12.5 times. Or in the case of VXLAN offloaded with an IPSec transport wrap - 9Gbit.
+* **Scale**: While it is theoretically possible to scale a system built around tunnel mode IPSEC to hundreds of nodes,
+it will be difficult to achieve reliability and stability, especially if there is a complex topology of tunnels. In
+order to achieve such scales, using other L2/L3 VPN tech combined with an optional external IPSEC protection provides
+a better approach, because it allows the use of dynamic routing protocols instead of complex static routing. Furthermore,
+it allows the use of traffic engineering and Site to Site QoS which is impossible in the case of IPSEC Tunnel mode.
```
mkolesnik, line 25: I'm not sure I understand what's the benefit here, in terms or routing.

Which routing are we talking about, the internet or the routing configuration on the gateways?

How would transport mode achieve better routing site to site?

submariner/IPSec-transport-mode.md
```diff
@@ -0,0 +1,106 @@
...
+goal of this proposal is to outline the enhancements needed for the Submariner
+Cable Driver to enable IPSec in transport mode.
+
+For Overlay-over-IPSec in transport mode with further configuration, the benefits fall into the
+following categories:
+
+* **Performance**: Taking advantage of an overlay protocol wrapped in offloaded IPSec in transport mode can
+yield significant performance improvements. The existing IPSEC configuration performance on a pair of modern
+servers ~ 2Gbit maximum. Switching to IP GRE Tunnel protected by an offloaded transport mode IPSEC external
+“wrap” - 25Gbit. A difference of 12.5 times. Or in the case of VXLAN offloaded with an IPSec transport wrap - 9Gbit.
```
mkolesnik, line 20: Is there source material to demonstrate these claims?

submariner/IPSec-transport-mode.md
```diff
@@ -0,0 +1,106 @@
...
+* **Scale**: While it is theoretically possible to scale a system built around tunnel mode IPSEC to hundreds of nodes,
+it will be difficult to achieve reliability and stability, especially if there is a complex topology of tunnels. In
+order to achieve such scales, using other L2/L3 VPN tech combined with an optional external IPSEC protection provides
+a better approach, because it allows the use of dynamic routing protocols instead of complex static routing. Furthermore,
+it allows the use of traffic engineering and Site to Site QoS which is impossible in the case of IPSEC Tunnel mode.
+* **Debuggability**: Tunnel mode IPSEC is a debugging nightmare (especially if it is offloaded in part or in full). There
+is no visibility of what goes into the VPN and once it has entered the tunnel it is encrypted and thus unsuitable for debugging.
```
mkolesnik, line 27: AFAICT this is stronger security though, and the proposal weakens security which might not be desirable?

submariner/IPSec-transport-mode.md
```diff
@@ -0,0 +1,106 @@
...
+This problem goes away if the VPN is using standard L2/L3 VPN technology which provides suitable interfaces on which the system
+administrator can snoop.
+
+## Proposal
+
+This enhancement proposes to extend the available IPSec offering in Submariner to also provide
```
mkolesnik, line 33: What's the use case this proposal is targeting?

submariner/IPSec-transport-mode.md
```diff
@@ -0,0 +1,106 @@
...
+    ceIPSecPfsGroup: ["modp2048"]
+    ceIPSecEsp: ["aes_gcm-null"]
+```
+
+On the invocation of the `NewLibreswan()` function the configuration params
+are read and and establish if the request is for a tunnel or transport based
+connection. If the connection is transport based - then the `libreswan` driver
+will need to also drive the VXLAN Cable Driver as well as the setup, and
+teardown of the IPSec connections. The existing VXLAN configuration will be the
+same.
```
mkolesnik, line 84: Personally I'd like to see examples of how this would look on the gateway, e.g. which interfaces are configured and how are they connected to each other.

Maybe a diagram would help illustrate the proposed deployment mode

submariner/IPSec-transport-mode.md
donaldh: In IPSec Tunnel mode packets look like [IP][ESP][IP][Data] where it's IPSec that does the encapsulation. 

In IPSec Transport mode packets look like [IP][ESP][Data], i.e. no encapsulation carried out by IPSec.  

The way we propose to use IPSec Transport mode is to layer it on top of an interchangeable encapsulation where Submariner controls the encapsulation. 

For IP in IP the resulting packet looks like [IP][ESP][IP][Data] which is the same as IPSec Tunnel mode. The benefits are:

* Submariner is in control of the encapsulation allowing us to choose encapsulations that work better
* IPSec Transport mode is better supported for hardware offload on the most common 10G NICs

Responding to the commonly reported flaws of IPSec Transport mode:

* Submariner does not require compatibility with security gateways. Submariner is talking to Submariner, not to a security gateway.
* Submariner is already handling the NAT traversal and is somewhat competing with IPSec Tunnel mode today.

The other way to think of this is that Submariner already uses encapsulation between gateways for e.g. VXLAN cables and already implements NAT traversal. This encapsulated traffic between Submariner gateways can be thought of as host to host traffic which is ideally suited to IPSec Transport mode encryption.

submariner/IPSec-transport-mode.md
```diff
@@ -0,0 +1,106 @@
...
+data:
+  ceCustomConfig: |
+    ceIPSecMode: "transport",
+    ceIPSecOverlay: "vxlan",
+    ceIPSecPfsGroup: ["modp2048"]
+    ceIPSecEsp: ["aes_gcm-null"]
+```
+
+On the invocation of the `NewLibreswan()` function the configuration params
+are read and and establish if the request is for a tunnel or transport based
```
sridhargaddam:
```suggestion
are read to establish if the request is for a tunnel or transport based
```

submariner/IPSec-transport-mode.md
```diff
@@ -0,0 +1,106 @@
...
+teardown of the IPSec connections. The existing VXLAN configuration will be the
+same.
+
+An example configuration (that would need to be done by the driver) for a VXLAN over
+IPSec Transport connection between two Submariner clusters gateways cluster1-worker
+(172.18.0.11) and cluster2-worker (172.18.0.9):
+
+<!-- markdownlint-disable line-length -->
+```console
+[root@cluster1-worker]# ipsec whack --psk --encrypt --name submariner-cable-cluster2-172-18-0-9-0 --host 172.18.0.11 --clientproto udp/vxlan --to --host 172.18.0.9 --clientproto udp
```
sridhargaddam: The VxLAN cable driver used in Submariner has interfaces named "vxlan-tunnel" and ip-address in the range "242.x.x.x". In these commands, I dont see any references to the vxlan-tunnel interface. Can you please elaborate how it works?

maryamtahhan: Sorry I should've been more clear... assuming a Submariner GW was deployed with the default VXLAN cable - this is the additional configuration that needs to be done to enable transport mode for the VXLAN traffic.

So all the existing VXLAN configuration is done as normal and this is in addition to it.

We are using Port Selectors (filters) with --clientproto udp/vxlan to filter vxlan traffic exiting/entering the host

I will document this...

maryamtahhan: and add a diagram

submariner/IPSec-transport-mode.md
dfarrell07: @mkolesnik Isn't it to scale Submariner's use cases to larger cluster sets at higher perf? The benefits above seem to kinda cover this.

submariner/IPSec-transport-mode.md
mkolesnik: > In IPSec Tunnel mode packets look like [IP][ESP][IP][Data] where it's IPSec that does the encapsulation.
> 
> In IPSec Transport mode packets look like [IP][ESP][Data], i.e. no encapsulation carried out by IPSec.
> 
> The way we propose to use IPSec Transport mode is to layer it on top of an interchangeable encapsulation where Submariner controls the encapsulation.
> 
> For IP in IP the resulting packet looks like [IP][ESP][IP][Data] which is the same as IPSec Tunnel mode. The benefits are:

Based on the previous statement, wouldn't it look like this then?
[IP (IPinIP)][IP (IPSec tunnel)][ESP][Data]

> 
>     * Submariner is in control of the encapsulation allowing us to choose encapsulations that work better
> 
>     * IPSec Transport mode is better supported for hardware offload on the most common 10G NICs

If indeed IPSec is inside the IP-in-IP tunnel, would that even work on offload?
I though the HW offload handles the outermost layer.

> 
> 
> Responding to the commonly reported flaws of IPSec Transport mode:
> 
>     * Submariner does not require compatibility with security gateways. Submariner is talking to Submariner, not to a security gateway.
> 
>     * Submariner is already handling the NAT traversal and is somewhat competing with IPSec Tunnel mode today.
> 
> 
> The other way to think of this is that Submariner already uses encapsulation between gateways for e.g. VXLAN cables and already implements NAT traversal. This encapsulated traffic between Submariner gateways can be thought of as host to host traffic which is ideally suited to IPSec Transport mode encryption.

So security gateway in this context is literally a firewall?
In that case I agree we don't work with that mode anyhow.

However, from my very brief and limited understanding the tunnel mode of IPSec is more secure but I might be wrong.

submariner/IPSec-transport-mode.md
mkolesnik: If it's a scale/perf issue then shouldn't we replace the default implementation rather than add another mode?

If, for our use cases, the two offer similar level of security but one is much slower than the other, why even keep the slow option?

I feel this proposal leaves much to the reader to understand the pros and cons, hence I asked for feedback from people more versed in IPSec.

submariner/IPSec-transport-mode.md
skitt: It allows someone with administrative privileges to “weaken security” to be able to debug issues. “Stronger security” always needs to reference specific scenarios; the scenarios in which the ability to debug reduces security are scenarios where the attacker is already on the wrong side of the fence.

submariner/IPSec-transport-mode.md
maryamtahhan: I've pushed a diagram there

submariner/IPSec-transport-mode.md
maryamtahhan: Yes, @kot-begemot-uk ran these benchmarks and has a report - he will be able to share on return from PTO

submariner/IPSec-transport-mode.md
maryamtahhan: the differences in IP tunnel mode and transport mode are shown [here](https://www.cisco.com/c/en/us/td/docs/security/vpn_modules/6342/vpn_cg/6342site3.html)
image

For IP in IP:
```
[Outer IP Header][Inner IP Header][Payload]
                 |-------------------------|
                   Considered the payload
 ```                         
 Another view: 
image

so for IPSec transport mode both the Inner IP header and the payload get encrypted as the data. Then the packet is constructed using the outer IP hdr

Screenshot 2022-04-13 at 11 22 31

**Comment by maryamtahhan** [comment]
surprisingly I don't think this linting problem is my fault :$
Screenshot 2022-04-13 at 11 56 22

**Comment by sridhargaddam** [comment]
> surprisingly I don't think this linting problem is my fault :$ Screenshot 2022-04-13 at 11 56 22

Yes, it seems to be broken as discussed here - https://kubernetes.slack.com/archives/C010RJV694M/p1649844788372089

File: submariner/IPSec-transport-mode.md [hunk 12]
```diff
@@ -0,0 +1,114 @@
+
+# Cable Driver Enhancement - IPSec Transport Mode
+
+## Summary
+
+Submariner today supports three cable drivers: VXLAN, IPSec (tunnel mode) and WireGuard. The
+goal of this proposal is to outline the enhancements needed for the Submariner
+Cable Driver to enable IPSec in transport mode.
+
+For Overlay-over-IPSec in transport mode with further configuration, the benefits fall into the
```
yboaron, line 14: Does ESP HW offload is relevant only for transport mode?  I couldn't find a reference to ipsec mode in HW offload RHEL doc [1] 

What about the pod <--> GW segment? do we need to use some kind fast datapath tech here?  I assume to reach the numbers specified on the EP we should use fast datapath technology  also for this segment

[1]
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9-beta/html/securing_networks/configuring-a-vpn-with-ipsec_securing-networks#proc_configuring-automatic-detection-and-usage-of-esp-hardware-offload-to-accelerate-an-ipsec-connection_configuring-a-vpn-with-ipsec

submariner/IPSec-transport-mode.md
```diff
@@ -0,0 +1,114 @@
...
+* **Scale**: While it is theoretically possible to scale a system built around tunnel mode IPSEC to hundreds of nodes,
+it will be difficult to achieve reliability and stability, especially if there is a complex topology of tunnels. In
+order to achieve such scales, using other L2/L3 VPN tech combined with an optional external IPSEC protection provides
+a better approach, because it allows the use of dynamic routing protocols instead of complex static routing. Furthermore,
+it allows the use of traffic engineering and Site to Site QoS which is impossible in the case of IPSEC Tunnel mode.
+* **Debuggability**: Tunnel mode IPSEC is a debugging nightmare (especially if it is offloaded in part or in full). There
+is no visibility of what goes into the VPN and once it has entered the tunnel it is encrypted and thus unsuitable for debugging.
+This problem goes away if the VPN is using standard L2/L3 VPN technology which provides suitable interfaces on which the system
+administrator can snoop.
+
```
yboaron, line 30: I think we should cover also the following topics in the context of this EP:

- SM deployment 
	e.g: Do we need to 'tell' K8S control plane that our GW engine should be scheduled on a node with HW offload capabilities? if yes, how we should do that?

  - GW HA
	   Does it affect GW HA ?  how we should mark node as a candidate for GW deployment?

- Datapth
  Do we need to update SM ipsec NAT traversal 
  handling [1]? 
  Do we need to update/configure something at host's network stack/kernel 
 Any effects on datapath metrics? 

[1]
https://submariner.io/operations/nat-traversal/

submariner/IPSec-transport-mode.md
yboaron: Thanks for the explanations @maryamtahhan  & @donaldh .

I'm not sure about the added value of VxLAN encapsulated in ipsec in the context of SM. I don't see how secured L2 VPN  can be useful for SM.

submariner/IPSec-transport-mode.md
sridhargaddam: > If it's a scale/perf issue then shouldn't we replace the default implementation rather than add another mode?

We can do that in future once the implementation becomes stable and mature.

submariner/IPSec-transport-mode.md
mkolesnik: OK but from my understanding, and the material linked by Maryam, it seems security is indeed weaker in "transport mode"

> In IPSec transport mode, only the IP payload is encrypted, and the original IP headers are left intact. (See [Figure 3-6](https://www.cisco.com/c/en/us/td/docs/security/vpn_modules/6342/vpn_cg/6342site3.html#wp1036240).) This mode has the advantage of adding only a few bytes to each packet. It also allows devices on the public network to see the final source and destination of the packet. With this capability, you can enable special processing in the intermediate network based on the information in the IP header. However, the Layer 4 header will be encrypted, limiting the examination of the packet. Unfortunately, by passing the IP header in the clear, transport mode allows an attacker to perform some traffic analysis.

So doesn't seem to me like the attacker needs to be "on the wrong side of the fence" to exploit this, unless I'm misunderstanding the documentation in this area.

submariner/IPSec-transport-mode.md
mkolesnik: I'm not saying we should immediately replace the current connectivity model with this, but if it suits our needs while offering better performance then I'd like to see a proposal to deprecate the current model.

I don't think we need to support two very similar tools for the same task, if one has a clear advantage over the other (as the EP suggests)

submariner/IPSec-transport-mode.md
maryamtahhan: I think that the orchestration/deployment of a HW offloaded IPSec Submariner instance is a separate issue to the EP - the EP simply wants to add a new cable type for Overlay over IPSec in transport mode. Extending this EP beyond this is scope creep IMHO.

More info can be found here: https://libreswan.org/wiki/Cryptographic_Acceleration 
But from a first pass it looks like Libreswan is capable of automatically picking up on the acceleration capabilities and taking advantage of them when they are available

Re the NAT traversal:
 There are no changes/updates needed for NAT traversal for the Overlay over IPSec implementation we are proposing.  From the Submariner perspective the packets will appear the exact same as they do for tunnel mode. (the difference will be that we didn't encrypt an additional 20bytes (IP HDR) per packet)
 
IPSec ESP packets in transport mode can be encapped with UDP as per [RFC 3948]( https://datatracker.ietf.org/doc/html/rfc3948#section-3.2) and using the ipsec whack `--forceencaps` param in combination with `--ikeport 4500` can be used to accomplish this, you can see the below (partial) captures from a GW I used configured with VXLAN over IPSEC in transport mode:
Screenshot 2022-04-14 at 12 37 08



Screenshot 2022-04-14 at 12 36 36

submariner/IPSec-transport-mode.md
mkolesnik: Unless the proposal is that first the traffic would be encapsulated (eg VXLAN) and then sent through the IPSec tunnel, which I guess achieves the goal of security by obfuscation as the original IP header is encapsulated and the VXLAN IP header is the one sent in the plain?

It wasn't clear to me how is this going to work

submariner/IPSec-transport-mode.md
yboaron: 1. According to EP summary one of the main benefits of this approach is performance, if the performance improvement depends on HW offload, I think we should at least specify what are the steps to consume the offload service.

I'm not saying we should have full design of the orchestration here, just to understand the bigger picture,  things like what is needed to consume the offload service (k8s scheduling? etc) and availability of offload service in different platforms.

2. Regarding NAT just wanted to make sure that the current nat discovery approach can be used with overlay + IPSec transport.

3. Not an IPSec expert, but I think we'll encrypt the same number of bytes in the current (tunnel)  mode and in the proposed mode (IPinIP overlay + transport IPSec) . right?

submariner/IPSec-transport-mode.md
```diff
@@ -0,0 +1,114 @@
...
+enable Overlay-over-IPSec in transport mode.
+
+For Overlay-over-IPSec in transport mode with further configuration, the benefits fall into the
+following categories:
+
+* **Performance**: Taking advantage of an overlay protocol wrapped in offloaded IPSec in transport mode can
+yield significant performance improvements. The existing IPSEC configuration performance on a pair of modern
+servers ~ 2Gbit maximum. Switching to IP GRE Tunnel protected by an offloaded transport mode IPSEC external
+“wrap” - 25Gbit. A difference of 12.5 times. Or in the case of VXLAN offloaded with an IPSec transport wrap - 9Gbit.
+* **Scale**: While it is theoretically possible to scale a system built around tunnel mode IPSEC to hundreds of nodes,
```
yboaron, line 21: Regarding scale, in order to use dynamic routing and QoS services we need to develop a full (complicated ) SW stack for these features.
Having the overlay layer is just one piece of the puzzle. IMO it isn't a significant advantage here.

submariner/IPSec-transport-mode.md
maryamtahhan: From previous Experience - consuming HW offloads that aren't detected automatically through the software/app we are using generally involves the use of device plugins (to detect and advertise the available HW devices) that are vendor specific and tying them to resource pools that get dished out to pods that request them. This adds a high level of complexity to the system and the need for Submariner to lose its HW agnosticism. 

Instead, I would suggest the following: the NICs that support HW offload capabilities for Libreswan are pretty standard and available in all cloud environments. What we could document is to encourage the 'selection' of one these NICs for a node where one intends to deploy the Submariner GW and leverage much simpler technique using labels: a node label for the node that has the right NIC and for the GW pod to use a label selector (for that node label) to ensure that the GW lands on the right platform. Then rely on Libreswan to autodetect the HW features and do its thing. This uses a very simple approach - which keeps submariner away from the vendor specific approach.
(I can easily document this).

I should've clarified that I was excluding IPinIP from the discussion and focusing as VXLAN over IPSec transport mode as this is what's available in Submariner today. But yes for IPinIP over transport mode vs pure tunnel mode you will have the same number of bytes yes. 

On the NAT - there will be no changes to the current nat discovery approach for overlay over IPSec transport.

submariner/IPSec-transport-mode.md
maryamtahhan: The proposal is for an overlay over IPSec in transport mode - I've pushed a couple of updates to call that out specifically.

submariner/IPSec-transport-mode.md
kot-begemot-uk: I ran a set of tests on ixgbe (most popular 10G adapter - > 50% market) and Melanox. You can run them yourself: https://github.com/kot-begemot-uk/tunnel-tests

Tunnel mode offload is totally disabled in the ixgbe upstream driver at present (I can dig out the exact commit, it is somewhere in 5.x and got backported to all -stable series). It can be enabled on TX, but not on RX - I submitted a patch to the kernel, but it got ignored. The maintainer decided that it is safer just to keep it off completely. Even with TX-only enabled you get at most 5G, usually less.

All in all, the performance in tunnel mode on most cards likely to be found in an SME and basic bare metal hosting is guaranteed to be abysmal. 

Compared to that transport mode works both on ixgbe and Melanox for supported adapters. As a result you get anything between 6-9G and line rates where offload works for all encaps levels. Up to 20 times performance difference in some cases.

submariner/IPSec-transport-mode.md
kot-begemot-uk: It is not an advantage. It is a required prerequisite. Dynamic routing and tunnel mode are two incompatible paradigms.

submariner/IPSec-transport-mode.md
yboaron: This seems like an advantage for transport mode and I guess we can benefit from HW offload in BM K8s (k8s node == BM server) deployments,

but is HW offload also applicable in public clouds 'virtual machines' and OSP VMs?

IIRC, to consume HW offload capabilities the 'vm' needs to be connected 'directly' (for example: sriov) to the NIC.

submariner/IPSec-transport-mode.md
yboaron: I'm not sure how much L2/L3 VPN capabilities (like dynamic routing) are needed in SM context.
But, yeah using ipsec in transport mode enables the usage of L2/L3 VPN capabilities.

submariner/IPSec-transport-mode.md
mkolesnik: @kot-begemot-uk thanks for sharing, Can this supporting evidence be added in some form to the EP?
Either as a sub section or if you have some place else then a link to the results and how to reproduce them would be good.

submariner/IPSec-transport-mode.md
```diff
@@ -0,0 +1,132 @@
...
+_"libreswan as of version 3.23 supports the new cryptographic hardware offload as implemented by Linux 4.11 and up using the
+native (XFRM) IPsec stack. Libreswan autodetects supports for any hardware supporting this crypto offload API"_ [1].
+
+The [NICs](https://libreswan.org/wiki/Cryptographic_Acceleration#Supported_hardware) that support Hardware offload capabilities
+for Libreswan are pretty standard and available in all cloud environments. Submariner encourages the 'selection' of one these
+NICs for a node where one intends to deploy the Submariner gateway and leverage the use of labels to schedule the pod to the
+appropriate node: a node label for the node that has the right NIC and for the GW pod to use a label selector (for that node label)
+to ensure that the gateway lands on the right platform. Then rely on Libreswan to autodetect the HW features and take advantage of them.
+
+> **_NOTE:_** The IPSec cable driver will need to be modified to run `ipsec _stackmanager start` to load the appropriate
```
yboaron, line 131: Thanks for adding the HW offload section  @maryamtahhan .

IIUC,` ipsec _stackmanager start` should run at node level and should affect all ipsec consumers on this node.

Maybe it makes sense to run this operation in some other component outside SM ? because SM GW pod may not be the only ipsec HW offload consumer  on this node

So, to allow ipsec HW offload :

A. some other component should run all the bits to enable ipsec HW offload and set the relevant label for the appropriate nodes

B. Guide K8S to schedule SM gateway pods on these nodes

maryamtahhan:
Closing out this EP with a link to the reference implementation (vxlan over IPSec tunnel in transport mode):  for [Submariner](https://github.com/maryamtahhan/submariner/tree/feat_ipsec_transport), [shipyard](https://github.com/maryamtahhan/shipyard/tree/feat_ipsec_transport)

These changes introduce a new cable driver called vxlan-libreswan, the reason a new cable driver approach was adopted was that in the future if support for tunnel mode were to be removed, the associated files could also be removed. The reference implementation doesn't include the enablement for HW offload through the use of `ipsec _stackmanager start` it's thought that this is a separate step that will be carried out by the operator or system admin.

tpantelis (approved):

dfarrell07:
This is marked as Deferred, so I think we want to wrap up the conversation and merge it. @yboaron can you summarize your questions and we can get an Owner (@tpantelis volunteered) to push them to add to the doc before merging?

yboaron:
A. Since one of the advantages of transport mode is performance, it is worth mentioning that better performance is achieved mainly by HW offload and HW offload is only relevant when SM runs on a node connected to a physical NIC (eg: k8s node == baremetal server).

B. The proposal allows flexibility by letting SM control the overlay over IPSec in transport mode, so we have 
[IP][ESP][Overlay].

These are all cool, but I think we need to understand:
* Which SM use cases might benefit from having a flexible overlay
* What are the required changes in SM components

For example, to configure inter-cluster routes dynamically instead of the current method (static routes using the information from the broker), SM needs to configure an overlay (ie VxLAN), and a dynamic routing instance needs to run in each cluster.

This information will help us understand the need for this feature and the required efforts.


cc @tpantelis

tpantelis:
@yboaron I added your comments in a TODO section. Please review.

yboaron (approved):

PR #110, opened 2022-04-14
cable driver policy

maryamtahhan:
Breaking out cable driver policy from finer grained connectivity EP. This EP specifies how to configure different cable drivers connecting clusters in a ClusterSet


submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,250 @@
+# Cable Driver Policy Support
+
+## Summary
+
+Submariner currently offers three cable drivers: VXLAN, IPSec or wireguard. However only one of these cable drivers
+is enabled across all the clusters in a ClusterSet; That is all clusters that join a broker must use the same cable
+driver. The goal of this proposal is to extend the current model to support the selection of different cable types
+between connected clusters. For example, an IPSec cable between public and private clouds but a VXLAN cable between
+the clusters in the same cloud.
```
tpantelis: I think this is redundant with L11-14.

```suggestion
between connected clusters.
```

submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,250 @@
...
+cable type.
+
+This enhancement will propose a single API for cable driver policy support using custom resource definition (CRDs).
+
+### Cable driver policy API
+
+This following CRD is proposed:
+
+```Go
+type CablePolicies struct {
```
tpantelis: I'd prefer a more general, user-friendly CRD name. 

```suggestion
type ClusterConnectionPolicies struct {
```

submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,250 @@
...
+    metav1.TypeMeta `json:",inline"`
+
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+
+    ConnectionPolicies []CablePolicySpec `json:"spec"`
```
tpantelis:
```suggestion
    Spec ClusterConnectionPoliciesSpec `json:"spec"`
    ...
    type ClusterConnectionPoliciesSpec struct {
        Policies []ClusterConnectionPolicy `json:"policies,omitempty"`
    }
```

submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,250 @@
...
+- Search for a policy whose left and right label selectors match the corresponding Cluster resources of the local
+  and remote Endpoints.
+
+If any matching policy is found, the clusters are connected using the cable driver specified in the cable driver policy.
+
+If no matching policy is found, the `default` cable driver policy is used. The `default` policy can be auto-created at broker
+deploy time.
+
+> **_NOTE:_** In the case of multiple labels being specified, the selector will need to match all the labels.
+> Equality-based requirements will be supported. For e.g. labels such as `env: !production` will match any clusters
```
skitt: Is this supposed to say

```suggestion
> Inequality-based requirements will be supported. For e.g. labels such as `env: !production` will match any clusters
```

instead?

submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,250 @@
...
+> that aren't labelled with `env: production`.
+
+The initial `CablePolicies` resources would be created (at startup time) and maintained on the broker cluster
+and synced to each participating cluster. Policies can subsequently be created/deleted on a cluster and will be
+automatically exported to the broker. A controller to reconcile updates to policies will need to be implemented in
+Submariner.
```
skitt: Why do we want to allow updates on participating clusters? Presumably the clusterset administrator has access to the broker.

submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,250 @@
...
+deploy CablePolicy CRDs, as well as the extension of Submariner configuration YAML to include parameters that
+support cable driver configuration.
+
+> **_NOTE:_** The default policy can be deduced from the existing Submariner configuration if it is not
+explicitly specified.
+
+The `CablePolicies` can be created/managed after Broker deployment through subctl using a command such as:
+
+```cmd
+subctl cable-policy add | list | delete
```
skitt: Ah, this would answer my earlier question, unless we require `subctl` to be given a kubeconfig pointing to the broker. I reckon it’s worth requiring that instead of having to develop a synchronising controller with conflict resolution.

submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,250 @@
...
+| `--kubecontext` `<string>`   | Kubeconfig context to use
+
+#### Connection Policy Examples
+
+- we want to specify a default policy that uses IPsec as the default cable driver.
+
+```yaml
+kind: CablePolicies
+metadata:
+  name: default
```
skitt, line 244: Which namespace does all this go in? (If we require broker access, the broker namespace, but we should mention this either way.)

submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,250 @@
...
+### Labelling clusters
+
+Clusters are defined as Custom Resources and can be labeled in the CRD. A new parameter to label clusters at
+`subctl join` time could also be added for usability purposes.
+
+### Policy management with subctl
+
+The `CablePolicies` can be created at the Broker deployment through the extension of the Submariner-Operator to
+deploy CablePolicy CRDs, as well as the extension of Submariner configuration YAML to include parameters that
+support cable driver configuration.
```
tpantelis: I don't think the `Submariner` CRD would be the right place for cluster set-wide config. We could extend `deploy-broker ` and the `Broker` CRD to deploy policy CRD(s) but I'm not sure that would buy us much since they can simply be deployed directly via **kubectl**. Either way I think this is kind of "sugar on top" that we could consider later as an enhancement. The important thing for this EP is to define the policy CRD and the associated behavior.

submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,250 @@
...
+The proposed approach for additional Submariner cable driver configuration parameters is:
+
+- To continue to use the existing environment variables injected into the Gateway Pod (with the exception of PSKs/secrets).
+  These environment variables will come from the `Submariner.io_Submariners.yaml`.
+- To extend `Submariner.io_Submariners.yaml` with some new global parameters for multiple cable driver configurations
+  in a ClusterSet.
+- To use a volume mounted ConfigMap for more detailed cable driver configurations (as these may vary and should be shared across
+  gateway instances in the ClusterSet).
+- To use mounted secrets for parameters that need to be protected, e.g. the Pre-Shared Key (PSK) and to abandon the use of
+  environment variables for secrets.
```
skitt: See https://github.com/submariner-io/submariner-operator/issues/1869 and https://github.com/skitt/submariner-operator/pull/24

skitt (changes requested):

submariner/cable-driver-policy.md
maryamtahhan: Yeah - I will update

submariner/cable-driver-policy.md
tpantelis: I'm not sure I follow all of this. Isn't the cable driver configuration specified via the new policy CRD outlined above?

submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,250 @@
...
+are listed in the following table:
+
+<!-- markdownlint-disable line-length -->
+| Parameter name              | Description
+|:----------------------------|:----------------------------------------------------------------------------|
+| cableDrivers                | An array of strings listing the drivers to configure as separate cables.
+| combinedCableDrivers        | An array of strings listing the drivers to configure as a combined cable e.g vxlan + IPSEC in transport mode
+| cableDriverCustomConfig     | The name of the ConfigMap for custom driver configuration.
+| cableDriverCustomConfigPath | The path to the mounted ConfigMap volume.
+<!-- markdownlint-enable line-length -->
```
tpantelis: I'm not sure I follow this either. Hoe does this differ from or relate to the policy CRD?

submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,250 @@
...
+```Go
+type CablePolicies struct {
+    metav1.TypeMeta `json:",inline"`
+
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+
+    ConnectionPolicies []CablePolicySpec `json:"spec"`
+}
+
+type CablePolicySpec struct {
```
tpantelis:
```suggestion
type ClusterConnectionPolicy struct {
```

submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,250 @@
...
+    // LeftClusterSelector identifies the Cluster resources representing the clusters on one end of a connection. An empty
+    // selector indicates wildcard, ie matches any cluster.
+    LeftClusterSelector metav1.LabelSelector `json:"leftClusterSelector,omitempty"`
+
+    // RightClusterSelector identifies the Cluster resources representing the clusters on the other end of a connection.
+    // An empty selector indicates wildcard, ie matches any cluster.
+    RightClusterSelector metav1.LabelSelector `json:"rightClusterSelector,omitempty"`
+
+    // Name of the cable driver implementation to use for this connection.
+    CableDriver string `json:"name"`
```
tpantelis:
```suggestion
    CableDriverName string `json:"cableDriverName"`
```

submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,250 @@
...
+
+    // Options specifies a ConfigMap to use for additional configuration parameters for the cable driver.
+    CableOptions *v1.ConfigMap `json:"options,omitempty"`
```
tpantelis: I don't think we need to/should embed an entire `ConfigMap` CR. Wouldn't just `map[string]string` suffice here?

```suggestion
    CableDriverOptions *v1.ConfigMap `json:"cableDriverOptions,omitempty"`
```

tpantelis (changes requested):

submariner/cable-driver-policy.md
tpantelis: I imagine since policies are cluster set-wide that they should be created on the broker in the broker NS and synced to the member clusters. Also we can certainly consider new `subctl` commands but, as mentioned in another comment, I think we should focus on the policy CRD and behavior for this EP.

submariner/cable-driver-policy.md
maryamtahhan: Ok - I will update it

submariner/cable-driver-policy.md
maryamtahhan: Sure - makes sense

submariner/cable-driver-policy.md
maryamtahhan: oops - yes

submariner/cable-driver-policy.md
maryamtahhan: Ok - I will remove the additional subctl support

submariner/cable-driver-policy.md
maryamtahhan: So the CRD will tell you what cable to use to connect a set of clusters and give some additional config for them if that's needed.

Then there's the actual practical configuration that needs to be passed to the GW for it to setup the cable drivers on startup... so things like what drivers the cable should enable as well as how to configure them... 

We can either try to process the connection CRDs on the GW at startup - or have a simple configuration field that says what set of drivers are needed by the Gateways...

submariner/cable-driver-policy.md
maryamtahhan: So these were supposed to be new global params in the submariner CRD to simplify the start up process for a GW.

I guess the GW does need to eventually process the connection CRDs to figure out how it should connect to each cluster - but I was trying to keep the GW startup simple. Aka a have a field that tells the GW I need you to configure both the ipsec driver and the vxlan driver.

submariner/cable-driver-policy.md
maryamtahhan: Brilliant. I will remove this from the EP

submariner/cable-driver-policy.md
maryamtahhan: yeah - I will update to include the broker namespace

submariner/cable-driver-policy.md
maryamtahhan: I will update

submariner/cable-driver-policy.md
maryamtahhan: I will update

submariner/cable-driver-policy.md
maryamtahhan: Yeah - a map will do.

the thinking behind the config map was that potentially we could explore updates to driver configurations without having to restart the GW pod. But for simplicity sake and since we are removing the subctl support for extensions to the policies this can be removed...

submariner/cable-driver-policy.md
maryamtahhan: Yeah - I will remove this for now as I'm removing subctl support for later updates to the policies

submariner/cable-driver-policy.md
maryamtahhan: i will remove these

maryamtahhan:
This still needs a little TLC. I will update when I get a chance.

submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,165 @@
...
+    metav1.TypeMeta `json:",inline"`
+
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+
+    ClusterConnectionPoliciesSpec `json:"spec"`
+}
+
+type ClusterConnectionPoliciesSpec struct {
+ 
+    Policies []ClusterConnectionPolicy `json:"policies,omitempty"`
```
maryamtahhan: Even Though I updated this to a set of policies based on previous comments - this will result in one long CRD, this is not inline with what the original intention was - to be able to define individual policies rather than a set of them in a single CR, with a particular policy being identifiable as the default one...

so my suggestion would be to change this to:

```Go
type ClusterConnectionPolicy struct {
    metav1.TypeMeta `json:",inline"`

    metav1.ObjectMeta `json:"metadata,omitempty"`

    ClusterConnectionPolicySpec `json:"spec"`
}

type ClusterConnectionPolicySpec struct {
    // LeftClusterSelector identifies the Cluster resources representing the clusters on one end of a connection. An empty
    // selector indicates wildcard, ie matches any cluster.
    LeftClusterSelector metav1.LabelSelector `json:"leftClusterSelector,omitempty"`

    // RightClusterSelector identifies the Cluster resources representing the clusters on the other end of a connection.
    // An empty selector indicates wildcard, ie matches any cluster.
    RightClusterSelector metav1.LabelSelector `json:"rightClusterSelector,omitempty"`

    // Name of the cable driver implementation to use for this connection.
    CableDriverName string `json:"CableDriverName"`

    // CableOptions specifies a additional configuration parameters for the cable driver.
    CableOptions map[string]string `json:"CableOptions,omitempty"`
}
```

submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,165 @@
...
+    // CableOptions specifies a additional configuration parameters for the cable driver.
+    CableOptions map[string]string `json:"CableOptions,omitempty"`
+}
+```
+
+The `ClusterConnectionPolicies` CR defines the criteria for selecting which cable driver to use to connect cluster pairs (or subsets)
+and their associated (optional) configuration options. To resolve the connectivity policy for a remote cluster's Endpoint,
+Submariner would:
+
+- Search for a policy whose left and right label selectors match the corresponding Cluster resources of the local
```
maryamtahhan: The suggestion here would be that the gw at startup time pulls in the connection policy CRs, and builds a local connection map[string]string , where the key is a cluster identifier or potentially even an endpoint identifier and the value is cable driver. On the first pass the default policy is applied to all values in the map. 

The next thing the GW should do is parse the rest of the policies - where it finds a policy whose left selectors match it, it establishes the list of clusters that match the right selectors and updates the connectivity map with what's specified in the policy.

When the time comes to actually install the cable - the connectivity map is consulted...

maryamtahhan: rather than doing constant lookups of the CRs

submariner/cable-driver-policy.md
tpantelis: All the info needed to connect clusters should be in the CRDs so I was imagining the tunnel controller in the GW would set up cable drivers based on that. It could read all the policies on startup and start the referenced drivers prior to processing `Endpoints` but it's going to have to react to policy changes as well so it's going to have to be able to spin up cable driver instances on the fly anyway so I was thinking we may as well handle startup in the same manner. So, on startup, no cable drivers are started. When an `Endpoint` is received, it consults the policies to determine which driver to use and spins it up if not already started.

submariner/cable-driver-policy.md
tpantelis: Yes - I meant to talk about this before. There's been much [discussion](https://github.com/submariner-io/enhancements/pull/4#discussion_r576234584) about whether to use one singleton CRD with a list of policies or multiple policy CRDs. There's pros and cons to both and I go back and forth but I think the latter had more consensus. So your suggestion above looks correct. 

Either way tho, even with a specially-defined default/fallback policy, I think we will need a priority field to resolved ambiguities.

submariner/cable-driver-policy.md
tpantelis: We can build and maintain a cache for the local `Cluster` based on the current view of the connection policies. This would be dynamic and react to changes to the policies and the local `Cluster` labels. I haven't thought about the matching algorithm in detail but I think we could build a priority-ordered list of policies where the local `Cluster` matches the left label selector and another list where the the local `Cluster` matches the right selector. Then when processing a remote `Endpoint`, we walk the left selector list and, for each, check if the remote `Cluster` matches the right selector. If so then it's a candidate. Then do similar for the right selector list. If there's matches from both lists then take the higher priority. Or we could use one list and, for each entry, record which label selector to use for matching with remote clusters. Either way, if no match is found then use the cable driver from the default policy. Anyway, that's my initial thoughts.

submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,165 @@
...
+
+    // RightClusterSelector identifies the Cluster resources representing the clusters on the other end of a connection.
+    // An empty selector indicates wildcard, ie matches any cluster.
+    RightClusterSelector metav1.LabelSelector `json:"rightClusterSelector,omitempty"`
+
+    // Name of the cable driver implementation to use for this connection.
+    CableDriverName string `json:"CableDriverName"`
+
+    // CableOptions specifies a additional configuration parameters for the cable driver.
+    CableOptions map[string]string `json:"CableOptions,omitempty"`
```
tpantelis: @maryamtahhan I was wondering how you envision this being used. This is included in the policy spec so that would imply that we could have 2 or more policies referencing the same driver name but different options, eg

```yaml
  leftClusterSelector:
    env: production
  rightClusterSelector:
    env: production
  cableDriverName:
    name:  "ipsec"
  cableDriverOptions:
    foo: 123

  leftClusterSelector:
    env: testing
  rightClusterSelector:
    env: testing
  cableDriverName:
    name:  "ipsec"
  cableDriverOptions:
    foo: 456
```

Would we then spin up 2 instances of the driver, each with its own config? Is it possible to have 2 instances of the same driver running in a GW? 

Or should there only be one set of config options for a driver? If so then it seems the options shouldn't be defined here but rather in a separate `ConfigMap` (one per cable driver).

tpantelis (changes requested):

submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,165 @@
...
+default/fallback driver.
+
+The specification additions to `Submariner.io_Submariners.yaml` are shown below:
+
+```yaml
+cableDrivers:
+  items:
+    type: string
+  type: array
+```
```
maryamtahhan: Update diagram to full mesh...

submariner/cable-driver-policy.md
maryamtahhan: For the VXLAN driver today this won't be possible without creating an additional vxlan interface with the new configuration (as it's not using the lightweight tunnel implementation [ip route encap vxlan ...])...
For IPSec in theory - this could be easily done to support different encryption algos... as these are simply whack params...

But as both scenarios cannot be catered for today - it's probably best to leave out the cableDriverOptions and as you say cater for the cable driver configuration via ConfigMap - one config map should be fine across all the drivers with a naming convention for cable configuration parameters `ceCableNameParam` or something along those lines?

```yaml
kind: ConfigMap
metadata:
  name: cabledriver-config
  namespace: submariner-operator
data:
  ceConfig: |
  ceVXLANVni: 3220
  ceIPSecDebug: false
  ceIPSecForceUDPEncaps: false
  ceIPSecIKEPort: false
  ceIPSecNATTPort: 0
  ceIPSecPSK: "test"
  ceIPSecPreferredServer: false
  ceIPSecPermitEncAlgs: ["a", "b"]
  ceIPSecDenyEncAlgs: ["c", "d"]
```

then we can mount the single ConfigMap as a volume on the GW pod

submariner/cable-driver-policy.md
maryamtahhan: I'm not sure about priority - if I look at other policy examples I don't typically see a priority field... maybe a status field is more appropriate? 

If there's ambiguity - to avoid lack of connectivity between clusters the controller could mark the status as error on the policy and the GW would fall back to the default driver to connect clusters till the error is resolved?

The GW would only apply status `approved` policies?

maryamtahhan:
```Go
type ClusterConnectionPolicy struct {
    metav1.TypeMeta `json:",inline"`

    metav1.ObjectMeta `json:"metadata,omitempty"`

    ClusterConnectionPolicySpec `json:"spec"`

    // Status is the current state of the ClusterConnectionPolicy.
    Status ClusterConnectionPolicyStatus `json:"status,omitempty"`
}

type ClusterConnectionPolicySpec struct {
    // LeftClusterSelector identifies the Cluster resources representing the clusters on one end of a connection. An empty
    // selector indicates wildcard, ie matches any cluster.
    LeftClusterSelector metav1.LabelSelector `json:"leftClusterSelector,omitempty"`

    // RightClusterSelector identifies the Cluster resources representing the clusters on the other end of a connection.
    // An empty selector indicates wildcard, ie matches any cluster.
    RightClusterSelector metav1.LabelSelector `json:"rightClusterSelector,omitempty"`

    // Name of the cable driver implementation to use for this connection.
    CableDriverName string `json:"CableDriverName"`
}

// ClusterConnectionPolicyConditionType is the type for status conditions on
// a ClusterConnectionPolicy This type should be used with the
// ClusterConnectionPolicyStatus.Conditions field.
type ClusterConnectionPolicyConditionType string

const (
  // ClusterConnectionPolicyConditionStatusApproved represents status of a ClusterConnectionPolicy that is approved
  // the ClusterConnectionPolicy provider and will be implemented in the cluster
  ClusterConnectionPolicyConditionStatusApproved ClusterConnectionPolicyConditionType = "Approved"

  // ClusterConnectionPolicyConditionStatusFailure represents status of a ClusterConnectionPolicy that is not approved
  // ClusterConnectionPolicy provider and will not be implemented in the cluster
  ClusterConnectionPolicyConditionStatusFailure ClusterConnectionPolicyConditionType = "Failure"
)

// ClusterConnectionPolicyConditionReason defines the set of reasons that explain why a
// particular ClusterConnectionPolicy condition type has been raised.
type ClusterConnectionPolicyConditionReason string

const (
  // ClusterConnectionPolicyConditionReasonPolicyError represents a reason where the ClusterConnectionPolicy may not have been
  // implemented in the cluster due to an issue detected by the ClusterConnectionPolicy provider with another policy
  ClusterConnectionPolicyConditionReasonPolicyError ClusterConnectionPolicyConditionReason = "PolicyError"
)

// ClusterConnectionPolicyStatus describe the current state of the ClusterConnectionPolicy.
type ClusterConnectionPolicyStatus struct {
  // Conditions holds an array of metav1.Condition that describe the state of the ClusterConnectionPolicy.
  Conditions []metav1.Condition `json:"conditions,omitempty"`
}
```

tpantelis: yeah we could take the fail fast approach to conflicts/ambiguity. I'm not sure tho we should automatically fallback to the default policy. Connectivity may still **seem** to be working to the user but they might not be aware that their intended policy didn't get applied. 


+1 on adding status. However it wouldn't be a singular status but one per remote cluster, ie the policy may be successfully applied for one cluster but not another.  Also the `ClusterConnectionPolicy` CRD is synced from the broker so we'd have to think about those dynamics, ie do we update the local synced copy or create another local CRD to store the status.

submariner/cable-driver-policy.md
tpantelis: Perhaps we should just remove the cable driver configuration from this proposal altogether and leave it as a possible future enhancement.

submariner/cable-driver-policy.md
maryamtahhan: The intention for the status in `ClusterConnectionPolicy` to reflect the status of the connection policy rather than its application. So it's either approved or declined (I will change the terminology from failed)... We could do something along the lines of only approved policies will be synced to clusters from the broker and a GW will only apply approved policies...

The `ConnectionStatus` is already available per cluster pair through the GW status which can be retrieved through subctl I'm not sure if we want to duplicate that.

```go
type GatewayStatus struct {
	Version       string       `json:"version"`
	HAStatus      HAStatus     `json:"haStatus"`
	LocalEndpoint EndpointSpec `json:"localEndpoint"`
	StatusFailure string       `json:"statusFailure"`
	Connections   []Connection `json:"connections"`
}

type Connection struct {
	Status        ConnectionStatus `json:"status"`
	StatusMessage string           `json:"statusMessage"`
	Endpoint      EndpointSpec     `json:"endpoint"`
	UsingIP       string           `json:"usingIP,omitempty"`
	UsingNAT      bool             `json:"usingNAT,omitempty"`
	// +optional
	LatencyRTT *LatencyRTTSpec `json:"latencyRTT,omitempty"`
}

type ConnectionStatus string

const (
	Connected       ConnectionStatus = "connected"
	Connecting      ConnectionStatus = "connecting"
	ConnectionError ConnectionStatus = "error"
)
```

The only concern I have about not connecting/fallingback to the default driver is that this is not what submariner has done historically. The clusters are always connected in a full mesh fashion. If we only sync approved policies to all `clusters` in a `ClusterSet` - when a policy is declined a cluster will never see it and fall back to the default connection policy anyway. When the updated policy is approved - the connection would need to be restarted between a pair of clusters...

But as per your suggestion, then we would need to sync all policies regardless of their status (approved or declined) then we can extend the ConnectionStatus to include waiting for policy? or potentially just fallback to error and then there's no need for a new ConnectionStatus. But this introduces the concept of no connection existing between a pair of clusters.

```go
type ConnectionStatus string

const (
	Connected       ConnectionStatus = "connected"
	Connecting      ConnectionStatus = "connecting"
	ConnectionError ConnectionStatus = "error"
	ConnectionWaiting ConnectionStatus = "waiting"
)
```

tpantelis: I'm not clear on what it means for a `ClusterConnectionPolicy` to be _approved_, ie what is the criteria for approval right after creation? My view is that a conflict with other policies would be detected once it's synced to a cluster and when a GW tries to apply it to a particular remote `Endpoint`.  And conflicts would be per `Endpoint`, depending on the connection policy label selection criteria and the labels on the associated `Cluster` resources. And all of this is dynamic so even if we wanted to somehow do a pre-check after policy creation on the broker before syncing to member clusters to examine all current `Cluster` labels and policies to ensure no conflicts per cluster, a subsequent `Cluster` label change or policy change could introduce a conflict in which case would you somehow "un-approve" a policy to resolve the conflict? Which policy would you choose to un-approve and what would that mean (there may be a conflict with one `Cluster` but not another)? 

Re: status, I think it would make sense to reflect the policy decision in the GW's `Endpoint` connection status. We could add a _not-connected_ `ConnectionStatus` with the `StatusMessage`  indicating due to policy enforcement. 

If there's a policy conflict, we'd need to report that somewhere. We could simply use the _error_ `ConnectionStatus`  with an appropriate `StatusMessage`. I understand your point about falling back to the default but I'm not sure we should assume that's what the user would want plus the user may not be aware that there was a conflict. Failing the connection would effectively raise it to their attention.

maryamtahhan: I thought conflict could be detected much earlier on - via a controller or the syncer on the broker, covering both the pre-check and post deployment policy creation (as I also thought we are going to limit policy creation to the broker only and sync one way rather than bi-directionally)... And then I even thought that maybe a validation webhook might even be the way to go (build in the logic to detect conflicts much earlier on, this way only conflict free policies are approved for admission) so issues will be reported pretty quickly to the user...

I think if we are getting to the point of trying to establish a connection and then detecting a policy conflict we're a little late to the party. There could be many conflicting policies already causing no connections across the ClusterSet. Say for example (bear with me) an insider bad actor creates a bunch on conflicting policies to result in a set of no connections across your ClusterSet. Or even an unintended consequence could be for users to deliberately create conflicting policies to allow them to have no connections between clusters in a ClusterSet.

I'm not sure what the best approach is here, early detection would certainly avoid unintended consequences ... Later detection could be simpler but if we don't fall back to the default connection policy then could result in some unintended consequences/uses of the no connect feature.

submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,157 @@
...
+  // Backends lists all the enabled backends on an endpoint: vxlan, ipsec ...
+  Backends           []string          `json:"backends"`
+  // BackendsConfig is a map listing the configuration for a particular backend:
+  // e.g. 
+  //    backends_config:
+  //      vxlan:
+  //        Natt - Discovery - Port:  4490
+  //        Preferred - Server:       false
+  //        Udp - Port:               4500
+  BackendsConfig     map[string]map[string]string `json:"backends_config,omitempty"`
```
maryamtahhan: update the nested maps to a slice of structs...

submariner/cable-driver-policy.md
tpantelis: > I thought conflict could be detected much earlier on 

That would be ideal but, as I mentioned, conflicts could occur due to a mislabeled cluster, not just on policy creation/update. Let's consider an example.

I create 2 policies:

```yaml
leftClusterSelector:
  matchLabels:
    org: sales
rightClusterSelector:
  matchLabels:
    org: sales
cableDriverName: "driver1"

leftClusterSelector:
  matchLabels:
    org: engineering
rightClusterSelector:
  matchLabels:
    org: engineering
cableDriverName: "driver2"
```

The default cable driver is "_ipsec_".

There's no clusters labeled yet so no conflicts - the policies are approved, synced and all is good. Some time later, clusters _sales1_ and _sales2_  are labeled with `org: sales`. Only the first policy matches so there's no conflicts  and the GWs reconnect using "_driver1_". Clusters _engineering1_ and _engineering2_  are labeled with `org: engineering` and are reconnected using "_driver2_". Then _sales-and-eng1_ is labeled with `org: sales` and `org: engineering`. When connecting to  _sales1_ and _sales2_  only the first policy matches and when connecting to  _engineering1_ and _engineering2_ only the second policy matches. Again no conflicts. Then _sales-and-eng2_ is also labeled with `org: sales` and `org: engineering`. Now, when attempting to connect with _sales-and-eng1_, both policies match and there's a conflict.

So what remediation steps should be taken... Should we un-approve/revoke a policy? If so, which one? Or both? But what about the connectivity between _sales1_ and _sales2_? There's no conflict there so if the first policy is  revoked then they would implicitly and silently fall back to using "_ipsec_".  Similar with _engineering1_ and _engineering2_. Is it even the policies that need to be corrected or is it a mislabeled cluster?

I think this example illustrates how doing upfront checking and and rejecting policies can be problematic, especially since conflicts are per connection. 

We could do validation of a policy wrt the label selection syntax and the cable driver name.

> via a controller or the syncer on the broker, covering both the pre-check and post deployment policy creation  
> (as I also thought we are going to limit policy creation to the broker only and sync one way rather than bi-directionally)... 

yeah we sync one way, ie member cluster to or from the broker cluster. We've strived to keep the broker simple, ie no controllers although we did add an operator controller running on the broker. But that doesn't preclude us from adding another controller if there's a compelling need. 

> And then I even thought that maybe a validation webhook might even be the way to go 

yeah that would seem preferable over a controller for whatever validation we want to do.

> I'm not sure what the best approach is here, early detection would certainly avoid unintended consequences ... 

As outlined above, early detection would be preferable but it doesn't seem feasible in all cases. I'd think the common scenario would be that policies are created up front with conflicts occurring when clusters are subsequently labeled.

> Later detection could be simpler but if we don't fall back to the default connection policy then could result in some unintended consequences/uses of the no connect feature.

yeah perhaps you're right. We still need a way to report/record conflicts, perhaps update the `StatusMessage` in the connection status at the very least.

maryamtahhan: I see your point on the early checking not covering all scenarios.

And even with the no-connection and marking the status of a connection as a conflict - is certainly a better approach then a full fall back to a default driver and no error reporting...

I will write up an update based on the discussion points above. and then we can review and see what else is needed :)

submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,327 @@
...
+Kind:         Endpoint
+Metadata:
+  Name:         cluster1-submariner-cable-cluster1-172-18-0-14
+  Namespace:    submariner-operator
+Spec:
+  Backends:
+    cable_driver:
+        name: vxlan
+        config:
+          Natt - Discovery - Port:  4490
```
yboaron, line 292: I think we need to have a separate udp port for ipsec and VxLAN , right ? 

Otherwise, we might have a port conflict in case node needs to run both cable drivers.

submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,327 @@
...
+    // An empty selector indicates wildcard, ie matches any cluster.
+    RightClusterSelector metav1.LabelSelector `json:"rightClusterSelector,omitempty"`
+
+    // Name of the cable driver implementation to use for this connection.
+    CableDriver string `json:"CableDriver"`
+
+    // +optional
+    // NoConnectOnConflict defines the intention to fallback to no connection if there's a policy conflict
+    // If defined - should be part of the default policy
+    NoConnectOnConflict bool `json:"NoConnectOnConflict"`
```
tpantelis: This is really a global setting and as such seems awkward to define it per policy with the semantic that it only applies to the default policy. I think we should just decide the behavior one way or another. I still lean towards no-connect on conflict and mark the connection as such. A conflict means the user made a logic error with either the policies and/or labeling and thus the policy choice is non-deterministic so I think no-connect is the safest remediation. I think conflicts would be uncommon anyway - TBH, I struggled to even come up with a conflict example in a prior comment 😄

submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,327 @@
...
+    // LeftClusterSelector identifies the Cluster resources representing the clusters on one end of a connection. An empty
+    // selector indicates wildcard, ie matches any cluster.
+    LeftClusterSelector metav1.LabelSelector `json:"leftClusterSelector,omitempty"`
+
+    // RightClusterSelector identifies the Cluster resources representing the clusters on the other end of a connection.
+    // An empty selector indicates wildcard, ie matches any cluster.
+    RightClusterSelector metav1.LabelSelector `json:"rightClusterSelector,omitempty"`
+
+    // Name of the cable driver implementation to use for this connection.
+    CableDriver string `json:"CableDriver"`
```
tpantelis:
```suggestion
    CableDriver string `json:"cableDriver"`
```

submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,327 @@
...
+
+The `ClusterConnectionPolicy` CR defines the criteria for selecting which cable driver to use to connect cluster pairs (or
+subsets). To resolve the connectivity policy for a remote cluster's Endpoint, Submariner would:
+
+- Search for a policy whose left and right label selectors match the corresponding Cluster resources of the local and remote
+  Endpoints.
+
+> **_NOTE:_** In the case of multiple labels being specified, the selector will need to match all the labels.
+> Inequality-based requirements will be supported. For e.g. labels such as `env: !production` will match any clusters
+> that aren't labelled with `env: production`.
```
tpantelis: These semantics (and more) are defined by `metav1.LabelSelector`.

submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,327 @@
...
+
+If any matching policy is found, the clusters are connected using the cable driver specified in the cable driver policy.
```
tpantelis:
```suggestion
If one matching policy is found, the clusters are connected using the cable driver specified in the cable driver policy.
```

submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,327 @@
...
+If no matching policy is found, the `default` cable driver policy is used. The `default` policy can be auto-created at broker
+deploy time.
+
+If more than one matching policy is found, the default policy will consulted to establish if `NoConnectOnConflict` is set to true.
+In the case where `NoConnectOnConflict` is true the clusters are **not connected** and the connection status between a pair of clusters
+is marked as `conflict`. In the case where `NoConnectOnConflict` is false, the clusters are connected but the connection status between
+a pair of clusters is marked as `conflict`. If `NoConnectOnConflict` is undefined, it's assumed to be set to True.
+
+The initial `ClusterConnectionPolicy` resources would be created and maintained on the broker cluster and synced to each cluster in the
+ClusterSet.
```
tpantelis: What are the initial resources?

submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,327 @@
...
+The Connection information stored on a Gateway should be updated to reflect what policy is in use for that connection.
+
+```go
+type Connection struct {
+  Status        ConnectionStatus  `json:"status"`
+  StatusMessage string            `json:"statusMessage"`
+  Endpoint      EndpointSpec      `json:"endpoint"`
+  UsingIP       string            `json:"usingIP,omitempty"`
+  UsingNAT      bool              `json:"usingNAT,omitempty"`
+  PolicyID      string            `json:"policy_id,omitempty"`
```
tpantelis: A policy doesn't have an ID. We want to convey the policy used. 

```suggestion
  UsingPolicyID      string            `json:"usingPolicy,omitempty"`
```

submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,327 @@
...
+  cableType     string            `json:"cable_type,omitempty"`
```
tpantelis: I think you mean cable driver?

```suggestion
  UsingCableDriver     string            `json:"usingCableDriver,omitempty"`
```

submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,327 @@
...
+ClusterSet.
+
+Submariners `EndpointSpec` needs to be modified as `Backend` and `BackendConfig` only reflect a single cable driver. An alternative parameter
+`Backends` is proposed. This parameter represents a slice of supported backends and their configurations on an Endpoint. This information
+can be used in conjunction with the `ClusterConnectionPolicy` to validate a connection can be established (i.e. both ends of the connection
+support the same driver).
+
+The `EndpointSpec` could also be modified to advertise which policies it supports to further support the search for a suitable policy for
+a remote gateway trying to connect to that endpoint. The `SupportedPolicies` parameter is a simple slice of strings that identifies the set
+of policies that a Gateway has deemed as being applicable to it from the set of `ClusterConnectionPolicy` CRs.
```
tpantelis: I don't see a need for this. The set of `ClusterConnectionPolicy` CRs should be the source of truth used by a remote gateway. The `SupportedPolicies` would be calculated from the `ClusterConnectionPolicy` CRs and, due to propagation lag thru the broker, a remote gateway might not observe the most up-to-date setting and thus can't rely on it.

submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,327 @@
...
+If more than one matching policy is found, the default policy will consulted to establish if `NoConnectOnConflict` is set to true.
+In the case where `NoConnectOnConflict` is true the clusters are **not connected** and the connection status between a pair of clusters
+is marked as `conflict`. In the case where `NoConnectOnConflict` is false, the clusters are connected but the connection status between
+a pair of clusters is marked as `conflict`. If `NoConnectOnConflict` is undefined, it's assumed to be set to True.
+
+The initial `ClusterConnectionPolicy` resources would be created and maintained on the broker cluster and synced to each cluster in the
+ClusterSet.
+
+Submariners `EndpointSpec` needs to be modified as `Backend` and `BackendConfig` only reflect a single cable driver. An alternative parameter
+`Backends` is proposed. This parameter represents a slice of supported backends and their configurations on an Endpoint. This information
```
tpantelis: The `BackendConfig` has been overloaded to store NAT discovery and other gateway config in addition to cable driver config. The settings are extracted from labels/annotations on the GW mode. We may have to leave this as is.

The `Backend` is basically used by route agent code to determine if the cable driver is _wireguard_. Of course this assumes a single, statically-defined cable driver which will be problematic with multiple, dynamically-defined cable drivers. I'm not sure how to handle this.

submariner/cable-driver-policy.md
tpantelis: I assume you're referring to the `gateway.submariner.io/udp-port` annotation that a user can specify on the gateway node. This is used by the _libreswan_, _vxlan_, and _wireguard_ drivers which would seem to be problematic. I think we'll need to define a udp-port key for each driver.

submariner/cable-driver-policy.md
```diff
@@ -0,0 +1,327 @@
...
+
+### Cluster local Connections Resources
+
+The detailed connection information for a Gateway is currently maintained in the Gateway and Submariner CRs,
+however when the number of clusters grows to a large scale these could be difficult to parse. It maybe preferable
+to maintain connection information in a separate set of `ClusterConnection` CRs. This could be useful in the case of:
+
+- Gateway failover, so the new gateway doesn't need to do the same work again to build a connection definition. Or
+- In the case of `ClusterConnectionPolicy` updates. A Gateway could quickly cycle through the connection policies its
+using in the `ClusterConnection` CRs to see if an update impacts it.
```
tpantelis: I'm not clear on the purpose of this. For policy updates, I imagine the GW process would maintain connectivity information in memory in some form to determine the impact of a policy change. In fact I think it could simply re-queue all remote endpoints to re-evaluate them.

submariner/cable-driver-policy.md
yboaron: Yeah, these ports are the 'frontend' ports of the GW pod, currently, we have one external udp port (default 4500) that is used for both ipsec and vxlan.

I assume we need to update also the firewall configuration (cloud prepare) in this context (and this should affect also GW K8S LB service when gw pods will run as backends of k8S LBS, which is planned to be supported in  0.13).

tpantelis: Code like [this](https://github.com/submariner-io/submariner/blob/c047698ccbf1fcc43b2dcd7c0773acc902914f61/pkg/routeagent_driver/handlers/kubeproxy/routes_iface.go#L115) in the route agent is also problematic for supporting multiple cable drivers. There may be other hidden assumptions lurking that are tied to having one driver. It seems we need to determine if it's even feasible to run multiple cable drivers in the same cluster and, if so, how.

yboaron: Agree, I think it makes sense first to analyze what components (the ones that assume 1:1 mapping) need to be refactored and refactor these components (I assume that should be a separate Epic)  and after that design&implement new features (like connections policy).

submariner/cable-driver-policy.md
maryamtahhan: Ok - I will remove it and modify the description of the behaviour to reflect no connection on policy conflict

submariner/cable-driver-policy.md
maryamtahhan: Will remove

submariner/cable-driver-policy.md
maryamtahhan: the default ClusterConnectionPolicy will be the initial resource

submariner/cable-driver-policy.md
maryamtahhan: This is indeed problematic

submariner/cable-driver-policy.md
maryamtahhan: Fair enough - I will remove

submariner/cable-driver-policy.md
maryamtahhan: I guess it can be maintained in memory also... Just thought if it was in a CR it would be easy enough to filter through to the right CR using labels pretty quickly. I will remove reference to the `ClusterConnectionPolicy` updates

submariner/cable-driver-policy.md
maryamtahhan: So is the suggestion here to close this out till such analysis is complete?

tpantelis: I suggest we note these problematic issues in the proposal and address the remaining comments and merge it as in progress so we can pick it up later.

maryamtahhan: > I think we need to have a separate udp port for ipsec and VxLAN , right ?
> 
> Otherwise, we might have a port conflict in case node needs to run both cable drivers.

It looks like ports in the range 4489-4499 could be used https://www.speedguide.net/port.php?port=4490

tpantelis (approved):

tpantelis:
@skitt Can you re-review?

skitt (approved):

PR #111, opened 2022-04-18
Aggregate ServiceImport enhancement proposal

donaldh:
Signed-off-by: Donald Hunter <donald.hunter@redhat.com>

### DIFF
--- a/lighthouse/aggregate-service-import.md
+++ b/lighthouse/aggregate-service-import.md
@@ -1,0 +1,137 @@
+# Aggregate Service Import for Lighthouse
+
+## Summary
+
+The current Lighthouse implementation deviates from the MCS specification in the following ways:
+
+- ServiceExports with matching namespace/name are not combined into a single ServiceImport.
+  Each importing cluster receives multiple ServiceImports that are multiplexed at DNS lookup.
+- Conflict resolution of ServiceExports is not implemented [^1].
+
+In order to better align with the MCS specification the following functionality needs to be
+added to the Lighthouse agent:
+
+- Merge matching `ServiceExport` objects into a single `ServiceImport` object
+- Mark conflicting `ServiceExport` objects with the `ServiceExportConflict` condition
+
+There are changes required in Lighthouse's handling of `EndpointSlices` before single
+`ServiceImports` can be introduced.
+
+- The `ServiceImports` for an exported service each contain the service VIP from the exporting
+  cluster. `ServiceImports` cannot be merged like this because a `ServiceImport` cannot have
+  more than one IP. The semantics of the IP are that it should be a cluster local Service VIP,
+  if needed. The service VIPs from the exporting clusters should be stored in `EndpointSlices`,
+  with a separate `EndpointSlice` per exporting cluster.
+
+It is worth noting that:
+
+- Headless services already have `EndpointSlices` containing the correct pod IP addresses, and contain
+  no VIP information in the multiple `ServiceImports`. A merged  `ServiceImport` for headless
+  services would comprise a merge of the port definitions in those `ServiceImports`.
+
+- The current GlobalNet implementation exports `EndpointSlices` for normal services that contain
+  pod IP addresses from the originating cluster.
+
+## Proposal
+
+Submariner has a passive broker, it is a central apiserver that is accessed by each cluster in
+the clusterset. Clusters synchronise resources to and from the broker with all sync and
+reconciliation logic implemented in the clusters.
+
+In this proposal we retain the passive broker and take an approach to merging `ServiceImports`
+where exporting clusters handle the merging. Simplistically, a merged `ServiceImport` outcome
+can be achieved by allowing the first exporting cluster to create a `ServiceImport` and all
+subsequent exporting clusters to reuse it, assuming they have matching port lists.
+
+1. First exporting cluster creates a `ServiceImport` on the broker.
+1. Subsequent exporting clusters reuse the existing `ServiceImport`.
+1. Exporting clusters create and sync `EndpointSlices` to the broker with the service VIP and a
+   `multicluster.kubernetes.io/service-name` label that matches the `ServiceImport`.
+1. All clusters sync the `ServiceImport` and `EndpointSlices` from the broker.
+
+Looking at the `ServiceImport` merging in more detail, the MCS specification asks for all port
+definitions from the `ServiceExports` to be combined in the merged `ServiceImport`. This means
+that an exporting cluster would need to merge its port definitions into the `ServiceImport`.
+Conversely, when removing a `ServiceExport` any merged port definitions would need to be
+removed. However an exporting cluster only has visibility of its local `ServiceExport` so cannot
+unmerge with the information available locally.
+
+Note: It is an open question what behaviour is expected when a connection request for service
+port is directed to a cluster that does not expose that port.
+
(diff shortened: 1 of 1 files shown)


donaldh:
This proposal is linked to https://github.com/submariner-io/lighthouse/issues/214

lighthouse/aggregate-service-import.md
```diff
@@ -0,0 +1,137 @@
...
+
+In this proposal we retain the passive broker and take an approach to merging `ServiceImports`
+where exporting clusters handle the merging. Simplistically, a merged `ServiceImport` outcome
+can be achieved by allowing the first exporting cluster to create a `ServiceImport` and all
+subsequent exporting clusters to reuse it, assuming they have matching port lists.
+
+1. First exporting cluster creates a `ServiceImport` on the broker.
+1. Subsequent exporting clusters reuse the existing `ServiceImport`.
+1. Exporting clusters create and sync `EndpointSlices` to the broker with the service VIP and a
+   `multicluster.kubernetes.io/service-name` label that matches the `ServiceImport`.
```
aswinsuryan, line 53: I think we can also think of maintaining ServiceImport per cluster which contain the cluster-IP of the exported cluster in the submariner-operator namespace of each cluster and have an aggregated service import in the Service Namespace. Conflict resolution still can happen based on the presence of a ServiceImport in the submariner-operator namespace.

This way we can leave the endpointslices as it is. Though Submariner does not create multiple endpointslices now it is expected that an enpointslice should contain only a specified number of endpoints (100 by default). So there may be multiple endpoint slices per service per cluster.

lighthouse/aggregate-service-import.md
```diff
@@ -0,0 +1,137 @@
...
+`EndpointSlices` associated with the `ServiceImport` instead of using `ServiceImports`
+directly.
+
+DNS Resolution behaviour for headless services will remain unchanged.
+
+### Service VIPs in Importing Clusters
+
+Given the way that Submariner operates, there is no need to assign a Service VIP to a
+`ServiceImport` in the importing clusters. This means that `ServiceImports` can always have an
+empty IP address attribute, like they do for headless services today.
```
aswinsuryan, line 76: We can think of eventually adding VIP as well so that it can work with any component that consumes MCS API. Maybe globalnet can help in providing a globally unique VIP per service.

donaldh: I think the goal would be to perform load balancing across the workload clusters behind a cluster local Service VIP, so this would be a move from DNS load balancing to a ServiceImport load balancer. I suggest that it be tracked as follow-on work that is out of scope for this enhancement.

lighthouse/aggregate-service-import.md
donaldh: `EndpointSlices` can remain as they are for non-global net deployments, but this implies not using service load balancing in the destination cluster, which I thought was desired for normal clusterip services. For globalnet they currently contain unreachable pod IPs for normal clusterip services. I think it is desirable to change `EndpointSlice` to contain the target cluster's globalnet service VIP for clusterip services and retain the existing behaviour for headless services.

lighthouse/aggregate-service-import.md
vthapar: Yes, wouldbe good to have this option but it will have to be a config option that can be enabled/diabled. At least for now we don't have any plans to implement an mcs controller that will loadbalance using endpointslices in datapath instead of DNS, but it will allow us to use Lighthouse-only deployments to provide discovery for other cluster connectivity solutions.

lighthouse/aggregate-service-import.md
```diff
@@ -0,0 +1,137 @@
...
+definitions from the `ServiceExports` to be combined in the merged `ServiceImport`. This means
+that an exporting cluster would need to merge its port definitions into the `ServiceImport`.
+Conversely, when removing a `ServiceExport` any merged port definitions would need to be
+removed. However an exporting cluster only has visibility of its local `ServiceExport` so cannot
+unmerge with the information available locally.
+
+Note: It is an open question what behaviour is expected when a connection request for service
+port is directed to a cluster that does not expose that port.
+
+The notion that services are equivalent when they have matching name and namespace but different
```
vthapar: Can we make it a requirement/limitation that services must have same name, namespace *and* matching port definitions? Mismatch in ports also results in a ServiceExport conflict. Maybe add a Design Limitation section and explain it all there, alongwith a comment on how it aligns with MCS API [as in implementation choice].

lighthouse/aggregate-service-import.md
```diff
@@ -0,0 +1,137 @@
...
+can be achieved by allowing the first exporting cluster to create a `ServiceImport` and all
+subsequent exporting clusters to reuse it, assuming they have matching port lists.
+
+1. First exporting cluster creates a `ServiceImport` on the broker.
+1. Subsequent exporting clusters reuse the existing `ServiceImport`.
+1. Exporting clusters create and sync `EndpointSlices` to the broker with the service VIP and a
+   `multicluster.kubernetes.io/service-name` label that matches the `ServiceImport`.
+1. All clusters sync the `ServiceImport` and `EndpointSlices` from the broker.
+
+Looking at the `ServiceImport` merging in more detail, the MCS specification asks for all port
```
vthapar: Can you reword this section? This being EP, focus should be on our design approach. So open with what we're going to do. Then explain in a note how it varies from MCS by requiring limitation of name, namespace and ports. Starting off with explanation of why we made this design choice breaks the flow of sequence of steps.

vthapar (changes requested):
Thanks for this @donaldh Overall looks good.

lighthouse/aggregate-service-import.md
```diff
@@ -0,0 +1,137 @@
...
+- Merge matching `ServiceExport` objects into a single `ServiceImport` object
+- Mark conflicting `ServiceExport` objects with the `ServiceExportConflict` condition
+
+There are changes required in Lighthouse's handling of `EndpointSlices` before single
+`ServiceImports` can be introduced.
+
+- The `ServiceImports` for an exported service each contain the service VIP from the exporting
+  cluster. `ServiceImports` cannot be merged like this because a `ServiceImport` cannot have
+  more than one IP. The semantics of the IP are that it should be a cluster local Service VIP,
+  if needed. The service VIPs from the exporting clusters should be stored in `EndpointSlices`,
```
vthapar, line 27: We currently use information in endpointslices for ClusterIP Services to determine if backing pods in that cluster are available or not. This is used to avoid directing queries to clusters which have no active pods. How will this be addressed? We should withdraw that cluster's ServiceIP from EndpointSlice and add it back once at least one active pod is available.

vthapar (changes requested):

lighthouse/aggregate-service-import.md
donaldh: Good suggestion – I will add an explicit requirement that service equivalence will be same name, namespace and port definitions. I will add a design limitation section.

lighthouse/aggregate-service-import.md
donaldh: Agreed, I will rework this to separate out the proposed approach from the background and design constraint.

vthapar (approved):

tpantelis (approved):

aswinsuryan (approved):

PR #121, opened 2022-05-30
Add cloud-prepare/vpc-peering proposal

ruromero:
Signed-off-by: ruromero <rromerom@redhat.com>


cloud-prepare/vpc-peering.md
```diff
@@ -0,0 +1,210 @@
+# Submariner VPC Peering Support
+
+This document describes the technical details and impact of our proposal design
+to include Virtual Private Cloud (VPC) Peering support for Submariner and
```
nyechiel: Note: VPC Peering is a term used by AWS and GCP, but other providers have different names for it. For e.g, the equivalent in Azure is "virtual network peering".

cloud-prepare/vpc-peering.md
```diff
@@ -0,0 +1,210 @@
...
+`subctl` located at this [Pull Request](https://github.com/submariner-io/cloud-prepare/pull/190).
+
+## Abstract
+This idea was born meanwhile the CockroachDB project focused to deploy a
+multi-cluster setup of CockroachDB and communicate the different clusters across
+regions using Submariner.
+
+One of the requirements generated from that project was to deploy CockroachDB on
```
nyechiel, line 13: One may wonder why Submariner is even needed if something like VPC Peering is in place. Doesn't the underlying cloud provide L3 routing between the clusters on top of the peering connection? If so, the case for using Submariner might be limited to Lighthouse (DNS) only? @sridhargaddam can you comment on that?

cloud-prepare/vpc-peering.md
```diff
@@ -0,0 +1,210 @@
...
+This command creates a VPC Peering between different clusters of the same cloud
+provider.
+
+Usage:
+  subctl cloud VPC Peering create [command]
+
+Available Commands:
+  aws         Create a VPC Peering on AWS cloud
+  gcp         Create a VPC Peering on GCP cloud
+  generic     Create a VPC peering between generic clusters for Submariner
```
nyechiel: What does `generic` mean in this case?

cloud-prepare/vpc-peering.md
```diff
@@ -0,0 +1,210 @@
...
+
+Global Flags:
+      --kubeconfig string    absolute path(s) to the kubeconfig file(s)
+      --kubecontext string   kubeconfig context to use
+
+Use "subctl cloud VPC Peering create [command] --help" for more information
+about a command.
+```
+
+#### VPC Peering cloud-provider option (Available for create & clean subcomands)
```
nyechiel: What about the use-case where a user deploys the VPC Peering outside of subctl control?

cloud-prepare/vpc-peering.md
ruromero: Can you or someone propose a better name?

cloud-prepare/vpc-peering.md
ruromero: We understood the `generic` implementation is some kind of no-op that you agree to have. It is implemented in other functionalities. Shall we remove it?

cloud-prepare/vpc-peering.md
aswinsuryan: With the peering, only the nodes of two Kubernetes clusters will be reachable.  We still need a component like Submariner. for the pods/service of two Kubernetes clusters to be reachable,

cloud-prepare/vpc-peering.md
```diff
@@ -0,0 +1,210 @@
...
+
+We purpose to add some subcommands to `subctl` to use the new VPC Peering
+features described on this document.
+
+#### VPC Peering Subcommands
+
+This will be how new added sub-commands would be like.
+
+```bash
+$ ./bin/subctl cloud VPC Peering
```
aswinsuryan: I think we can consider making this a flag for the cloud prepare command  too, like --peer( or some other name ) network-info

ruromero: the problem with the flag approach is that it doesn't follow the same setup as the other cases because we have to specify between which clusters we want to do the peering

cloud-prepare/vpc-peering.md
nyechiel: Yes, and I think we should explicitly mention that in the design doc, i.e what is the value of Submariner in this type of configuration.

cloud-prepare/vpc-peering.md
nyechiel: Maybe "network peering"? @sridhargaddam @mkolesnik wdyt?

cloud-prepare/vpc-peering.md
```diff
@@ -0,0 +1,210 @@
...
+This command creates a VPC Peering between different clusters of the same cloud
+provider.
+
+Usage:
+  subctl cloud VPC Peering [command]
```
nyechiel: See my above comment on the VPC Peering term. It would make sense for AWS and GCP, but not for Azure.

cloud-prepare/vpc-peering.md
vthapar: My vote for `peering` IT is generic enough and all clouds use `peering` in their offering.

AWS: VPC Peering
GCP: VPC NEtwork Peering or Cloud Peering.
Azure: Virtual Network Peering.

So command would look like `subctl cloud peering`

cloud-prepare/vpc-peering.md
vthapar: Yeah, we discussed this early on. Flag can't work coz it explicitly requires at least two. Only other potion would be to setup peering between many with a single command. Which can be done in a future enhancement, with input taking list of clouds from a yaml/json file.

cloud-prepare/vpc-peering.md
vthapar: Yeah, it will only lead to confusion for users. +1 to remove it.

cloud-prepare/vpc-peering.md
vthapar: Should be possible to track if peering was done by submariner or not. We should only cleanup if done by submariner.

ruromero:
updated proposal with your comments

ruromero:
@nyechiel I have rebased against `devel` any further comments?

nyechiel:
LGTM, but I'd like @skitt @vthapar @tpantelis @sridhargaddam @mkolesnik to review.

mkolesnik (commented):
My main concern with proposals such as this one, is that it goes very deep into the infrastructure side which is not our main focus in submariner (nor is it an area of expertise).

The thing is, there's nothing missing in Submariner to have an infra set up with any sort of peering to work, right? You'd still run `cloud prepare` on that infra to open the FW ports and spin up dedicated nodes, just as you would any infra that's not using peering.

Hence, adding this support adds a maintenance burden for submariner maintainers to set up this part of the infra, whereas I feel there are already much more powerful and versatile tools (eg terraform) which can take care of more profound infrastructure setup.

I feel we should focus on the features important for submariner, and solve any impediments necessary to make different infrastructure work, but not so much on setting up the infrastructure itself.

r2dedios:
Hi @mkolesnik, Is not strictly mandatory to config something in submariner but it could help a lot to connect cluster between them and also with a better configuration for this scenario using VXLAN instead of Libreswan (IPSec) because a VPC-peering is a private bridge between two VPCs

mkolesnik:
Closing based on recent discussion on Jul 27, 2022 - https://docs.google.com/document/d/19UzvTi9FaWK2ps8SNi-xXmw7q71LpIygBp_KmGEZ-s0/edit#

Specifically:
> Proposal is to maintain current cloud-prepare code base, but not expanding much or adding new significant features/providers

PR #125, opened 2022-07-20
Add empty commit

### LINKED ISSUE #124: PR dependencies bot fails on this repo: "SHA and context has reached the maximum number of statuses"
I think this is because there are very few new commits merged to this repo relative to the number of times this check runs (on each comment). That results in the same SHA getting checked many times, which is hitting a limit at the GH API.

Merging any commit should fix this. Should maybe also run the dependency check less frequently to avoid in the future. A better solution would be welcome.

tpantelis:
...to fix https://github.com/submariner-io/enhancements/issues/124.


skitt (approved):

dfarrell07 (approved):
:crossed_fingers:

dfarrell07:
It doesn't seem like this worked, a run after this failed.

https://github.com/submariner-io/enhancements/actions/runs/2706720380

PR #128, opened 2022-08-11
Submariner metrics redesign

vthapar:
Submariner Gateway and Globalnet metrics require port to be opened
on host for them to be accessible. This EP proposes a redesign
which allows metrics to be accessible without requiring to open
any ports in infra.

Refer github.com/submariner-io/submariner/issues/1944

### DIFF
--- a/submariner-operator/submariner-metrics-redesign.md
+++ b/submariner-operator/submariner-metrics-redesign.md
@@ -1,0 +1,164 @@
+# Submariner metrics redesign
+
+Related Issue:
+[Exporting Submariner metrics without needing to open ports](https://github.com/submariner-io/submariner/issues/1944)
+
+## Summary
+
+Submariner Gateway and Globanet pods run with HostNetworking.
+To access metrics exported by these two we need to open ports in the infra firewall.
+With the move towards minimizing changes needed in cloud infra for Submariner,
+there is a use case to export metrics in a way that doesn't require any of these ports to be opened.
+The current design also has a limitation where these ports might be in use by some other service,
+causing a conflict and resulting in metrics becoming inaccessible.
+
+## Proposal
+
+This enhancement proposes a redesign of existing metrics for Submariner Gateway and Globalnet by adding a new metrics proxy
+service. Pods running on the same node can access ports without requiring ports to be opened in the cloud infra.
+Queries for metrics will reference the proxy service instead of the Gateway and Globalnet metrics ports directly,
+and each proxy will provide access to the metrics on the same node.
+
+This proposal also recommends changing the ports used by the metrics HTTP servers in the Gateway and Globalnet pods to a free range
+port to avoid conflicts with ports already in use on the host network.
+
+## Design Details
+
+* A new DaemonSet `submariner-metrics-proxy` will be created. It will use the NodeSelector `submariner.io/gateway=true`
+to run on all gateway nodes, active or not.
+* The DaemonSet will also use tolerations to make sure it can always be scheduled on the gateway node, irrespective of any taints.
+* The `submariner-metrics-proxy` Pod will use `netcat` to act as proxy and forward all incoming requests on metrics service port to the
+metrics port on the node IP.
+* The proxy pod will run two containers if Globalnet is enabled: one each for Gateway and Globalnet.
+* The existing metrics service for Gateway and Globalnet will be modified to use the new metrics proxy pod. This will be
+done by using same label selector as the metrics proxy DaemonSet. This will make sure users can continue using existing services
+and ports without needing to make any changes themselves.
+* `submariner-gateway` will be modified to use port `32780` instead of `8080` and `submariner-globalnet` will be
+modified to use `32781` instead of `8081`. A new env variable will be used for these ports.
+* [Optional] Modify `subctl cloud prepare` to no longer open metrics ports.
+
+### Submariner Metrics DaemonSet
+
+```yaml
+---
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: submariner-metrics-proxy
+spec:
+  selector:
+    matchLabels:
+      app: submariner-metrics-proxy
+  template:
+    metadata:
+      labels:
+        app: submariner-metrics-proxy
+    spec:
+      nodeSelector:
+        "submariner.io/gateway": "true"
+      containers:
+        - name: gateway-metrics-proxy
+          image: quay.io/submariner/nettest:devel
+          imagePullPolicy: IfNotPresent
+          env:
+          - name: NODE_IP
+            valueFrom:
+              fieldRef:
+                fieldPath: status.hostIP
+          - name: SUBM_GW_METRICS_PORT
+            value: "32780"
+          command: ["/usr/bin/nc"]
+          args: ["-v", "-lk", "-p", "8080", "-e", "nc", "$(NODE_IP)", "$(SUBM_GW_METRICS_PORT)"]
+        - name: globalnet-metrics-proxy
+          image: quay.io/submariner/nettest:devel
+          imagePullPolicy: IfNotPresent
+          env:
+          - name: NODE_IP
+            valueFrom:
+              fieldRef:
+                fieldPath: status.hostIP
+          - name: SUBM_GN_METRICS_PORT
+            value: "32781"
+          command: ["/usr/bin/nc"]
+          args: ["-v", "-lk", "-p", "8081", "-e", "nc", "$(NODE_IP)", "$(SUBM_GN_METRICS_PORT)"]
+      tolerations:
+      - operator: Exists
+      restartPolicy: Always
+```
+
+### Changes to Submariner Gateway Metrics service
+
+```diff
+metadata:
+  name: submariner-gateway-metrics
+  labels:
+-   app: submariner-gateway
++   app: submariner-metrics-proxy
+...
+spec:
+  selector:
+-   app: submariner-gateway
++   app: submariner-metrics-proxy
+```
+
+### Changes to Submariner Globalnet Metrics service
+
+```diff
+metadata:
+  name: submariner-globalnet-metrics
+  labels:
+-   app: submariner-globalnet
++   app: submariner-metrics-proxy
+...
+spec:
+  selector:
+-   app: submariner-globalnet
++   app: submariner-metrics-proxy
+```
+
+### Additional Details
+
+These changes will be done in the following order to avoid any breakage in CI:
+
+1. Modify `subctl diagnose` to actually query metrics rather than check for metrics port availability on firewall.
+2. Create and deploy the new metrics proxy service. As part of this change the Gateway and Globalnet
+DaemonSets will be modified to set the metrics port in an env variable. They will still use the current ports, `8080`
+and `8081`. This change will be made in the `submariner-operator` repo.
+3. Modify gateway and globalnet code to use the new metrics port env variable when creating their metrics HTTP servers. If not
+available, it will default to the new values. This change will be in the `submariner` repo.
+4. Modify the metrics port env variable to use new values. This change will be in `submariner-operator` repo.
+
+### Pros
+
+* No user facing changes to existing services.
+* Only one extra pod created for both gateway and globalnet.
+
+### Cons
+
+* Adds extra pod(s).
+
+### Backward Compatibility
+
+`subctl diagnose` checks for metrics port in firewall. This test will need to be modified to not check for port itself
+but try acessing the metrics irrespective of ports being open or not. This makes this change not backward compatible with
+older versions of `subctl`.
+
+Older versions of `subctl diagnose` will not work these changes. So users will have to make sure they're using compatible
+version of `subctl`. The new `subctl diagnose` will work with older versions of subctl as `diagnose` changes will be fully
+backwards compatible.
+
+### Alternatives
+
+* Run separate proxy for globalnet and gateway. This would add an extra pod without adding any value as those pods will
+still need to run on the same node. Rejected because extra overhead without any value add.
+* Create a single service for globalnet and gateway metrics. This would not be backward compatible because users using
+existing service names to pull metrics will have to modify their setups.
+
+## External Dependencies
+
+None.
+
+## User Impact
+
+None.


submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,146 @@
...
+This enhancement proposes a redesign of existing metrics for Submariner Gateway and Globalnet by adding a new metrics proxy
+service. Pods running on the same node can access ports without requiring pods to be opened in the cloud infra. Any queries
+for these metrics will land on the proxy which will always run on the same node.
+
+Note that this proposal doesn't address the issue where metrics port is unavailable on the node. That should be addressed as
+a separate enhancement.
+
+## Design Details
+
+* A new `status.submariner.io/gateway=active` label will be added to gateway node that is currently active.
```
mkolesnik: Why does it matter if it runs only on the active node, or on all gateway labeled nodes?

Also why label node and not use selector based on the pod itself?
There's already a `gateway.submariner.io/status: active` label on the active pod.

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,146 @@
...
+* Submarine RouteAgent will set the label on the relevant Node.
```
mkolesnik: How is RA even connected to all this? The metrics are exported on GW and on GN pods, no?

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,146 @@
...
+* A new DaemonSet `submariner-metrics` will be created. It will use the NodeSelector `status.submariner.io/gateway=active`
```
mkolesnik: As mentioned, why not use node affinity to have it run on nodes that have the labeled pod?

Ref https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#inter-pod-affinity-and-anti-affinity

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,146 @@
...
+  selector:
+    matchLabels:
+      app: submariner-metrics
+  template:
+    metadata:
+      labels:
+        app: submariner-metrics
+    spec:
+      nodeSelector:
+        "status.submariner.io/gateway": "active"
```
mkolesnik:
```suggestion
        submariner.io/gateway: "true"
      affinity:
        podAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
          - labelSelector:
              matchLabels:
                gateway.submariner.io/status: active
            topologyKey: "kubernetes.io/hostname"
```

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,146 @@
...
+          image: quay.io/submariner/nettest:devel
+          imagePullPolicy: IfNotPresent
+          env:
+          - name: NODE_IP
+            valueFrom:
+              fieldRef:
+                fieldPath: status.hostIP
+          - name: SUBM_GW_METRICS_PORT
+            value: "8080"
+          command: ["/bin/sh", "-c", "nc -v -lk -p 8080 -e nc $(NODE_IP) $(SUBM_GW_METRICS_PORT)"]
```
mkolesnik:
```suggestion
          command: ["/bin/sh", "-c", "nc -v -lk -p 8080 -e nc ${NODE_IP} ${SUBM_GW_METRICS_PORT}"]
```

Don't you mean `${}`?

Won't `$()` try to execute the value?

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,146 @@
...
+## Design Details
+
+* A new `status.submariner.io/gateway=active` label will be added to gateway node that is currently active.
+* Submarine RouteAgent will set the label on the relevant Node.
+* A new DaemonSet `submariner-metrics` will be created. It will use the NodeSelector `status.submariner.io/gateway=active`
+to make sure it always runs on the active gateway node.
+* This DaemonSet will use `schedulerName: default-scheduler` to make sure that metrics proxy pod moves with the active gateway
+label.
+* DaemonSet will also use toleration to make sure it can always be scheduled on gateway node, irrespective of any taints.
+* `submariner-metrics` Pod will use `netcat` to act as proxy and forward all incoming requests on metrics port to the
```
mkolesnik: The idea to utilize a proxy makes sense :+1:

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,146 @@
...
+done by using same label selector as metrics DaemonSet.
+
+### Submariner Metrics DaemonSet
+
+```yaml
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: submariner-metrics
```
mkolesnik:
```suggestion
  name: submariner-metrics-proxy
```

Should probably suffix with `proxy` to make sure its not confused with the metrics services themselves

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,146 @@
...
+spec:
+  selector:
+    matchLabels:
+      app: submariner-metrics
```
mkolesnik:
```suggestion
      app: submariner-metrics-proxy
```

Same here..

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,146 @@
...
+  template:
+    metadata:
+      labels:
+        app: submariner-metrics
```
mkolesnik:
```suggestion
        app: submariner-metrics-proxy
```

Same here..

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,146 @@
...
+```
+
+### Changes to Submariner Gateway Metrics service
+
+```diff
+metadata:
+  name: submariner-gateway-metrics
+  labels:
+-   app: submariner-gateway
++   app: submariner-metrics
```
mkolesnik:
```suggestion
+   app: submariner-metrics-proxy
```

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,146 @@
...
+...
+spec:
+  selector:
+-   app: submariner-gateway
++   app: submariner-metrics
```
mkolesnik:
```suggestion
+   app: submariner-metrics-proxy
```

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,146 @@
...
+```
+
+### Changes to Submariner Globalnet Metrics service
+
+```diff
+metadata:
+  name: submariner-globalnet-metrics
+  labels:
+-   app: submariner-globalnet
++   app: submariner-metrics
```
mkolesnik:
```suggestion
+   app: submariner-metrics-proxy
```

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,146 @@
...
+...
+spec:
+  selector:
+-   app: submariner-globalnet
++   app: submariner-metrics
```
mkolesnik:
```suggestion
+   app: submariner-metrics-proxy
```

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,146 @@
...
+```
+
+### Pros
+
+* No user facing changes to existing services
+* Only one extra pod created for both gateway and globalnet
+
+### Cons
+
+* Requires changes in route-agent code to add the new label.
```
mkolesnik: Not if we use pod selector

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,146 @@
...
+* Adds an extra pod
+
+### Backward Compatibility
+
+Could modified gateway and globalnet metrics services be an issue? Upgrades?
```
mkolesnik: Should be transparent no?
Once the new proxy is deployed and the services pointed to it, the callers should be unaffected.

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,146 @@
...
+causing a conflict and metrics becoming inaccessible. With the move towards minimizing changes needed in cloud infra for
+submariner, there is a use case to export metrics in a way that doesn't require any of these ports to be opened.
+
+## Proposal
+
+This enhancement proposes a redesign of existing metrics for Submariner Gateway and Globalnet by adding a new metrics proxy
+service. Pods running on the same node can access ports without requiring pods to be opened in the cloud infra. Any queries
+for these metrics will land on the proxy which will always run on the same node.
+
+Note that this proposal doesn't address the issue where metrics port is unavailable on the node. That should be addressed as
```
mkolesnik: Why not address it as part of this work?

We can have different metrics ports with this, so why not choose a free range [1] and use ports from there?
User's shouldn't be affected (since they're using the service), and we'll be solving both the problems mentioned in the summary.

Once the solution is implemented, the scope of work of changing the ports is much smaller, so it can be part of the work here.

[1] https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers#Registered_ports

mkolesnik (dismissed):
Good idea, I had a few comments and suggestions

submariner-operator/submariner-metrics-redesign.md
vthapar: It needs to run `only` on active nodes coz only there we will have metrics available. If it runs on inactive nodes, metrics will not be available.

nodeSelector only uses labels availble on nodes, not pods.

submariner-operator/submariner-metrics-redesign.md
vthapar: RA was suggested by sridhar as code in RA already tracks active GW and labels the pod with active. So we can add code in there to label node as active gw node.

submariner-operator/submariner-metrics-redesign.md
vthapar: Wasn't familiar with it. This would definitely be even better way. If I'm understanding correctly, this means our pod will be hosted on whichever node has the gw pod, right?

UPDATE: Tried with pod affinity. It creates a new pod in `Pending` state on node which is inactive.

```
submariner-metrics-khxsc                         2/2     Running   0          3m33s
submariner-metrics-tfql9                         0/2     Pending   0          43s
```

submariner-operator/submariner-metrics-redesign.md
vthapar: Yeah, `{}` will work better. `()` works too. In an earlier version I was trying to get pod's ip so had used (). Missed change to `{}` after moving to env coz it was working.

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,146 @@
+# Submariner metrics redesign
+
+Related Issue:
+[Exporting Submariner metrics without needing to open ports](https://github.com/submariner-io/submariner/issues/1944)
+
+## Summary
+
+Submariner Gateway and Globanet pods run with HostNetworking. To access metrics exported by these two we need to open ports
+in the infra firewall. Currently design has a limitation where these ports might be in use by some other service,
```
sridhargaddam:
```suggestion
in the infra firewall. Current design has a limitation where these ports might be in use by some other service,
```

submariner-operator/submariner-metrics-redesign.md
mkolesnik: >  code in RA already tracks active GW and labels the pod with active.

That sounds like a bad design, unless I'm missing something?

The active GW should mark itself active, or perhaps the operator, but surely not the route agent as that one actually is supposed to only handle routing

submariner-operator/submariner-metrics-redesign.md
sridhargaddam: +1 to addressing it as part of this work. The submariner-gateway and submariner-globalnet pods can listen on non standard ports and the proxy can listen on the standard ports.

submariner-operator/submariner-metrics-redesign.md
mkolesnik: Yes but we can generally ignore this pod.

submariner-operator/submariner-metrics-redesign.md
sridhargaddam: Actually my idea behind using RA was to check the feasibility to run the proxy container as part of RA pod itself, but I later realized that even RA runs with HostNetworking, so it may not work. 

Anyways, if we decide to label the nodes, then we can use the submariner-gateway pod itself (instead of RA pod).
Currently, as soon as Submariner Gateway pod becomes active, it labels itself as "gateway.submariner.io/node:active" as shown here - https://github.com/submariner-io/submariner/blob/24bf0b818ab0bfffe0facf90daeb1e20a5382a02/main.go#L181

We can include additional code in this API. Since we are discussing about getting rid of node labelling (for LB use-cases), probably we should look at other options to achieve this use-case.

submariner-operator/submariner-metrics-redesign.md
mkolesnik: Yes I'm not sure exaclty why it's working with `$()` I'd expect it to complain about missing command or something, perhaps @skitt has an idea?

sridhargaddam (dismissed):

submariner-operator/submariner-metrics-redesign.md
sridhargaddam: What is the significance of topologyKey: "kubernetes.io/hostname" ?

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,146 @@
...
+        - name: globalnet-metrics-proxy
+          image: quay.io/submariner/nettest:devel
+          imagePullPolicy: IfNotPresent
+          env:
+          - name: NODE_IP
+            valueFrom:
+              fieldRef:
+                fieldPath: status.hostIP
+          - name: SUBM_GN_METRICS_PORT
+            value: "8081"
```
sridhargaddam: Lets use some non standard port here

submariner-operator/submariner-metrics-redesign.md
sridhargaddam: As clarified above, we can remove the changes associated with Route-Agent

submariner-operator/submariner-metrics-redesign.md
skitt: In a container spec, `$()` expands variables and is handled by Kubernetes, not the shell; see https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#use-environment-variables-to-define-arguments — so you don’t actually need the shell here:

    command: ["/usr/bin/nc"]
    args: ["-v", "-lk", "-p", "8080", "-e", "nc", "$(NODE_IP)", "$(SUBM_GW_METRICS_PORT)"]

submariner-operator/submariner-metrics-redesign.md
skitt: The main point here is to ensure that we’re creating and updating the service if necessary, not just creating it.

submariner-operator/submariner-metrics-redesign.md
mkolesnik: Good to know, thanks :smile:

submariner-operator/submariner-metrics-redesign.md
yboaron: +1 to avoid relying on node selectors

submariner-operator/submariner-metrics-redesign.md
mkolesnik: From observed behavior, it adds a `nodeAffinity` to each pod which ties it to a specific node.
This is why pods on "inactive" nodes appear as "pending".

From documentation:
> You express the topology domain (X) using a topologyKey, which is the key for the node label that the system uses to denote the domain.

And:
> For Pod affinity and anti-affinity, an empty topologyKey field is not allowed in both requiredDuringSchedulingIgnoredDuringExecution and preferredDuringSchedulingIgnoredDuringExecution.

https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,165 @@
...
+
+Submariner Gateway and Globanet pods run with HostNetworking. To access metrics exported by these two we need to open ports
+in the infra firewall. Current design has a limitation where these ports might be in use by some other service,
+causing a conflict and metrics becoming inaccessible. With the move towards minimizing changes needed in cloud infra for
+submariner, there is a use case to export metrics in a way that doesn't require any of these ports to be opened.
+
+## Proposal
+
+This enhancement proposes a redesign of existing metrics for Submariner Gateway and Globalnet by adding a new metrics proxy
+service. Pods running on the same node can access ports without requiring pods to be opened in the cloud infra. Any queries
```
sridhargaddam:
```suggestion
service. Pods running on the same node can access ports without requiring the ports to be opened in the cloud infra. Any queries
```

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -16,43 +16,46 @@ This enhancement proposes a redesign of existing metrics for Submariner Gateway
...
+* `submariner-metrics-proxy` Pod will use `netcat` to act as proxy and forward all incoming requests on metrics port to the
 metrics port on the node IP.
 * The proxy pod will run two containers if globlanet is enabled, one each for gateway and globalnet.
 * Existing metrics service for gateway and globalnet will be modified to use the new metrics proxy pod. This will be
-done by using same label selector as metrics DaemonSet.
+done by using same label selector as metrics proxy DaemonSet. This will make sure users can continue using existing services
+and ports without needing changes.
+* `submariner-gateway` will be modified to use port `32780` instead of `8080` and `submariner-globalnet` will be
+modified to use `32781` instead of `8081`. A new env variable will be used for these ports.
+* [Optional] Modify `subctl cloud prepare` to no longer open metrics ports.
```
mkolesnik, line 38: :+1: 
There will be no longer need for this, and keeping dead code will serve no purpose.

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -124,12 +141,14 @@ spec:
 
 ### Backward Compatibility
 
-Could modified gateway and globalnet metrics services be an issue? Upgrades?
-
 `subctl diagnose` checks for metrics port in firewall. This test will need to be modified to not check for port itself
 but try acessing the metrics irrespective of ports being open or not. This makes this change not backward compatible with
 older versions of `subctl`.
 
+Older versions of `subctl diagnose` will not work these changes. So users will have to make sure they're using compatible
```
mkolesnik, line 147: This is fine IMHO.

Maybe we can add a check in `diagnose` that alerts users using older cli on newer server resources, but we'll have to see how to achieve it.
In any case, it doesn't affect this EP, it's just a thought.

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,165 @@
...
+for these metrics will land on the proxy which will always run on the same node.
+
+This proposal also recommends changing the port used by metrics http servers in gateway and globalnet pods to a free range
+port to avoid problem of port being already in use on host network.
+
+## Design Details
+
+* A new DaemonSet `submariner-metrics-proxy` will be created. It will use the NodeSelector `submariner.io/gateway=true`
+to run on all gateway nodes, active or not.
+* This DaemonSet will use `schedulerName: default-scheduler` to make sure that metrics proxy pod moves with the active gateway
```
mkolesnik:
```suggestion
```

Not needed anymore

submariner-operator/submariner-metrics-redesign.md
mkolesnik: I agree with @sridhargaddam.

Although at this level we don't risk collision (as we're using the pod subnet), the proxy itself can use the same port(s) as the pod it's proxying for simplicity.

On the service we'll still be using the 8080/8081 ports.

submariner-operator/submariner-metrics-redesign.md
vthapar: I left it here as something to be release noted and/or documented.

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,163 @@
...
+
+Related Issue:
+[Exporting Submariner metrics without needing to open ports](https://github.com/submariner-io/submariner/issues/1944)
+
+## Summary
+
+Submariner Gateway and Globanet pods run with HostNetworking. To access metrics exported by these two we need to open ports
+in the infra firewall. Current design has a limitation where these ports might be in use by some other service,
+causing a conflict and metrics becoming inaccessible. With the move towards minimizing changes needed in cloud infra for
+submariner, there is a use case to export metrics in a way that doesn't require any of these ports to be opened.
```
skitt: Port conflicts are an additional concern, discuss them last so that the infrastructure considerations aren’t split up.

```suggestion
Submariner Gateway and Globanet pods run with HostNetworking.
To access metrics exported by these two we need to open ports in the infra firewall.
With the move towards minimizing changes needed in cloud infra for Submariner,
there is a use case to export metrics in a way that doesn't require any of these ports to be opened.
The current design also has a limitation where these ports might be in use by some other service,
causing a conflict and resulting in metrics becoming inaccessible. 
```

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Proposal
+
+This enhancement proposes a redesign of existing metrics for Submariner Gateway and Globalnet by adding a new metrics proxy
+service. Pods running on the same node can access ports without requiring ports to be opened in the cloud infra. Any queries
+for these metrics will land on the proxy which will always run on the same node.
```
skitt:
```suggestion
service. Pods running on the same node can access ports without requiring ports to be opened in the cloud infra.
Queries for metrics will reference the proxy service instead of the Gateway and Globalnet metrics ports directly,
and each proxy will provide access to the metrics on the same node.
```

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,163 @@
...
+
+This proposal also recommends changing the port used by metrics http servers in gateway and globalnet pods to a free range
+port to avoid problem of port being already in use on host network.
```
skitt:
```suggestion
This proposal also recommends changing the ports used by the metrics HTTP servers in the Gateway and Globalnet pods to a free range
port to avoid conflicts with ports already in use on the host network.
```

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Design Details
+
+* A new DaemonSet `submariner-metrics-proxy` will be created. It will use the NodeSelector `submariner.io/gateway=true`
+to run on all gateway nodes, active or not.
+* DaemonSet will also use toleration to make sure it can always be scheduled on gateway node, irrespective of any taints.
```
skitt:
```suggestion
* The DaemonSet will also use tolerations to make sure it can always be scheduled on the gateway node, irrespective of any taints.
```

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,163 @@
...
+* `submariner-metrics-proxy` Pod will use `netcat` to act as proxy and forward all incoming requests on metrics port to the
```
skitt:
```suggestion
* The `submariner-metrics-proxy` Pod will use `netcat` to act as proxy and forward all incoming requests on metrics service port to the
```

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,163 @@
...
+metrics port on the node IP.
+* The proxy pod will run two containers if globlanet is enabled, one each for gateway and globalnet.
```
skitt:
```suggestion
* The proxy pod will run two containers if Globalnet is enabled: one each for Gateway and Globalnet.
```

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,163 @@
...
+* Existing metrics service for gateway and globalnet will be modified to use the new metrics proxy pod. This will be
+done by using same label selector as metrics proxy DaemonSet. This will make sure users can continue using existing services
+and ports without needing changes.
```
skitt:
```suggestion
* The existing metrics service for Gateway and Globalnet will be modified to use the new metrics proxy pod. This will be
done by using same label selector as the metrics proxy DaemonSet. This will make sure users can continue using existing services
and ports without needing to make any changes themselves.
```

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,163 @@
...
+...
+spec:
+  selector:
+-   app: submariner-globalnet
++   app: submariner-metrics-proxy
+```
+
+### Additional Details
+
+These changes will be done in following order to avoid any breakage in CI:
```
skitt:
```suggestion
These changes will be done in the following order to avoid any breakage in CI:
```

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,163 @@
...
+
+1. Modifying `subctl diagnose` to actually query metrics than check for metrics port availability on firewall.
```
skitt:
```suggestion
1. Modify `subctl diagnose` to actually query metrics rather than check for metrics port availability on firewall.
```

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,163 @@
...
+2. Create and deploy the new metrics proxy service. As part of this change gateway and globalnet
+daemonset will be modified to set the metrics port in an env variable. It will still use the existing ports `8080`
+and `8081`. This change will be in `submariner-operator` repo.
```
skitt:
```suggestion
2. Create and deploy the new metrics proxy service. As part of this change the Gateway and Globalnet
DaemonSets will be modified to set the metrics port in an env variable. They will still use the current ports, `8080`
and `8081`. This change will be made in the `submariner-operator` repo.
```

submariner-operator/submariner-metrics-redesign.md
```diff
@@ -0,0 +1,163 @@
...
+3. Modify gateway and globalnet code to use the new metrics port env variable when creating metrics http server. If not
+available, it will default to the new values. This change will be in `submariner` repo.
```
skitt:
```suggestion
3. Modify gateway and globalnet code to use the new metrics port env variable when creating their metrics HTTP servers. If not
available, it will default to the new values. This change will be in the `submariner` repo.
```

skitt (changes requested):

skitt (approved):

submariner-operator/submariner-metrics-redesign.md
sridhargaddam: >Maybe we can add a check in diagnose that alerts users using older cli on newer server resources, but we'll have to see how to achieve it.

Yes, we can do that. We can add a small check in the older diagnose cli to check for the presence of the metrics daemonSet deployment (or the env' variable SUBM_GW_METRICS_PORT) and alert the user accordingly.

sridhargaddam (approved):

mkolesnik (approved):

dfarrell07:
We're hitting the issue where the SHA at HEAD has been tested too many times and the Check Dependencies job is failing with:

> HttpError: Validation Failed: {"resource":"Status","code":"custom","message":"This SHA and context has reached the maximum number of statuses."}
    at /home/runner/work/_actions/z0al/dependent-issues/0fae07162bc9e0d8e116a133bd03686eed6efa21/webpack:/dependent-issues/node_modules/@octokit/request/dist-node/index.js:86:1
    at processTicksAndRejections (internal/process/task_queues.js:97:5)

https://github.com/submariner-io/enhancements/actions/runs/2872203836

PR #134, opened 2022-09-15
Proposal for `subctl diagnose pre-install`

mkolesnik:
This proposes a new command to validate the infrastructure before installing Submariner.

Signed-off-by: Mike Kolesnik <mkolesni@redhat.com>


subctl/diagnore-pre-install.md
```diff
@@ -0,0 +1,102 @@
...
+
+Add a `subctl diagnose pre-install` command which checks that the infrastructure supports installing Submariner by checking:
+
+* Gateway nodes are set-up.
+  * Kube proxy, if there, doesn't run in IPVS mode - as it's currently not supported.
+* The broker is reachable from the cluster.
+* K8s version is supported.
+* The CNI is supported.
+* Conflicts with existing clusters in the clusterset.
+  * CIDRs should be distinct, unless the clusterset is using globalnet.
```
nyechiel: Perhaps I am missing something, but at the point where the tool is going to be used there isn't a clusterset...?

subctl/diagnore-pre-install.md
```diff
@@ -0,0 +1,102 @@
...
+* The necessary ports are open.
+  * On the gateway, when not using Load Balancer, the cable driver port(s).
```
nyechiel: Again, at the point where the tool is going to be used Submariner is not deployed yet. How can we tell if the user intends to use `LoadBalancer` or not? Or are you suggesting that we assume that ROSA and ARO must use it, and all other platforms not?

subctl/diagnore-pre-install.md
mkolesnik: There could already be a clusterset, if it's not the first cluster you're adding..

In fact, the common case would be that you already have a clusterset, unless it's the first cluster you're ever adding.

Anyway I'll add clarification.

subctl/diagnore-pre-install.md
mkolesnik: We could detect it for some platforms, but for starters we can get the info from the user.

I'll add some clarification

mkolesnik (commented):
Thanks for speedy the review

subctl/diagnore-pre-install.md
nyechiel: The number one use case I was thinking of for this tool is greenfield deployments of Submariner. I can see your point as well.

nyechiel: The number one use case I was thinking of for this tool is greenfield deployments of Submariner. I can see your point as well.

mkolesnik: From a workflow perspective, though, you'd deploy the broker before the clusters, so you'd still have an empty clusterset, and as you may know from set theory, an empty set is still a set ;)

subctl/diagnore-pre-install.md
sridhargaddam: Along the same lines, the ports used for tunnels, nat-discovery can be configured during deployment. So we should provide an option to get this info from the user.

subctl/diagnore-pre-install.md
```diff
@@ -0,0 +1,113 @@
...
+  * Query K8s and check we have a minimal supported version (1.17).
+* CNI (Reuse if possible with refactor)
+  * Check if CNI as detected by us is supported (from a list of plugins).
+    * Reuse submariner-operator/pkg/discovery/network -> Discover
+  * Warn that Submariner might not work if the CNI not detected (but don't fail the check).
+  * Extra checks for:
+    * OVN-K:
+      * Version of NB DB is supported.
+    * Calico:
+      * IP Pools used on node labeled as gateway are disabled.
```
sridhargaddam:
```suggestion
      * IP Pools are configured in the cluster to instruct Calico not to perform NAT on the subnets associated with the Pod and Service CIDRs of the remote clusters.
```

subctl/diagnore-pre-install.md
```diff
@@ -0,0 +1,113 @@
...
+  * Make sure the broker is reachable from the cluster.
+* K8s Version (Reuse as is)
+  * Query K8s and check we have a minimal supported version (1.17).
+* CNI (Reuse if possible with refactor)
+  * Check if CNI as detected by us is supported (from a list of plugins).
+    * Reuse submariner-operator/pkg/discovery/network -> Discover
+  * Warn that Submariner might not work if the CNI not detected (but don't fail the check).
+  * Extra checks for:
+    * OVN-K:
+      * Version of NB DB is supported.
```
sridhargaddam:
```suggestion
      * Version of northbound database is supported.
```

subctl/diagnore-pre-install.md
```diff
@@ -0,0 +1,113 @@
...
+    * Skipped when deploying with Load Balancer
+    * Quite problematic to reuse as we don't necessarily have the public IP, we would need to figure out how to deduce it.
+    * Also the existing diagnose check requires a second cluster...
+      * Although we could try to initiate from the same cluster to an external IP, but the success depends on the routing policy.
+    * Spawn on a gateway labeled node on the local cluster a pod running
+      `tcpdump -ln -Q in -A -s 100 -i any udp and dst port <dest port> | grep '<client message>'`
+      where the client message is a random string (part of UUID).
+    * Spawn on a non gateway node on the local cluster running
+      `echo <client message> | for i in $(seq 5); do timeout 2 nc -n -p 9898 -u <gateway ip> <dest port>; done`
+      where the gateway ip is the public IP of the pod running the "sniffer"
```
sridhargaddam: Unfortunately, assuming that gateway-ip is the public-ip does not work all the time. In particular, it will not work for onPrem clusters which are behind the same Router and will also not work for disconnected/air-gapped clusters.

subctl/diagnore-pre-install.md
```diff
@@ -0,0 +1,113 @@
...
+
+We can reuse existing code in `subctl diagnose` commands as is, or refactor as necessary:
+
+* Gateway node (New check)
+  * Check that we have at least one node labeled as gateway.
+  * This check is fundamental for firewall checking.
+    * Unless we're changing deployment model not to use labeled gateway, in which case those checks will have to be updated anyhow.
+  * Kube Proxy Mode (Reuse as is)
+    * Run "ip a s kube-ipvs0" on a gateway node and check that the interface isn't there.
+    * "The cluster is deployed with kube-proxy ipvs mode which Submariner does not support"
```
sridhargaddam:
```suggestion
* Kube Proxy Mode (Reuse as is)
  * Run "ip a s kube-ipvs0" on a gateway node and check that the interface isn't there.
  * "The cluster is deployed with kube-proxy ipvs mode which Submariner does not support"
```

subctl/diagnore-pre-install.md
```diff
@@ -0,0 +1,113 @@
...
+
+We would have more code to maintain (although sharing with existing diagnose commands could minimize the impact).
+
+### Backward Compatibility
+
+As this is a new command, no impact on existing behavior is expected.
+
+### Alternatives
+
+We could simply focus our efforts on expanding `subctl diagnose` to cover more cases instead, at the cost of forcing users to first install
```
sridhargaddam: This is not a bad idea. If we do not have a fool proof solution to validate the firewall ports, IMO the benefit of this additional command might be somewhat limited because the other commands like K8s version, CNI, Gateway node count etc are mostly well documented.

In fact, we can enhance the `subctl diagnose` to validate the possible checks depending on the state of Submariner deployment (i.e., allow the user to run the same command even when Submariner is not deployed). 
If Submariner is not installed, it will validate whatever is technically feasible based on the available information from the BrokerInfo/user-input. OTOH, if Submariner is installed, it can validate much more than the pre-requisites captured in this EP - like firewall ports, service-discovery objects, state of SM deployment, globalnet objects etc.

Thoughts?

sridhargaddam (changes requested):

subctl/diagnore-pre-install.md
```diff
@@ -0,0 +1,113 @@
...
+* Conflicts with existing clusters in the clusterset.
+  * CIDRs should be distinct, unless the clusterset is using globalnet.
+* The necessary ports are open.
+  * On the gateway, when not using Load Balancer, the cable driver port(s).
+  * On the nodes, when not using OVN, the vxlan port.
+
+The broker/clusterset checks could be optional and run only if a broker file is supplied.
+
+Alternatively, instead of adding a new command under the `diagnose` commands, these checks could be done as part of`subctl join`.
+This would make sense, since at the point of running the command, we need some information from the user:
```
tpantelis: I think this makes sense. Perhaps it's possible a user may want to just check prerequisites w/o actually joining (yet) in  which case we could have an option, eg "check-prerequisites-only".

subctl/diagnore-pre-install.md
```diff
@@ -0,0 +1,113 @@
...
+
+As a user of Submariner, it would be useful in a myriad of situations to have a tool capable of testing the Kubernetes infrastructure
+where Submariner is to be installed.
+The tool should check all of Submariner's pre-requisites from the infrastructure, and report any deficiencies.
+It would be nice if the tool can also offer guidance for the deficiencies, so that the user can address them and install Submariner successfully.
+
+This tool would compliment other `subctl diagnose` commands which check that Submariner is installed and working properly, by allowing
+similar checks **before** the actual installation.
+
+Alternatively, the new command could be hosted elsewhere, if nesting under `diagnose` is deemed confusing.
```
tpantelis: I think it should be elsewhere, ie "diagnose" connotes troubleshooting after installation.

subctl/diagnore-pre-install.md
dfarrell07: +1 that this makes sense as a part of join with an option to not actually join

subctl/diagnore-pre-install.md
```diff
@@ -0,0 +1,113 @@
...
+The broker/clusterset checks could be optional and run only if a broker file is supplied.
+
+Alternatively, instead of adding a new command under the `diagnose` commands, these checks could be done as part of`subctl join`.
+This would make sense, since at the point of running the command, we need some information from the user:
+
+* Broker info file.
+* Cable driver ports.
+* Is load balancer being deployed.
+
+Some of these checks can always be done as part of `join`, while others can be done only when using a flag, e.g. `--check-infra`.
```
dfarrell07: Why wouldn't we want to do as many checks as a part of default-join as possible? I guess it would waste a bit of time. But always running as many prereq validations as possible seems more reliable and user-friendly to me than hoping the user runs a seperate check. We can also expose all the pre-install checks without join, but I wonder why some checks wouldn't be a part of join.

sridhargaddam: In the absence of a fool proof way to check the firewall ports, I would prefer to have these checks as part of "join" operation or as part of existing "diagnose" (with enhancements to validate whatever is possible before Submariner is deployed).

Currently, we are validating the following things at the time of join
https://github.com/submariner-io/subctl/blob/a13c11a5b11ec767cdad53b98e1fc70d61402b11/pkg/join/join.go#L43

1. Min K8s Version
2. Valid customCoreDNSConfig
3. Overlapping GlobalnetCIDRs when user explicitly supplies a globalCIDR

subctl/diagnore-pre-install.md
mkolesnik: We can handle air-gapped clusters when we have proper support for that.

As for on prem clusters behind NAT, what do you propose?
Won't they fail the `subctl diagnose firewall` check as well in this case?

subctl/diagnore-pre-install.md
mkolesnik: It seems there's more support to add all these checks as part of `join`, hence I'm updating the EP to reflect this direction.

mkolesnik:
Since this pivoted to being part of join, converted to an issue - https://github.com/submariner-io/submariner/issues/2176

Issue #143, opened 2022-06-10
[EPIC] Option for recovering the broker-info.subm file from a running broker cluster


josecastillolema:
**Epic Description**

**What would you like to be added**:
It would be nice to have a `subctl` subcommand (or a well defined procedure) to recover the `broker-info.subm` from a running broker cluster.

**Why is this needed**:
There may be situations were the `broker-info.subm` file gets lost over time.

**Acceptance Criteria**
`broker-info.subm` file can be successfully recovered from the user specified environment.

## Definition of Done (Checklist)

* [x] Code complete
* [ ] Relevant metrics added
* [ ] The acceptance criteria met
* [x] Unit/e2e test added & pass
* [x] CI jobs pass
* [ ] Deployed using cloud-prepare+subctl
* [ ] Deployed using ACM/OCM addon
* [ ] Deploy using Helm
* [ ] Deployed on supported platforms (for e.g kind, OCP on AWS, OCP on GCP)
* [ ] Run subctl verify, diagnose and gather
* [ ] Uninstall
* [ ] Troubleshooting (gather/diagnose) added
* [x] Documentation added
* [x] Release notes added

## Work Items

- [x] Add a new `subctl` command to recover `broker-info.subm`
- [x] Add logic in `pkg/broker` to fetch the data
- [x] Add system tests
- [x] Add release notes
- [x] Add explanation in the website

##  PRs
Depends on https://github.com/submariner-io/subctl/pull/467
Depends on https://github.com/submariner-io/subctl/pull/588
Depends on https://github.com/submariner-io/subctl/pull/621
Depends on https://github.com/submariner-io/subctl/pull/630
Depends on https://github.com/submariner-io/submariner-website/pull/910
Depends on https://github.com/submariner-io/enhancements/pull/173
Depends on https://github.com/submariner-io/subctl/pull/649

nyechiel:
@Jaanki any reason to keep this open or is it fully done now?

PR #144, opened 2022-11-23
Proposal to recover broker-info.subm file from running broker

Jaanki:
Epic: https://github.com/submariner-io/enhancements/issues/143

Signed-off-by: Janki Chhatbar <jchhatba@redhat.com>

### DIFF
--- a/subctl/issue-89-reconstructing-broker-info.md
+++ b/subctl/issue-89-reconstructing-broker-info.md
@@ -1,0 +1,71 @@
+# Recover broker-info.subm file from the running Broker
+
+Related Issue:
+[Option for recovering the broker-info.subm file from a running broker cluster](https://github.com/submariner-io/enhancements/issues/143)
+
+## Summary
+
+`broker-info.subm` file is created when a broker is installed. The file contains information critical to have a functional environment.
+There is a possibility that the file could be deleted or lost for one or the other reason. In such cases, no new clusters could be joined.
+
+This proposal describes a way to recover the `broker-info.subm` file from the running broker.
+
+## Proposal
+
+A new `subctl` command will be added that would fetch needed data from the broker and write it to `broker-info.subm` file. The file
+would be Base64 encoded.
+
+## Design details
+
+* subctl command name: `recover-brokerinfo`
+* Flags for the command:
+  * `brokerconfig` to specify kubeconfig of the cluster in which Broker is installed
+  * `brokernamespace` to specify the namespace in which broker is installed
+  * `ipsec-psk-from` to specify the file to read PSK secret from
+  * All flags that `restConfigProducer` provides
+    * `kubeconfig` to specify kubeconfig of the cluster in which Submariner is installed
+* Data to be fetched from running broker:
+  * Broker URL
+  * Client token
+  * IPSec PSK
+  * Components installed i.e. Submariner and Service discovery
+  * whether service discovery component is installed
+  * default custom domains
+
+The last three information will be retrieved from the broker `Spec`, Broker URL from the user provided broker kubeconfig, client token
+from the Broker admin service account `submariner-k8s-broker-admin` and IPSec PSK with the following logic:
+
+```mermaid
+flowchart TD
+  A{Submariner config provided via `--kubeconfig` flag?} -- Yes ---> D
+  A -- No ---> B
+  B[Broker kubeconfig provided via `--brokerconfig` flag] ---> C{Broker cluster has Submariner installed?}
+  C -- Yes ---> D[Get PSK secret from submariner.Spec]
+  C -- No ---> E{"Any cluster joined to Broker (Figure out from the endpoints)?"}
+  E -- Yes ---> F[Exit]
+  E -- No ---> G{PSK file specified via `--ipsec-psk-from` flag?}
+  G -- Yes ---> H[Use secret from the file]
+  G -- No --> I[Generate new secret]
+```
+
+If Submariner is not installed in the cluster specified by the `--kubeconfig` flag, end the execution and throw an error. This addresses
+the scenario where there is a cluster with Submariner in the environment but an incorrect config was provided via the flag. In such
+scenario, if the execution is not stopped, a different/new PSK secret would be populated which would break the environment.
+
+### Usage
+
+The new command can be used as follows (assuming Submariner is installed in cluster2:
+
+`subctl recover-brokerinfo --brokerconfig output/kubeconfigs/kind-config-cluster1 --kubeconfig output/kubeconfigs/kind-config-cluster2`
+
+### Alternatives
+
+None that is automated. User can always manually keep a copy of `broker-info.subm` file.
+
+## External Dependencies
+
+None.
+
+## User Impact
+
+Users would be able to retrieve information in cases where generated `broker-info.subm` file is lost.


subctl/issue-89-reconstructing-broker-info.md
```diff
@@ -0,0 +1,62 @@
...
+
+The last three information will be retrieved from the broker `Spec`, Broker URL from the user provided kubeconfig, client token from the 
+Broker admin service account `submariner-k8s-broker-admin` and IPSec PSK with the following logic:
+
+```
+if file specified via `ipsec-psk-from` flag by the user
+  read the secret from the specified file
+else if submariner is installed in the cluster
+  read the secret from submariner spec
+else
```
skitt: There’s a missing case here: if the user didn’t provide a file, and Submariner isn’t installed (or isn’t accessible using the available credentials), but the broker has one or more connected clusters, `subctl` should fail with an error — a `broker-info.subm` with a new PSK would not allow newly-connected clusters to connect to existing clusters.

skitt (changes requested):

subctl/issue-89-reconstructing-broker-info.md
```diff
@@ -0,0 +1,62 @@
...
+This proposal describes a way to reconstruct the `broker-info.subm` file from the running broker.
+
+## Proposal
+
+A new `subctl` command will be added that would fetch needed data from the broker and write it to `broker-info.subm` file. The file 
+would be Base64 encoded.
+
+## Design details
+
+* subctl command name: `reconstruct-brokerinfo`.
```
tpantelis:
```suggestion
* subctl command name: `recover-broker-info`.
```

tpantelis (changes requested):
Why can't the user just re-run `deploy-broker`?

subctl/issue-89-reconstructing-broker-info.md
Jaanki: Done.

subctl/issue-89-reconstructing-broker-info.md
Jaanki: Well, as per definitions in this [wikidiff](https://wikidiff.com/reconstruct/recover), we are constructing a **new file** with needed info. We are not getting the **same file** which was lost.

subctl/issue-89-reconstructing-broker-info.md
```diff
@@ -0,0 +1,69 @@
...
+## Proposal
+
+A new `subctl` command will be added that would fetch needed data from the broker and write it to `broker-info.subm` file. The file
+would be Base64 encoded.
+
+## Design details
+
+* subctl command name: `reconstruct-brokerinfo`.
+* Flags for the command:
+  * `submconfig` to specify the cluster in which Submariner is installed
```
sridhargaddam: Not sure I understand this. What exactly is this `submconfig`? and how is it related to `kubeconfig`? 
Does `submconfig` mean - SubmarinerConfig which is only applicable in OCM/ACM deployments?

Jaanki: @sridhargaddam its a kubeconfig of a cluster which has Submariner installed.

subctl/issue-89-reconstructing-broker-info.md
```diff
@@ -0,0 +1,69 @@
...
+## Proposal
+
+A new `subctl` command will be added that would fetch needed data from the broker and write it to `broker-info.subm` file. The file
+would be Base64 encoded.
+
+## Design details
+
+* subctl command name: `reconstruct-brokerinfo`.
+* Flags for the command:
+  * `submconfig` to specify kubeconfig of the cluster in which Submariner is installed
```
Jaanki: rename this name. Suggestions?

Jaanki: Address the case where the cluster has both broker and Submariner installed so no need to provide 2 kubeconfigs.

subctl/issue-89-reconstructing-broker-info.md
Jaanki: `submconfig` is an apt name for this flag. And anyways this is very subctl specific and only all Submariner users are OCM users.

subctl/issue-89-reconstructing-broker-info.md
Jaanki: Done.

subctl/issue-89-reconstructing-broker-info.md
Jaanki: Done

subctl/issue-89-reconstructing-broker-info.md
```diff
@@ -0,0 +1,70 @@
...
+  * Client token
+  * IPSec PSK
+  * Components installed i.e. Submariner and Service discovery
+  * whether service discovery component is installed
+  * default custom domains
+
+The last three information will be retrieved from the broker `Spec`, Broker URL from the user provided kubeconfig, client token from the
+Broker admin service account `submariner-k8s-broker-admin` and IPSec PSK with the following logic:
+
+```mermaid
```
Jaanki, line 38: Review this under `rich diff` format (select from the top right button) to see the flowchart

subctl/issue-89-reconstructing-broker-info.md
```diff
@@ -0,0 +1,70 @@
...
+## Proposal
+
+A new `subctl` command will be added that would fetch needed data from the broker and write it to `broker-info.subm` file. The file
+would be Base64 encoded.
+
+## Design details
+
+* subctl command name: `recover-brokerinfo`
+* Flags for the command:
+  * `submconfig` to specify kubeconfig of the cluster in which Submariner is installed
```
sridhargaddam: As mentioned during the call, I'm personally not convinced that `submConfig` is an apt name as we use it in ACM/OCM. Please come up with a better name.

sridhargaddam (changes requested):

subctl/issue-89-reconstructing-broker-info.md
tpantelis: I think it should be the normal `--context`. As discussed,

- First check If the `Submariner` (or `ServiceDiscovery`) resource exists, if so proceed with recovery
- If not then check if the broker is installed (see [findBrokerNamespace](https://github.com/submariner-io/subctl/blob/3d44295393e4ab4ca35976030e93aa194a3e2c3d/pkg/uninstall/all.go#L324))
- If broker installed, then check if it's in use (ie check if there are any `Endpoints` in the broker NS)
- If not in use then generate a new PSK
- If in use then exit and inform the user they must provide the kubeconfig of a joined cluster

tpantelis (dismissed):

subctl/issue-89-reconstructing-broker-info.md
skitt: That only works if we can recover everything using _only_ a `Submariner` resource; otherwise we need to be able to specify two contexts (one to access a broker, the other to access a joined cluster).

tpantelis: The `Submariner` resource contains the connectivity info to access the broker, if needed.

Jaanki: So, the original kubeconfig provided should be of Broker cluster. And if Submariner is not installed on that cluster, and there is a joined cluster, provide kubeconfig of the cluster that has Submariner installed. Right?

tpantelis: As outlined in the steps I posted above, assuming we need the `Submariner` resource from a previously joined cluster in order to recover the broker info, then we expect the user to provide the kubeconfig of a joined cluster. If not and it's the broker cluster then we can try to recover iff there are no joined clusters.

Jaanki: Client token is stored only on `broker`. So access to broker (ad hence its kubeconfig) is necessary. Unless we get broker restconfig from Submariner/service discovery like [gather does](https://github.com/submariner-io/subctl/blob/devel/internal/gather/gather.go#L183)

tpantelis: > Unless we get broker restconfig from Submariner/service discovery like [gather does](https://github.com/submariner-io/subctl/blob/devel/internal/gather/gather.go#L183)

Right - that's what I said above, also like the pods do.

Jaanki: Turns out serviceaccounts are not accessible from the Submariner cluster.
`error getting client token ServiceAccount submariner-k8s-broker-admin get failed: serviceaccounts "submariner-k8s-broker-admin" is forbidden: cannot get resource "serviceaccounts" in API group "" in the namespace "submariner-k8s-broker"`

tpantelis: So the broker credentials in the `Submariner` object used by the pods to sync with the broker NS are associated with the "_submariner-k8s-broker-cluster_" `Role` defined [here](https://github.com/submariner-io/subctl/blob/98f65ae7e95e0f5ebe695545acd03d4e815a2cb5/pkg/broker/rbac.go#L82).  So the `Role` doesn’t allow access to SA’s although it does allow access to `Secrets`. You need the `Secret` associated with the “_submariner-k8s-broker-admin_” SA but I guess the only way to get the `Secret` name is [via the SA](https://github.com/submariner-io/subctl/blob/98f65ae7e95e0f5ebe695545acd03d4e815a2cb5/internal/rbac/rbac.go#L36). 

Unless we want to grant the "_submariner-k8s-broker-cluster_" `Role` access to SA's then we do need the user to pass the broker kubeconfig but I don't see this as a compelling reason to expand the privileges. The kubeconfig of a joined cluster would be needed as well except if there's no joined clusters.

tpantelis: Back to the original comment, it seems we've deprecated the passing of kubeconfigs, so how about `--context` for the joined cluster and `--broker-context`.

skitt: We haven’t deprecated the passing of kubeconfigs, we’ve made them consistent; so `--kubeconfig` and `--context` for the joined cluster, `--brokerconfig` and `--brokercontext` for the broker (following existing practice).

subctl/issue-89-reconstructing-broker-info.md
```diff
@@ -0,0 +1,71 @@
...
+  B[Broker kubeconfig provided via `--brokerconfig` flag] ---> C{Broker cluster has Submariner installed?}
+  C -- Yes ---> D[Get PSK secret from submariner.Spec]
+  C -- No ---> E{"Any cluster joined to Broker (Figure out from the endpoints)?"}
+  E -- Yes ---> F[Exit]
+  E -- No ---> G{PSK file specified via `--ipsec-psk-from` flag?}
+  G -- Yes ---> H[Use secret from the file]
+  G -- No --> I[Generate new secret]
+```
+
+If Submariner is not installed in the cluster specified by `--config` flag, exit the execution and throw an error. This addresses
```
skitt:
```suggestion
If Submariner is not installed in the cluster specified by the `--kubeconfig` flag, end the execution and throw an error. This addresses
```

subctl/issue-89-reconstructing-broker-info.md
```diff
@@ -0,0 +1,71 @@
...
+the scenario where there is a cluster with Submariner in the environment but an incorrect config was provided via the flag. In such
+scenario, if the execution is not stopped, a different/new PSK secret would be populated which would break the environment.
+
+### Usage
+
+The new command can be used as follows (assuming Submariner is installed in cluster2:
+
+`subctl recover-brokerinfo --brokerkubeconfig output/kubeconfigs/kind-config-cluster1 --kubeconfig output/kubeconfigs/kind-config-cluster2`
```
skitt:
```suggestion
`subctl recover-brokerinfo --brokerconfig output/kubeconfigs/kind-config-cluster1 --kubeconfig output/kubeconfigs/kind-config-cluster2`
```

subctl/issue-89-reconstructing-broker-info.md
```diff
@@ -0,0 +1,71 @@
...
+    * `kubeconfig` to specify kubeconfig of the cluster in which Submariner is installed
+* Data to be fetched from running broker:
+  * Broker URL
+  * Client token
+  * IPSec PSK
+  * Components installed i.e. Submariner and Service discovery
+  * whether service discovery component is installed
+  * default custom domains
+
+The last three information will be retrieved from the broker `Spec`, Broker URL from the user provided broker kubeconfig, client token
```
skitt:
```suggestion
The last three information will be retrieved from the broker `Spec`, Broker URL from the user provided broker kubeconfig, client token
```

skitt (changes requested):

skitt (approved):

subctl/issue-89-reconstructing-broker-info.md
```diff
@@ -0,0 +1,71 @@
...
+  G -- No --> I[Generate new secret]
+```
+
+If Submariner is not installed in the cluster specified by the `--kubeconfig` flag, end the execution and throw an error. This addresses
+the scenario where there is a cluster with Submariner in the environment but an incorrect config was provided via the flag. In such
+scenario, if the execution is not stopped, a different/new PSK secret would be populated which would break the environment.
+
+### Usage
+
+The new command can be used as follows (assuming Submariner is installed in cluster2:
```
sridhargaddam, line 57:
```suggestion
The new command can be used as follows (assuming Submariner is installed in cluster2):
```

subctl/issue-89-reconstructing-broker-info.md
```diff
@@ -0,0 +1,71 @@
...
+This proposal describes a way to recover the `broker-info.subm` file from the running broker.
+
+## Proposal
+
+A new `subctl` command will be added that would fetch needed data from the broker and write it to `broker-info.subm` file. The file
+would be Base64 encoded.
+
+## Design details
+
+* subctl command name: `recover-brokerinfo`
```
sridhargaddam, line 20: Can you please highlight/show the mandatory flags and the optional flags?

I guess the mandatory flags are `brokerconfig` and all other flags are optional.

subctl/issue-89-reconstructing-broker-info.md
```diff
@@ -0,0 +1,71 @@
...
+
+If Submariner is not installed in the cluster specified by the `--kubeconfig` flag, end the execution and throw an error. This addresses
+the scenario where there is a cluster with Submariner in the environment but an incorrect config was provided via the flag. In such
+scenario, if the execution is not stopped, a different/new PSK secret would be populated which would break the environment.
+
+### Usage
+
+The new command can be used as follows (assuming Submariner is installed in cluster2:
+
+`subctl recover-brokerinfo --brokerconfig output/kubeconfigs/kind-config-cluster1 --kubeconfig output/kubeconfigs/kind-config-cluster2`
```
sridhargaddam, line 59: Please add another example, where the cluster1 is used both as Broker and Submariner cluster.

subctl/issue-89-reconstructing-broker-info.md
tpantelis: Neither necessarily has to be mandatory. `--kubeconfig` points to a possible `Submariner` cluster and `--brokerconfig` points to the possible broker cluster. If only `--kubeconfig` is specified and it also has the broker installed then we can proceed w/o needing `--brokerconfig`. Similarly, if only `--brokerconfig` is specified and it also has `Submariner` installed then we can proceed.

tpantelis (approved):

subctl/issue-89-reconstructing-broker-info.md
sridhargaddam: Right, it was not clear in the description. But anyways, if this is how its implemented, its good.

subctl/issue-89-reconstructing-broker-info.md
sridhargaddam: Small nit, if you happen to re-spin, otherwise its fine.

sridhargaddam (approved):

Issue #149, opened 2022-12-12
Add support for nodeSelector and toleration scheduling for submariner pods


vthapar:
## Epic Description
Submariner deploys many different agents as part of its core deployment. Except for Gateway and globalnet agent, that are deployed on designated gateway nodes, there is no way for users to control or limit where rest of the agent pods are deployed. Some customers want to run these agents on specific nodes for billing or networking reasons.

## Acceptance Criteria
User will have option(s) to specify nodeSelectors and tolerations configurations for the following:

1. Lighthouse agent
2. Lighthouse CoreDNS
3. NetworkPlugin Syncer

For this intial implementation, common selectors and tolerations that apply to all these agents would suffice.

## Definition of Done (Checklist)

* [ ] Code complete
* [ ] Relevant metrics added
* [ ] The acceptance criteria met
* [ ] Unit/e2e test added & pass
* [ ] CI jobs pass
* [ ] Deployed using cloud-prepare+subctl
* [ ] Deployed using ACM/OCM addon
* [ ] Deploy using Helm
* [ ] Deployed on supported platforms (for e.g kind, OCP on AWS, OCP on GCP)
* [ ] Run subctl verify, diagnose and gather
* [ ] Uninstall
* [ ] Troubleshooting (gather/diagnose) added
* [ ] Documentation added
* [ ] Release notes added

mkolesnik:
I think for these ones this doesnt make sense:

- Submariner gateway
- Submariner globalnet
- Metrics Proxy

Gateway and globalnet are already only being deployed on nodes with label `submariner.io/gateway=true` which gives the users a lot of control already.

The metrics pods need to be deployed alongside the other pods they're proxying, so those ones don't make any sense in allowing this control either.

This issue should largely target the operator and the LH bits.

vthapar:
@mkolesnik Use case is user providing single set of nodeSelectors/Tolerations for all OCP pods 
 not just submariner pods. Considering this, what if user sets gateway label on the wrong node? Other scenario, case where we create a new GW node as part of cloud-prepare? In this case, should we set gateway label only on nodes that match the selector/toleration?

mkolesnik:
> @mkolesnik Use case is user providing single set of nodeSelectors/Tolerations for all OCP pods not just submariner pods. 

OK but I think some Subm pods behave in a very specific way.
We're more of an infra service, not a typical app, so expecting "all pods" is just not a realistic scenario.

[...]

PR #156, opened 2023-01-10
Add support for configurable node scheduling

vthapar:
Add option for users to configure where submariner agent pods get scheduled using ndoe selectors and tolerations.

Epic: https://github.com/submariner-io/enhancements/issues/149

### DIFF
--- a/submariner-operator/issue-149-node-placement.md
+++ b/submariner-operator/issue-149-node-placement.md
@@ -1,0 +1,90 @@
+# Node Placement
+
+[Add support for nodeSelector and toleration scheduling for submariner pods](https://github.com/submariner-io/enhancements/issues/149)
+
+## Summary
+
+Submariner deploys many different agents as part of its core deployment. Gateway, Globalnet Agent and
+Metrics Proxy are deployed on designated gateway nodes, while Route Agents are deployed on all nodes. There is no way
+for users to control or limit where the rest of the agent pods are deployed. Some users want to run these agents on
+specific nodes for billing or networking reasons. Users can't add tolerations directly to these agents because they'll be
+overwritten by the operator. So, users require an easily configurable way to specify these parameters.
+
+## Proposal
+
+This enhancement proposes adding a new field to `Submariner` and `ServiceDiscovery` for users to configure node
+scheduling. This configuration will then be applied to pods of the following submariner agents:
+
+* submariner-operator
+* submariner-lighthouse-agent
+* submariner-lighthouse-coredns
+* submariner-networkplugin-syncer
+
+Following agents will be excluded because these are run on Gateway nodes. Users already control which nodes get labeled
+as gateways and ideally we don't want to run anything else on these nodes.
+
+* submariner-gateway
+* submariner-globalnet
+* submariner-metrics-proxy
+
+## Design Details
+
+Add new `NodeSelector` and `Tolerations` fields to `Submariner` and `ServiceDiscovery` specs.
+
+```Go
+type SubmarinerSpec struct {
+
+    // NodeSelector defines which Nodes the Pods are scheduled on. The default is an empty list.
+    // +optional
+    NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+
+    // Tolerations is attached by pods to tolerate any taint that matches
+    // the triple <key,value,effect> using the matching operator <operator>.
+    // The default is an empty list.
+    // +optional
+    Tolerations []v1.Toleration `json:"tolerations,omitempty"`
+}
+```
+
+```Go
+type SubmarinerDiscoverySpec struct {
+
+    // NodeSelector defines which Nodes the Pods are scheduled on. The default is an empty list.
+    // +optional
+    NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+
+    // Tolerations is attached by pods to tolerate any taint that matches
+    // the triple <key,value,effect> using the matching operator <operator>.
+    // The default is an empty list.
+    // +optional
+    Tolerations []v1.Toleration `json:"tolerations,omitempty"`
+}
+```
+
+Users can add these fields to CR manually through `kubectl`. Operator code will copy over these values to spec for
+applicable Pods.
+
+### Pros
+
+* Provides easy way for users to configure node scheduling
+
+### Cons
+
+* None
+
+### Backward Compatibility
+
+None
+
+### Alternatives
+
+* One alternative is to provide means to add tolerations and node selectors to agent pods directly without being
+those being overwritten by operator. This was rejected because it breaks the operator design philosophy.
+
+## External Dependencies
+
+None
+
+## User Impact
+
+None


submariner-operator/issue-149-node-placement.md
```diff
@@ -0,0 +1,94 @@
+# Node Placement
+
+[Add support for nodeSelector and toleration scheduling for submariner pods](https://github.com/submariner-io/enhancements/issues/149)
+
+## Summary
+
+Submariner deploys many different agents as part of its core deployment. Except for Gateway, globalnet agent and
+metrics proxy, which are deployed on designated gateway nodes, there is no way for users to control or limit where rest
```
skitt: This should mention the route agent as well, for completeness (even though it needs to be on every node).

submariner-operator/issue-149-node-placement.md
```diff
@@ -0,0 +1,94 @@
...
+### Cons
+* None
+
+### Backward Compatibility
+
+None
+
+### Alternatives
+
+None
```
skitt: Mention that users can’t currently add their own tolerations etc. since the operator always rewrites the relevant K8s objects and any user changes are lost. (So taking user changes into account would be an alternative approach.)

skitt (dismissed):

submariner-operator/issue-149-node-placement.md
```diff
@@ -0,0 +1,96 @@
+# Node Placement
+
+[Add support for nodeSelector and toleration scheduling for submariner pods](https://github.com/submariner-io/enhancements/issues/149)
+
+## Summary
+
+Submariner deploys many different agents as part of its core deployment. Gateway, Globalnet Agent and
+Metrics Proxy are deployed on designated gateway nodes, while Route Aents are deployed on all nodes. There is no way
+for users to control or limit where rest of the agent pods are deployed. Some customers want to run these agents on
+specific nodes for billing or networking reasons. Users can't add tolerations directly to these agents coz they'll be
```
mkolesnik:
```suggestion
specific nodes for billing or networking reasons. Users can't add tolerations directly to these agents because they'll be
```

nit :)

submariner-operator/issue-149-node-placement.md
```diff
@@ -0,0 +1,96 @@
...
+overwritten by the operator. So, users require an easily configurable way to specify such configurations.
+
+## Proposal
+
+This enhancement proposes adding a new field to `Submariner` and `ServiceDiscovery` for users to configure node
+scheduling. This configuration will then be applied to pods of following submariner agents:
+
+* submariner-operator
```
mkolesnik, line 18: I believe you can add the network-plugin-syncer pod to the list, IIUC it doesnt need a pecific place to run but I might be wrong

submariner-operator/issue-149-node-placement.md
```diff
@@ -0,0 +1,96 @@
...
+Following agents will be excluded because these are run on Gateway nodes. Users already control which nodes get labeled
+as gateways and ideally we don't want to run anything else on these nodes.
+
+* submariner-gateway
+* submariner-globalnet
+* submariner-metrics-proxy
+
+## Design Details
+
+Add a new `NodePlacement` field to `Submariner` and `ServiceDiscovery` specs.
```
mkolesnik: So on official [K8s PodSpec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#podspec-v1-core) these fields are directly under it, I wonder if we don't want the same here?

AFAICT we can still bundle them up in a single `struct` to avoid repetitiveness, and use the as an embedded struct.

submariner-operator/issue-149-node-placement.md
```diff
@@ -0,0 +1,96 @@
...
+
+```Go
+// NodePlacement describes node scheduling configuration for the pods.
+type NodePlacement struct {
+    // NodeSelector defines which Nodes the Pods are scheduled on. The default is an empty list.
+    // +optional
+    NodeSelector map[string]string `json:"nodeSelector,omitempty"`
```
mkolesnik, line 39: Don't we want to use the official type?

https://github.com/kubernetes/api/blob/master/core/v1/types.go#L2801

submariner-operator/issue-149-node-placement.md
```diff
@@ -0,0 +1,96 @@
...
+	// +optional
+	NodePlacement NodePlacement `json:"nodePlacement,omitempty"`
+}
+```
+
+Add option to `subctl join` to specify node selectors and tolerations.
+
+```bash
+subctl join --node-selectors LABEL_KEY_1=LABEL_VALUE_1,...,LABEL_KEY_N=LABEL_VALUE_N
+subctl join --node-tolerations KEY_1=VAL_1:TAINT_EFFECT_1,...,KEY_N=VAL_N:TAINT_EFFECT_N
```
mkolesnik: This is a very simplistic way of representing these, which can also include matchers and other info (eg toleration seconds).

I wonder if we shouldn't just let users give a yaml (string or file) for this?

mkolesnik (changes requested):

submariner-operator/issue-149-node-placement.md
vthapar: > Don't we want to use the official type?
> 
> https://github.com/kubernetes/api/blob/master/core/v1/types.go#L2801

I used https://github.com/open-cluster-management-io/enhancements/tree/main/enhancements/sig-architecture/58-addon-configuration as referece as that is where the initial use case came from. Okay to go with either, but do lean towards going with official type.

EDIT: BTW, PodSpec uses `map[string]string]` So I think this works better as makes it easier to just copy over to pod spec.

https://github.com/kubernetes/api/blob/master/core/v1/types.go#L3196

submariner-operator/issue-149-node-placement.md
vthapar: > So on official [K8s PodSpec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#podspec-v1-core) these fields are directly under it, I wonder if we don't want the same here?
> 
I didn't get what you meant by "under it", could you clarify?

submariner-operator/issue-149-node-placement.md
vthapar: Good catch, missed it.

submariner-operator/issue-149-node-placement.md
vthapar: I used `kubectl` cli for reference. Yaml/json string or file does sound like a better more flexible option, while this seems easier one.

submariner-operator/issue-149-node-placement.md
```diff
@@ -0,0 +1,97 @@
+# Node Placement
+
+[Add support for nodeSelector and toleration scheduling for submariner pods](https://github.com/submariner-io/enhancements/issues/149)
+
+## Summary
+
+Submariner deploys many different agents as part of its core deployment. Gateway, Globalnet Agent and
+Metrics Proxy are deployed on designated gateway nodes, while Route Aents are deployed on all nodes. There is no way
+for users to control or limit where rest of the agent pods are deployed. Some customers want to run these agents on
```
mkolesnik:
```suggestion
for users to control or limit where the rest of the agent pods are deployed. Some users want to run these agents on
```

submariner-operator/issue-149-node-placement.md
```diff
@@ -0,0 +1,97 @@
...
+specific nodes for billing or networking reasons. Users can't add tolerations directly to these agents because they'll be
+overwritten by the operator. So, users require an easily configurable way to specify such configurations.
```
mkolesnik:
```suggestion
overwritten by the operator. So, users require an easily configurable way to specify these parameters.
```

submariner-operator/issue-149-node-placement.md
mkolesnik: I mean they're just part of the spec (e.g. `PodSpec`) and not under another field, so to specify them you set `spec.nodeSelector` and such. You don't set `spec.nodePlacement.nodeSelector`..

So I think we can have it the same way as K8s for consistency.

submariner-operator/issue-149-node-placement.md
mkolesnik: Yeah seems the more elaborate terms are used as part of `NodeAffinity`

submariner-operator/issue-149-node-placement.md
mkolesnik: Do you have a reference to where in `kubectl` this is available? I couldn't find it

submariner-operator/issue-149-node-placement.md
vthapar: Aha, got it. Since we're adding these fields to a generic Submariner spec, so `NodePlacement` gives a better idea about what these two fields are being used for. Also coz that is what the similar change I used for reference in openshift was using. But I do see your point, Node Selector and toleration imply used for placement of pods on specific nodes. Am okay to go with it [simpler and less nesting] unless anyone has other ideas.

submariner-operator/issue-149-node-placement.md
vthapar:
```
Usage:
  kubectl taint NODE NAME KEY_1=VAL_1:TAINT_EFFECT_1 ... KEY_N=VAL_N:TAINT_EFFECT_N [options]

Usage:
  kubectl label [--overwrite] (-f FILENAME | TYPE NAME) KEY_1=VAL_1 ... KEY_N=VAL_N [--resource-version=version]
[options]
```

mkolesnik: Ok but that is for the taints, I haven't seen any example where tolerations are given on the command line, only as YAML examples (which I guess you're supposed to edit/apply/patch)

mkolesnik: @vthapar :point_up:

submariner-operator/issue-149-node-placement.md
```diff
@@ -0,0 +1,94 @@
...
+    // +optional
+    Tolerations []v1.Toleration `json:"tolerations,omitempty"`
+}
+```
+
+Add option to `subctl join` to specify node selectors and tolerations.
+
+```bash
+subctl join --node-selectors LABEL_KEY_1=LABEL_VALUE_1,...,LABEL_KEY_N=LABEL_VALUE_N
+subctl join --node-tolerations <path-to-yaml-file>
```
mkolesnik:
```suggestion
subctl join --tolerations <path-to-yaml-file>
```

Shouldn't it be just `--tolerations` then?

mkolesnik: Also if we're already talking about YAML, perhaps we can just let users specify fields via YAML directly then?

vthapar: @mkolesnik You mean allow node selector and tolerations both through yaml?

EDIT: As per discussion on slack, @mkolesnik idea is to add option for using a yaml to input all CR fields, not just tolerations and node-selector. Since such work is beyond the scope of this PR, plan is now to remove subctl changes from this EP. Users can modify the CR directly once created to apply node scheduling. A separate EP will cover adding option to `subctl` to take all CR fields through a yaml file.

mkolesnik (approved):

submariner-operator/issue-149-node-placement.md
```diff
@@ -0,0 +1,87 @@
+# Node Placement
+
+[Add support for nodeSelector and toleration scheduling for submariner pods](https://github.com/submariner-io/enhancements/issues/149)
+
+## Summary
+
+Submariner deploys many different agents as part of its core deployment. Gateway, Globalnet Agent and
+Metrics Proxy are deployed on designated gateway nodes, while Route Aents are deployed on all nodes. There is no way
```
Jaanki:
```suggestion
Metrics Proxy are deployed on designated gateway nodes, while Route Agents are deployed on all nodes. There is no way
```

submariner-operator/issue-149-node-placement.md
```diff
@@ -0,0 +1,87 @@
...
+for users to control or limit where the rest of the agent pods are deployed. Some users want to run these agents on
+specific nodes for billing or networking reasons. Users can't add tolerations directly to these agents because they'll be
+overwritten by the operator. So, users require an easily configurable way to specify these parameters.
+
+## Proposal
+
+This enhancement proposes adding a new field to `Submariner` and `ServiceDiscovery` for users to configure node
+scheduling. This configuration will then be applied to pods of following submariner agents:
```
Jaanki:
```suggestion
scheduling. This configuration will then be applied to the pods of following submariner agents:
```

submariner-operator/issue-149-node-placement.md
```diff
@@ -0,0 +1,87 @@
...
+
+* submariner-operator
+* submariner-lighthouse-agent
+* submariner-lighthouse-coredns
+* submariner-networkplugin-syncer
+
+Following agents will be excluded because these are run on Gateway nodes. Users already control which nodes get labeled
+as gateways and ideally we don't want to run anything else on these nodes.
```
Jaanki, line 24: `ideally we don't want to run anything else on these nodes.`

This line to me means that the 4 other agents (submariner-operator, submariner-lighthouse-agent, submariner-lighthouse-coredns, submariner-networkplugin-syncer) won't be running on gateway nodes. Is that true?

Jaanki:
I didn't understand how users will be able to use these tolrations and node selector fields and how these fields will be used to place the agents.

mkolesnik:
> I didn't understand how users will be able to use these tolrations and node selector fields and how these fields will be used to place the agents.

Those are standard K8s features - https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/ & https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/
The EP is about exposing them on the high level CRDs we use to manage the deployment.

submariner-operator/issue-149-node-placement.md
skitt: I would prefer

```suggestion
scheduling. This configuration will then be applied to pods of the following submariner agents:
```

skitt (approved):

submariner-operator/issue-149-node-placement.md
vthapar: This mainly refers to user's services, but we should prefer not to run anything else [LH, coredns etc.] on GW nodes to keep CPU free to tunnel data traffic. But that discussion is outside the scope of this PR. Scope of this is mainly to differentiate submariner pods from user pods. We can have a different EP for optimization, probably somethin we might have to when we do scale/perf testing.

submariner-operator/issue-149-node-placement.md
```diff
@@ -63,13 +63,6 @@ type SubmarinerDiscoverySpec struct {
 }
 ```
 
-Add option to `subctl join` to specify node selectors and tolerations.
```
Jaanki: This exactly answers my query here https://github.com/submariner-io/enhancements/pull/156#issuecomment-1446143690. Don't know why this was removed. But I guess others have understood and don't want to block this and hence approving.

Jaanki (approved):

skitt:
@vthapar I don’t know why the jobs aren’t running; could you squash your commits and push your PR again?

skitt:
> I didn't understand how users will be able to use these tolrations and node selector fields and how these fields will be used to place the agents.

The only way for users to specify these fields is to add the relevant entries to our CRs; support for this in `subctl` will be addressed in a later EP. Users are expected to specify tolerations and/or selectors, if they need them, in the CRs, and the values given there will be copied directly to the pod templates created by the operator.

@vthapar it might be worth updating the EP to clarify the above — for readers unaware of the history of the EP, and in particular the changes around `subctl`, the EP can be hard to understand!

submariner-operator/issue-149-node-placement.md
```diff
@@ -0,0 +1,87 @@
...
+    NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+
+    // Tolerations is attached by pods to tolerate any taint that matches
+    // the triple <key,value,effect> using the matching operator <operator>.
+    // The default is an empty list.
+    // +optional
+    Tolerations []v1.Toleration `json:"tolerations,omitempty"`
+}
+```
+
```
skitt, line 63: The design details should mention that the operator is going to be changed to copy these values as appropriate.

skitt (dismissed):

submariner-operator/issue-149-node-placement.md
Jaanki: Exactly the answer I was looking for.

vthapar: Done. Also added a line about user manually adding these to CR through kubectl.

vthapar:
> @vthapar I don’t know why the jobs aren’t running; could you squash your commits and push your PR again?

Done.

skitt (approved):

dfarrell07:
@vthapar Regarding your ask on the call for ideas about why CI isn't running here - I wonder if it's because we had a load of Dependabot PRs at about the same time? Maybe try again once the CI queue is less full?

PR #197, opened 2023-05-09
Proposal to build subctl with limited sub-commands

### LINKED ISSUE #196: Support disabling certain features in builds

Some builds of `subctl` (_e.g._ for OCM) don’t need certain sub-commands such as `deploy-broker`, `join` etc. It would be useful to be able to disable these features so that users don’t even see them in `--help` output.

**Why is this needed**:

See above.
Enhancement Proposal: https://github.com/submariner-io/enhancements/pull/197

Work items: 
* [x] Depends on https://github.com/submariner-io/subctl/pull/794
* [x] Depends on https://github.com/submariner-io/subctl/pull/805
* [x] Add tests to new build option

Jaanki:


subctl/issue-196-disbale-features-in-build.md
```diff
@@ -0,0 +1,88 @@
...
+
+    //go:build
+
+Constraints may appear in any kind of source file (not just Go), but they must appear near the top of the file, preceded only by blank
+lines and other line comments. These rules mean that in Go files a build constraint must appear before the package clause.
+
+To distinguish build constraints from package documentation, a build constraint should be followed by a blank line.
+
+A build constraint comment is evaluated as an expression containing build tags combined by ||, &&, and ! operators and parentheses.
+Operators have the same meaning as in Go.
```
skitt: This is copied from https://pkg.go.dev/cmd/go#hdr-Build_constraints; I don’t think it’s particularly useful to do so in this proposal. Instead, the proposal should explain how the feature will be used in Submariner:

```suggestion
Go provides [build constraints](https://pkg.go.dev/cmd/go#hdr-Build_constraints) which can be used to include or exclude files. In this proposal, these would be used to exclude files which are unnecessary for certain features.

We can exclude specific commands by adding `//go:build` constraints to the relevant `cmd` files, with a negated tag.
```

subctl/issue-196-disbale-features-in-build.md
```diff
@@ -0,0 +1,88 @@
...
+
+All the sub-commands that are related to deployment would be ignored from specific builds by adding build constraints.
+
+## Design
+
+Below `subctl` sub-commands would be having a build constraint so that they could be added in the package based on the user preferences.
```
skitt:
```suggestion
The following `subctl` sub-commands will have build constraint so that they can be excluded from the build based on tags provided at compile-time.
```

(This isn’t tied to user preferences, the users we’re considering here don’t build their own `subctl`.)

subctl/issue-196-disbale-features-in-build.md
```diff
@@ -0,0 +1,88 @@
...
+
+    deploy-broker
+    join
+    recover-broker-info
+    All cloud prepare
```
skitt:
```suggestion
    cloud
    deploy-broker
    join
    recover-broker-info
    uninstall
```

It’s not just `cloud prepare`, we want to exclude all `cloud` commands. `uninstall` should also be excluded.

subctl/issue-196-disbale-features-in-build.md
```diff
@@ -0,0 +1,88 @@
...
+
+The build directive would be `!troubleshoot`. An alternative directive could be `!non-deploy`
```
skitt: `troubleshoot` is a good idea, but let’s describe it from the tag rather than the build directive (starting with a negated term makes the description more difficult to get right). The executable we want to end up with when using tags contains only troubleshooting and service exporting features. So we need to find something which describes that when used in

    go build -tags foo

(I don’t have a good replacement right now.)

skitt (changes requested):

subctl/issue-196-disbale-features-in-build.md
Jaanki: Added a `Background` section to explain about build constraints and how to use it.

subctl/issue-196-disbale-features-in-build.md
```diff
@@ -0,0 +1,95 @@
...
+    
+        //go:build
+    
+    Constraints may appear in any kind of source file (not just Go), but they must appear near the top of the file, preceded only by blank
+    lines and other line comments. These rules mean that in Go files a build constraint must appear before the package clause.
+    
+    To distinguish build constraints from package documentation, a build constraint should be followed by a blank line.
+    
+    A build constraint comment is evaluated as an expression containing build tags combined by ||, &&, and ! operators and parentheses.
+    Operators have the same meaning as in Go.
```
skitt: This should be a block quote, not a code snippet:

```suggestion
> Go provides build constraints to identify files that should be included in the package.
> Build constraints are given by a line comment that begins
> 
>     //go:build
> 
> Constraints may appear in any kind of source file (not just Go), but they must appear near the top of the file, preceded only by blank
> lines and other line comments. These rules mean that in Go files a build constraint must appear before the package clause.
> 
> To distinguish build constraints from package documentation, a build constraint should be followed by a blank line.
> 
> A build constraint comment is evaluated as an expression containing build tags combined by ||, &&, and ! operators and parentheses.
> Operators have the same meaning as in Go.
```

subctl/issue-196-disbale-features-in-build.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Proposal
+
+All the sub-commands that are related to deployment would be excluded from builds by adding negated tags via build constraints. The
+downstream build of `subctl` would need to be built using the `tags` flag.
+
+## Design
+
+The following `subctl` sub-commands will have build constraint so that they can be excluded from the build based on tags provided at
```
skitt:
```suggestion
The following `subctl` sub-commands will have build constraints so that they can be excluded from the build based on tags provided at
```

skitt (changes requested):

skitt (approved):

subctl/issue-196-disbale-features-in-build.md
maayanf24: The sub-commands `export` and `unexport` are not part of troubleshooting capabilities.
How about using the tag `narrow`?

`go build -tags narrow`

subctl/issue-196-disbale-features-in-build.md
```diff
@@ -0,0 +1,95 @@
...
+>
+> To distinguish build constraints from package documentation, a build constraint should be followed by a blank line.
+>
+> A build constraint comment is evaluated as an expression containing build tags combined by ||, &&, and ! operators and parentheses.
+> Operators have the same meaning as in Go.
+
+## Proposal
+
+All the sub-commands that are related to deployment would be excluded from builds by adding negated tags via build constraints. The
+downstream build of `subctl` would need to be built using the `tags` flag.
```
maayanf24, line 31: Does that mean that the change in this proposal will only contain the build constraints of the sub-commands needed to be excluded?

subctl/issue-196-disbale-features-in-build.md
skitt: Right, that’s why I didn’t like `troubleshoot`, but `narrow` isn’t descriptive :-/. Janki’s `non-deploy` (which I think would have to be `non_deploy`) suggestion would be more accurate.

subctl/issue-196-disbale-features-in-build.md
skitt: Yes, that’s right. Ideally we’d also have a separate build test added to check that the constraints have the desired effect, but the existing build shouldn’t change at all.

maayanf24 (approved):

subctl/issue-196-disbale-features-in-build.md
mkolesnik: Since we're using `Makefile` for the build anyhow, wont it make more sense to tag all the deploy commands with `deploy` contraint tag, and specify it be default in the `Makefile`?
Users could then override the tags when they build with the Makefile should they want to.

This would also allow tagging other commands based on their utility, eg `service_discovery`, `diagnostics`, etc?
Then users could theoretically produce whatever CLI they want and turn on just the features they want (and the default would be all).

subctl/issue-196-disbale-features-in-build.md
```diff
@@ -0,0 +1,95 @@
...
+    deploy-broker
+    join
+    recover-broker-info
+    uninstall
+
+The build directive would be `!troubleshoot`. An alternative directive could be `!non-deploy`
+
+To ignore the above subcommands from the `subctl` build, it would need to be built with `-tags troubleshoot` flag.
+
+    go build -tags troubleshoot <other arguments>
```
mkolesnik: This specifically seems counterintuitive to me - this tag is not even in the troubleshooting commands but in the installation commands..

subctl/issue-196-disbale-features-in-build.md
Jaanki: The idea is to not change the upstream build but provide mechanism to ignore deploy related commands. The upstream build will be the vanilla build with all the commands and the build mechanism (go build command) to be run without any build tags.

This proposal just adds NEGATED `nondeploy` constraints to build command so that the downstream build (taken as an example to explain the use case) will be built using `--tags nondeploy` so that the deploy related commands are excluded.

So the upstream build mechanism is not changed. Build changes happen only for downstream builds where we dont want any commands. These `Makefile` changes will be made only downstream.

subctl/issue-196-disbale-features-in-build.md
Jaanki: Negated `nondeploy` tag will be used. I will update the PR to reflect that.

subctl/issue-196-disbale-features-in-build.md
skitt: One of the goals I had in mind was something we’d talked about a while ago, supporting `go install` — so that means ensuring that the default constraints produce a complete `subctl`. I know the current naming isn’t great...

subctl/issue-196-disbale-features-in-build.md
Jaanki: Updated.

tpantelis (approved):

subctl/issue-196-disbale-features-in-build.md
sridhargaddam: I see that the tag is still `troubleshoot`, did you miss updating the PR?

Jaanki: I updated the line above this. Missed this one. Rectified.

vthapar (approved):
Looks good. My concern was linking to upstream documentation for builds excluding subcommands, but looks like we already document limited subcommands in downstream docs.

aswinsuryan (approved):

sridhargaddam (approved):

dfarrell07 (approved):
Maybe we should mention that we should add a build and E2E test for this, but otherwise LGTM.

Issue #183, opened 2023-04-25
Add subctl upgrade command


Jaanki:
## Epic Description
A new `subctl upgrade` command should be added that upgrades the running enviroment to the version specified by the user. The command would do the following:

1. Upgrade subctl itself to the specified version
2. Upgrade Submariner to the specified version by re-running `deploy-broker` and `join`.

What it won't do is to change from a non-globalnet environment to globalnet enabled one or vice-versa. Feasibility of adding or removing the Service discovery component needs to be determined.

This is helpful in major and minor version upgrades.

## Acceptance Criteria
Submariner environment can be successfully upgraded to the specified version.

## Definition of Done (Checklist)

* [ ] Code complete
* [ ] Relevant metrics added
* [ ] The acceptance criteria met
* [ ] Unit/e2e test added & pass
* [ ] CI jobs pass
* [ ] Deployed using cloud-prepare+subctl
* [ ] Deployed using ACM/OCM addon
* [ ] Deploy using Helm
* [ ] Deployed on supported platforms (for e.g kind, OCP on AWS, OCP on GCP)
* [ ] Run subctl verify, diagnose and gather
* [ ] Uninstall
* [ ] Troubleshooting (gather/diagnose) added
* [x] Documentation added
* [x] Release notes added

## Work Items

- [ ] Write logic to upgrade subctl
- [ ] Add rerunning `deploy-broker` command with the specified Submariner version
- [ ] Add rerunning `join` command with the specified Submariner version

skitt:
Release notes and documentation: https://github.com/submariner-io/submariner-website/pull/1009

PR #200, opened 2023-05-26
feat(subctl): add `upgrade` sub-command to automate Submariner upgrade process

Jaanki:
This commit adds a new `upgrade` sub-command to `subctl` to automate the process of upgrading Submariner. The `upgrade` command takes a `to` flag to specify the version of Submariner to upgrade to. Once the command is run, `subctl` will upgrade itself to the specified version and then re-run `deploy-broker` and `join` sub-commands on all contexts where Broker and Submariner are installed respectively.

This feature addresses the issue #183.


subctl/issue-183-support-upgrading-submariner.md
```diff
@@ -0,0 +1,63 @@
...
+```mermaid
+flowchart TD
+    A[subctl upgrade --to 0.16.0] --> B{Check current subctl version?}
+    B --> |> 0.16.0|D[exit with error]
+    B --> |< 0.16.0|C[upgrade subctl to v0.16.0]
+    B --> |= 0.16.0|F[Prepare setup for upgrade. Nothing to be done currently]
+    C --> F[Prepare setup for upgrade. Nothing to be done currently]
+    subgraph Run `subctl upgrade` with upgraded subctl
+    F --> G[Run deploy-broker on Broker cluster with --image 0.16.0 flag]
+    G --> H[Run join on all joined clusters with --image 0.16.0 flag]
```
skitt, line 37: Do we need the `--image 0.16.0` flag? It should be possible to rely on the updated `subctl` to deploy its default version of Submariner.

However `subctl upgrade` will have to be able to handle deployments where the Submariner spec includes a version specifier; in such cases it should either return an error or be capable of replacing the version specifier.

subctl/issue-183-support-upgrading-submariner.md
```diff
@@ -0,0 +1,63 @@
...
+
+The proposal is to automate the process by adding a new `upgrade` sub-command that will do the above steps in the sequence.
+
+## Design details
+
+* subctl command name: `upgrade`
+* Flags for the command:
+  * `to`: mentions the version Submariner should be upgraded to. If not specified, upgrade to the latest stable version available.
+* Context to run on:
+  * Check on all contexts and re-run `deploy-broker` and `join` on clusters where Broker and Submariner are installed respectively.
```
sridhargaddam, line 21: Do we plan to support even kubeconfig as well? If so, I feel we should prompt the list of the contexts and allow the user to pick one (and maybe provide an option to skip this user-input when the command is executed from scripts)

subctl/issue-183-support-upgrading-submariner.md
```diff
@@ -0,0 +1,63 @@
...
+
+Once, the command is run, `subctl` will first be upgraded to the specified version. Once this is done, `subctl upgrade` will run again
```
sridhargaddam, line 23: So are you saying that if we are using say 0.15.x version of `subctl` and invoke upgrade, the `subctl` **CLI itself** will first be updated to `0.16.0` (assuming we have a release) or to the version specified by `--to` and then upgrade the deployment?

subctl/issue-183-support-upgrading-submariner.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+The upgrade process will only upgrade the `subctl` and `Submariner` version and not support changing any of the Broker or Submariner
+parameters nor changing repository from where these images are pulled.
+
+### Usage
+
+The new command can be used as follows:
+
+`subctl upgrade --to <version to upgarde to>`
```
sridhargaddam, line 51:
```suggestion
`subctl upgrade --to <version to upgrade to>`
```

subctl/issue-183-support-upgrading-submariner.md
Jaanki: yes, thats correct.

subctl/issue-183-support-upgrading-submariner.md
Jaanki: Right now, the idea is to have all clusters in the setup at the same version. There might be some functionality issues if all clusters are not at the same version, say for example: gather collects more info from 0.16 cluster than from 0.15 cluster. When we look at the gather data while troubleshooting, it might cause issues or the info might not be enough to troubleshoot properly.

subctl/issue-183-support-upgrading-submariner.md
Jaanki: `--image` flag re-iterates the version.

Can you please elaborate your sencond point.

skitt: If `Submariner.Spec.Version` is set to 0.15.0, and the user runs `subctl upgrade` with something newer, upgrading won’t actually do anything unless it changes `Submariner.Spec.Version`.

Jaanki: so `.Spec.Verion` will be updated on re-running `join` as part of the `upgrade` process.  `--image` flag would be helpful in this case.

skitt: We can’t use a flag for this. `subctl upgrade` has to preserve the intent the user indicated in the `Submariner` spec; using `--image` would result in a change to the spec which might not match what the user wants. If the spec specifies a version, we need to update that; if it doesn’t, we need to leave it alone.

subctl/issue-183-support-upgrading-submariner.md
sridhargaddam: Actually, I wanted to hear you thoughts on the following.

While debugging the issues, in most of the cases we have seen that the kubeconfig used by the users include multiple contexts all pointing to the same cluster. Ideally, we expect that there should be a single context for each cluster but at-least the reality is a bit different. So, if the kubeconfig includes multiple contexts for the same cluster, what steps are you planning to take to avoid re-running the upgrade commands on the cluster?

dfarrell07:
This feature has already been implemented. I guess we can make any easy tweaks to the EP, but otherwise we should likely not put in too much more work here and just merge it.

skitt (approved):

tpantelis (approved):

PR #230, opened 2024-08-27
Changes to make lighthouse MCS API complaint

vthapar:
Currently Lighthouse implementation of MCS API differs on 2 key points:

1. There is no VIP allocated for a MultiCluster Service.
2. EndpointSlices don't carry PodIPs for ClusterIP services.

This Enhancement proposes design changes to support above two in a user configurable way without impacting existing deployments.

### DIFF
--- a/lighthouse/mcs-compliance.md
+++ b/lighthouse/mcs-compliance.md
@@ -1,0 +1,186 @@
+# Add support for 100% MCS compliance
+
+Related issue:
+[Add support for 100% MCS Compliance](https://github.com/submariner-io/enhancements/issues/229)
+
+## Summary
+
+Current Lighthouse implementation differs from MCS API [^1] on two key points for ClusterIP Services:
+
+1. There is no VirtualIP [VIP] assigned in `ServiceImport`.
+2. `EndpointSlices` do not contain individual PodIPs.
+
+To better align with MCS API Lighthouse needs an option to allocate and track VirtualIPs in `ServiceImports` and use
+PodIPs in `EndpointSlices`.
+
+## Proposal
+
+When creating `ServiceExport`, users will have the option to add an annotation for enabling VIP for that Service.
+This annotation will also be added to the aggregated `ServiceImport`.
+
+A configuration flag will also be added to the `Submariner` and `ServiceDiscovery` CRDs, and `subctl` to set the
+options at the global level. This is for deployments where the user wants the option set for all exported Services and
+doesn't want to set the annotation explicitly for each `ServiceExport`.
+
+Default behavior with nothing configured would be same as it is currently i.e. no VIP and no PodIPs.
+
+Global flag can't be changed at runtime and requires a reinstall of whole clusterset. Change of flag in CRs will require
+a restart of Submariner on that cluster as it changes the behavior of `ServiceExports` without annotation.
+
+### VIP CIDR and allocation
+
+1. Default CIDR of `243.0.0/8` will be used for Clusterset. User can change this at the time of deploying broker.
+2. Each Cluster will get a default subset CIDR of `243.0.0.0/16` for allocating VIPs to Services exported from that
+cluster. User can change this at the time of join but this CIDR must be an unallocated subset of global VIP CIDR.
+3. VIP allocation is done by the first cluster to export the `Service`.
+4. When Service is exported on subsequent clusters, no VIP allocation is done if VIP already present on `ServiceImport`.
+5. VIP is deallocated only when `ServiceImport` is deleted i.e. `Service` or `ServiceExport` are no longer present on any
+of the clusters.
+6. If Submariner is uninstalled on the cluster that allocated a given VIP, VIP is not changed on `ServiceImport`.
+7. When `lighthouse-agent` starts, it should check `ServiceImport`s for any VIPs allocated from its VIP CIDR. Any such
+VIPs should be added to its local VIP allocation cache. This will also help with scenarios where cluster that allocated
+VIP has uninstalled Submariner and another cluster got the same VIP CIDR on installation.
+
+Note that `1` and `2` above are similar to Globalnet CIDR.
+
+### DNS Resolution
+
+Lighthouse DNS will return VIP for `ServiceImports` if available. It will be up to any other solution external to
+Submariner to load balance this VIP to PodIPs in `EndpointSlices`. If VIP is not present, it will behave as it does
+today, i.e. return one of the ClusterIPs from `EndpointSlices`.
+
+### Conflict Resolution
+
+In case of conflict in VIP setting when exporting `Service`, configuration on the oldest `ServiceExport`
+will be used. The `ServiceExport` in conflict will be exported with existing behavior and `ServiceExportConflict`
+condition will be set. The `Valid` condition will also be set to `true` as Service itself is exportable.
+
+### Creating `ServiceExport`
+
+Here are the detailed steps when creating a `ServiceExport`:
+
+1. When a new `ServiceExport` is created, check if enable VIP annotation is present. If not present, use global flag to
+determine if need to use VIP or not.
+2. If VIP is enabled based on condition above, check if `ServiceImport` exists or not.
+3. If `ServiceImport` doesn't exist, allocate the VIP and add it to new `ServiceImport`. Add an annotation
+`lighthouse.submariner.io/clusterset-ip-allocated-by=<cluster-id>` to identify the cluster that allocated the VIP.
+4. If a `ServiceImport` exists, compare the annotation on the `ServiceExport` and global flag against the annotation
+in `ServiceImport` for any conflicts.
+5. In case of conflict, set `ServiceExportConflict` condition on the `ServiceExport` and continue with export.
+6. Create `EndpointSlices` with PodIPs instead of ClusterIP. Follow existing rules for merging port information in
+`EndpointSlices`.
+
+### Updating annotation on ServiceExport
+
+Changing annotation will not modify the current behavior. It will only change the conflict
+condition on `ServiceExport` if it adds or resolves the conflict, depending on the change. If the users wish to change
+the behavior for a given `Service`, they will need to delete all `ServiceExports` and recreate them with required
+annotations.
+
+Same behavior will also apply on restart after changing global flag.
+
+### Deleting `ServiceExport`
+
+Here are the detailed steps when deleting a `ServiceExport`:
+
+1. Check the cluster list on `ServiceImport` to determine if this is the last cluster to Export the service.
+2. If this is last cluster, deallocate the VIP if it is same CIDR as current cluster and delete the `ServiceImport`.
+3. If not, proceed as usual irrespective of VIP annotation. This means deleting EndpointSlice, recalculating service
+ports and updating cluster list on `ServiceImport`.
+
+### Globalnet
+
+Globalnet will not be supported yet. Supporting Globalnet with PodIPs will be a significant scale issue as each Pod
+backing the Service will require a GlobalIngressIP.
+
+For any use cases that require Globalnet support with this feature, a separate Enhancement can be proposed in the
+future.
+
+### Migration
+
+Since there is no change in default behavior, nothing should be required for migration.
+
+If users want to change VIP behavior on existing ServiceExports post deployment, they will need to delete all
+`ServiceExports`.
+
+### Restart
+
+On restart, every lighthouse-agent should check aggregated `ServiceImport` for IPs allocated from its CIDR and allocate
+them in local allocation cache.
+
+## Design Details
+
+### Global Flags for deploy broker
+
+When deploying broker, `--enable-clusterset-ip` flag will be added. There will also be an `--clusterset-ip-cidr-range`
+option to set the VIP CIDR for clusterset to a non-default option. Irrespective of `--enable-clusterset-vip` setting,
+Clusterset CIDR Range will be set to default of `243.0.0.0/8` unless set to a different value. This is to support use
+cases where global flag is not set but user configures VIP on individual `ServiceExport`.
+
+Similar to Globalnet, this information will also be stored in `broker-info.subm` file and BrokerInfo ConfigMap.
+
+### Flags for subctl join
+
+Following flags will be added to `subctl join`:
+
+* `enable-clusterset-ip` - set the default behavior to enable VIP on this cluster. If not specified, it will pick the
+default from global configuration done during deploy broker.
+* `--clusterset-ip-cidr=a.b.c.d/x` - set the VIP CIDR for this cluster. Default will be allocated from Clusterset VIP
+CIDR Range confiured during deploy broker step.
+
+### New CRD flags
+
+Following CRDs will need to be modified to support these new flags
+
+```Go
+type BrokerSpec struct {
+    ClustersetIPEnabled bool `json:"clustersetIPEnabled,omitempty"`
+    ClustersetIPCIDRRange []string `json:"clustersetIPCIDRRange,omitempty"`
+}
+```
+
+```Go
+type SubmarinerSpec struct {
+    ClustersetIPEnabled bool `json:"clustersetIPEnabled,omitempty"`
+    ClustersetIPCIDR []string `json:"clusterserIpCidr,omitempty"`
+}
+```
+
+```Go
+type SubmarinerDiscoverySpec struct {
+    ClustersetIPEnabled bool `json:"clustersetIPEnabled,omitempty"`
+    ClustersetIPCIDR []string `json:"clusterserIpCidr,omitempty"`
+}
+```
+
+### Annotations for `ServiceExport`
+
+* `lighthouse.submariner.io/use-clusterset-ip` - Use VIP for the `ServiceExport`
+
+### Pros
+
+1. Doesn't modify existing behaviour
+2. Allows users to select behavior on a per Service basis.
+
+### Cons
+
+1. More configuration options for users to consider.
+2. More chances of `ServiceExports` on different clusters being in conflict due to wrong annotations etc.
+
+### Backward Compatibility
+
+None.
+
+### Alternatives
+
+* Only use Global flag for entire deployment. This was discarded in favor of annotations based approach to provide
+flexibility to users as they can mix and match services with and without VIP.
+
+## User Impact
+
+Existing users will not be impacted in any ways. Users who wish to use this feature will need to reinstall Submariner
+with new flags or use annotations on `ServiceExports`.
+
+## References
+
+[^1]: [KEP-1645: Multi-Cluster Services API](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#constraints-and-conflict-resolution)


lighthouse/mcs-compliance.md
```diff
@@ -0,0 +1,176 @@
...
+PodIPs but not VIP can already be taken care of by using HeadlessServices.
+
+Configuration flags will also be added to `Submariner` and `ServiceDiscovery` CRDs, and `subctl` to set these options at
+global level. This is for deployments where user only wants to use one specific option and doesn't want to set the
+annotation explicitly for each `ServiceExport`.
+
+Default behavior with nothing configured would be same as it is currently i.e. no VIP and no PodIPs.
+
+Global flag can't be changed at runtime and requires a reinstall of whole clusterset. Change of flag in CRs will require
+a reinstall of Submariner on that cluster as it changes he behavior of `ServiceExports` without annotation.
```
tpantelis: I don't think changing the behavior of a single exported service should require a complete reinstall. We can ignore an annotation change at the very least.

What happens if the annotation is different between exporting clusters, ie one specifies it but the another doesn't? I guess we can handle that as a conflict and apply the [resolution policy](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#global-properties)

```suggestion
a reinstall of Submariner on that cluster as it changes the behavior of `ServiceExports` without annotation.
```

lighthouse/mcs-compliance.md
```diff
@@ -0,0 +1,176 @@
...
+3. VIP allocation is done by the first cluster to export the `Service`.
+4. When Service is exported on subsequent clusters, no VIP allocation is done if VIP already present on `ServiceImport`.
+5. VIP is deallocated only when `ServiceImport` is deleted i.e. `Service` or `ServiceExport` are no longer present on any
+of the clusters.
+6. If Submariner is uninstalled on the cluster that allocated a given VIP, VIP is not changed on `ServiceImport`.
+7. When `lighthouse-agent` starts, it should check `ServiceImport`s for any VIPs allocated from its VIP CIDR. Any such
+VIPs should be added to its local VIP allocation cache. This will also help with scenarios where cluster that allocated
+VIP has uninstalled Submariner and another cluster got the same VIP CIDR on installation.
+
+Note that `1` and `2` above are similar to Globalnet CIDR.
```
tpantelis, line 44: and '7' I believe.

lighthouse/mcs-compliance.md
```diff
@@ -0,0 +1,176 @@
...
+### DNS Resolution
+
+Lighthouse DNS will return VIP for `ServiceImports` if available. It will be up to any other solution external to
+Submariner to load balance this VIP to PodIPs in `EndpointSlices`. If VIP is not present, it will behave as it does
+today, i.e. return one of the ClusterIPs from `EndpointSlices`.
+
+### Conflict Resolution
+
+In case of conflict in VIP and PodIP settings when exporting `Service`, configuration on the oldest `ServiceExport`
+will be used. The `ServiceExport` in conflict will not be exported and `ServiceExportConflict` condition will be set.
```
tpantelis: We can still export it as the conflict is resolvable.

lighthouse/mcs-compliance.md
```diff
@@ -0,0 +1,176 @@
...
+The `Valid` condition will also be set to `true` as Service itself is exportable.
+
+### Creating `ServiceExport`
+
+Here are the detailed steps when creating a `ServiceExport`:
+
+1. When a new `ServiceExport` is created, check if `ServiceImport` exists or not.
```
tpantelis: This refers to the aggregated SI on the broker. We won't allocate a VIP for the intermediate local Si's.

lighthouse/mcs-compliance.md
```diff
@@ -0,0 +1,176 @@
...
+2. If `ServiceImport` doesn't exist, allocate the VIP and add it to new `ServiceImport`.
+3. If `ServiceImport` exists, compare the combination of annotations on `ServiceExport` and global flags against
+annotations in `ServiceImport` for any conflicts.
+4. In case of conflict, set `ServiceExportConflict` condition on the `ServiceExport` and attempt conflict resolution.
+5. In case conflict can't be resolved, don't export the service.
```
tpantelis: It can always be resolved via oldest timestamp policy.

lighthouse/mcs-compliance.md
```diff
@@ -0,0 +1,176 @@
...
+Changing annotation will not modify the current behavior of using VIP or PodIPs. It will only change the conflict
+condition on `ServiceExport` if it adds or resolves the conflict, depending on the change. If the users wish to change
+the behavior for a given `Service`, they will need to delete all `ServiceExports` and recreate them with required
+annotations.
+
+### Deleting `ServiceExport`
+
+Here are the detailed steps when deleting a `ServiceExport`:
+
+1. Check the cluster list on `ServiceImport`. If this is the last cluster to Export the service. If not, proceed as
```
tpantelis:
```suggestion
1. Check the cluster list on `ServiceImport`. If this is not the last cluster to Export the service, proceed as
```

lighthouse/mcs-compliance.md
```diff
@@ -0,0 +1,176 @@
...
+usual irrespective of VIP and PodIP annotations. This means deleting EndpointSlice, recalculating service ports and
+updating cluster list on `ServiceImport`.
+2. If this is last cluster, deallocate the VIP if it is same CIDR as current cluster and delete the `ServiceImport`.
+
+### Globalnet
+
+Globalnet will not be supported yet. Supporting Globalnet with PodIPs will be a significant scale issue as each Pod
+backing the Service will require a GlobalIngressIP.
```
tpantelis, line 94: But we already do this for headless. I know it's possible there could be hundreds or thousands of pod IPs for a ClusterIP service but how prevalent is this use case? I mean I'd think most services would have no more than a handful of backing pods.

I'm wondering if this restriction will really hinder the use of the clusterset VIP functionality. Don't most deployments use globalnet?

lighthouse/mcs-compliance.md
```diff
@@ -0,0 +1,176 @@
...
+```Go
+type SubmarinerDiscoverySpec struct {
+    ClusterVIPCIDR []string `json:"clusterVipCidr,omitempty"`
+    EndpointSlicesWithPodIPs bool `json:"endpointSlicesWithPodIps"`
+}
+```
+
+### Annotations for `ServiceExport`
+
+* `lighthouse.submariner.io/use-vip` - Use VIP for the `ServiceExport`
```
tpantelis:
```suggestion
* `lighthouse.submariner.io/use-clusterset-vip` - Use VIP for the `ServiceExport`
```

lighthouse/mcs-compliance.md
```diff
@@ -0,0 +1,176 @@
...
+type BrokerSpec struct {
+    ClustersetVIPEnabled bool `json:"clustersetVIPEnabled,omitempty"`
+    ClustersetVIPCIDRRange []string `json:"clustersetVIPCIDRRange,omitempty"`
+    `EndpointSlices`WithPodIPs bool `json:"endpointSlicesWithPodIps,omitempty"`
+}
+```
+
+```Go
+type SubmarinerSpec struct {
+    ClusterVIPCIDR []string `json:"clusterVipCidr,omitempty"`
```
tpantelis:
```suggestion
    ClustersetVIPCIDR []string `json:"clusterVipCidr,omitempty"`
```

lighthouse/mcs-compliance.md
```diff
@@ -0,0 +1,176 @@
...
+    EndpointSlicesWithPodIPs bool `json:"endpointSlicesWithPodIps"`
+}
+```
+
+```Go
+type SubmarinerDiscoverySpec struct {
+    ClusterVIPCIDR []string `json:"clusterVipCidr,omitempty"`
```
tpantelis:
```suggestion
    ClustersetVIPCIDR []string `json:"clusterVipCidr,omitempty"`
```

lighthouse/mcs-compliance.md
```diff
@@ -0,0 +1,176 @@
...
+### Global Flags for deploy broker
+
+When deploying broker, `--enable-vip` and `--endpointslices-with-pod-ips` flags will be added. There will also be an
+`--clusterset-vip-cidr` option to set the VIP CIDR for clusterset to a non-default option.
+
+### Flags for subctl join
+
+Following flags will be added to `subctl join`:
+
+* `--cluster-vip-cidr=a.b.c.d/x` - set the VIP CIDR for this cluster. Default will be allocated from Clusterset VIP CIDR
```
tpantelis:
```suggestion
* `--clusterset-vip-cidr=a.b.c.d/x` - set the VIP CIDR for this cluster. Default will be allocated from Clusterset VIP CIDR
```

lighthouse/mcs-compliance.md
```diff
@@ -0,0 +1,176 @@
...
+
+Since there is no change in default behavior, nothing should be required for migration.
+
+If users want to add annotations on existing ServiceExports post deployment, they will need to delete ServiceExports.
+
+## Design Details
+
+### Global Flags for deploy broker
+
+When deploying broker, `--enable-vip` and `--endpointslices-with-pod-ips` flags will be added. There will also be an
```
tpantelis:
```suggestion
When deploying broker, `--enable-clusterset-vip` and `--endpointslices-with-pod-ips` flags will be added. There will also be an
```

lighthouse/mcs-compliance.md
```diff
@@ -0,0 +1,176 @@
...
+To better align with MCS API Lighthouse needs an option to allocate and track VirtualIPs in `ServiceImports` and use
+PodIPs in `EndpointSlices`.
+
+## Proposal
+
+When creating `ServiceExport`, users will have option to add annotation for enabling VIP for that Service and another
+to use PodIPs in `EndpointSlices`. These annotations will also be added to the aggregated `ServiceImport`.
+
+Option to add PodIPs to EndpointSlice will only be applicable when VIP option is also enabled. A use case that requires
+PodIPs but not VIP can already be taken care of by using HeadlessServices.
```
tpantelis: Then why do we need the EPS option? With VIP enabled, it seems we should just always publish the pod IPs. Wouldn't an external solution need the pod IPs? It seems L51-52 below imply that. I think we should simplify and only have one flag.

tpantelis (changes requested):

lighthouse/mcs-compliance.md
vthapar: I was thinking of a scaleable solution where someone wants to use Submariner with VirtualIP and use it to LoadBalance across clusters. e.g. when using something like reference MCS controller.

lighthouse/mcs-compliance.md
vthapar: 7 is not exactly same as Globalnet coz globalnet allocations are always of scope cluster.

lighthouse/mcs-compliance.md
vthapar: We can, but wanted to birng up for discussion if we want to. e.g. Someone setting the annotation on Export may expect VIP. But then, conflict should clue them in to why it didn't work that way.

lighthouse/mcs-compliance.md
vthapar: Yes, it implies Aggregated SI. How we handle local SI doesn't matter much.

lighthouse/mcs-compliance.md
vthapar: Left it to review, if we should resolve or not.

lighthouse/mcs-compliance.md
vthapar: Globalnet+Headless is already a very small use case. This is not an issue for Globalnet+ClusterIP normally as each service gets a single globalIP by creation of an internal ClusterIP type service.

Normally headless or stateful services don't have as many pods as a ClusterIP service. ClusterIP services are mostly used for dynamic scale up/down use cases so no. of pods can get really high. Headless services tend to have much lesser number of pods.

lighthouse/mcs-compliance.md
vthapar: I didn't use `clusterset` coz at broker level it is implied. Wanted to keep the keyword short.

lighthouse/mcs-compliance.md
vthapar: This is for Cluster, borrowing terminology from ACM. But I see the point in using term ClustersetCIDR vs ClusterCIDR which is used for local cluster.

Maybe `clusterset-cidr` works better? Gets point across and shorter keyword too.

lighthouse/mcs-compliance.md
vthapar: Same as above, `ClustersetCIDR` ?

lighthouse/mcs-compliance.md
tpantelis: We could also resolve the conflict by merging, ie any `ServiceExport` with the annotation results in a VIP. The first exporting cluster with the annotation allocates the VIP.

lighthouse/mcs-compliance.md
tpantelis: OK but what about the second part of my comment.

> I'm wondering if this restriction will really hinder the use of the clusterset VIP functionality. Don't most deployments use globalnet?

lighthouse/mcs-compliance.md
tpantelis: I think we should be consistent with our terminology everywhere.

lighthouse/mcs-compliance.md
tpantelis: 👍

lighthouse/mcs-compliance.md
tpantelis: 👍

lighthouse/mcs-compliance.md
tpantelis: I think we should initially focus on the known use case which is integration with `Istio`, ie VIP + pod IPs. I imagine the use case you mention could come from an upstream user (ie not ACM) and wouldn't be that prevalent, in which case we could add future support via the pod IP annotation only. I'd prefer to minimize the global settings only to the prevalent use case.

lighthouse/mcs-compliance.md
tpantelis: > The first exporting cluster with the annotation allocates the VIP.

I should clarify - that's if the annotation is set to `true`. We could still have a conflict if another cluster explicitly sets their annotation to `false`. In that case, report the conflict but take the `true` setting.

Another interesting scenario - what if the global VIP option is set but a `ServiceExport's` annotation is set to `false`.... this would be a way to disable VIP for a particular service.

tpantelis: Thinking more on this, I think we should keep this as simple as possible, ie VIP enablement for a service is determined by the exporting cluster that creates the aggregated `ServiceImport`.  So:

  - if global flag is set AND either `ServiceExport` annotation is not present OR the annotation is set to true THEN VIP is enabled

  - if global flag is not set OR the annotation is present AND set to true THEN VIP is enabled

If the `ServiceExport` annotation for a subsequent exporting cluster doesn't match the annotation on the existing aggregated `ServiceImport` then a conflict is reported.

tpantelis: Just to be clear, regardless of whether not the user specifies the global VIP flag, we always need to allocate a VIP CIDR for each cluster from the clusterset CIDR to support the ServiceExport annotation.

vthapar: > > The first exporting cluster with the annotation allocates the VIP.
> 
> I should clarify - that's if the annotation is set to `true`. We could still have a conflict if another cluster explicitly sets their annotation to `false`. In that case, report the conflict but take the `true` setting.
> 
Yes. I didn't word it properly. Maybe hsould use the phrase "if VIP is enabled" and VIP enabled definition is the combination of global+annotation as you clarified.
> Another interesting scenario - what if the global VIP option is set but a `ServiceExport's` annotation is set to `false`.... this would be a way to disable VIP for a particular service.
Yes. Global flag is only used when annotaiton is not available on export. If available, it overrides whatever is set in Global.

lighthouse/mcs-compliance.md
vthapar: Missed that. Istio doesn't support overlapping CIDRs today, so anyone looking to use Istio+Submariner wouldn't need Globalnet.

tpantelis: Ahh - lucky for us 😄

lighthouse/mcs-compliance.md
```diff
@@ -0,0 +1,187 @@
...
+1. There is no VirtualIP [VIP] assigned in `ServiceImport`.
+2. `EndpointSlices` do not contain individual PodIPs.
+
+To better align with MCS API Lighthouse needs an option to allocate and track VirtualIPs in `ServiceImports` and use
+PodIPs in `EndpointSlices`.
+
+## Proposal
+
+When creating `ServiceExport`, users will have option to add annotation for enabling VIP for that Service and another
+to use PodIPs in `EndpointSlices`. These annotations will also be added to the aggregated `ServiceImport`.
```
tpantelis:
```suggestion
When creating `ServiceExport`, users will have the option to add an annotation for enabling VIP for that Service. This annotation will also be added to the aggregated `ServiceImport`.
```

lighthouse/mcs-compliance.md
```diff
@@ -0,0 +1,187 @@
...
+
+Configuration flags will also be added to `Submariner` and `ServiceDiscovery` CRDs, and `subctl` to set these options at
+global level. This is for deployments where user only wants to use one specific option and doesn't want to set the
+annotation explicitly for each `ServiceExport`.
```
tpantelis:
```suggestion
A configuration flag will also be added to the `Submariner` and `ServiceDiscovery` CRDs, and `subctl` to set the options at
 the global level. This is for deployments where the user wants the option set for all exported Services and doesn't want to set the
annotation explicitly for each `ServiceExport`.
```

lighthouse/mcs-compliance.md
```diff
@@ -0,0 +1,187 @@
...
+
+1. When a new `ServiceExport` is created, check if enable VIP annotation is present. If not present, use global flag to
+determine if need to use VIP or not.
+2. If VIP is enabled based on condition above, check if `ServiceImport` exists or not.
+3. If `ServiceImport` doesn't exist, allocate the VIP and add it to new `ServiceImport`. Add an annotation
+`lighthouse.submariner.io/clusterset-ip-allocated-by=<cluster-id>` to identify the cluster that allocated the VIP.
+4. If `ServiceImport` exists, compare the combination of annotations on `ServiceExport` and global flags against
+annotations in `ServiceImport` for any conflicts.
+5. In case of conflict, set `ServiceExportConflict` condition on the `ServiceExport` and attempt conflict resolution.
+6. In case conflict can't be resolved, don't export the service.
```
tpantelis:
```suggestion
4. If a `ServiceImport` exists, compare the annotation on the `ServiceExport` and global flag against
the annotation in `ServiceImport` for any conflicts.
5. In case of conflict, set `ServiceExportConflict` condition on the `ServiceExport` and continue with export.
```

tpantelis (changes requested):

lighthouse/mcs-compliance.md
vthapar: Done

lighthouse/mcs-compliance.md
vthapar: Done

lighthouse/mcs-compliance.md
vthapar: Done

tpantelis (approved):

lighthouse/mcs-compliance.md
```diff
@@ -0,0 +1,186 @@
...
+### VIP CIDR and allocation
+
+1. Default CIDR of `243.0.0/8` will be used for Clusterset. User can change this at the time of deploying broker.
+2. Each Cluster will get a default subset CIDR of `243.0.0.0/16` for allocating VIPs to Services exported from that
+cluster. User can change this at the time of join but this CIDR must be an unallocated subset of global VIP CIDR.
+3. VIP allocation is done by the first cluster to export the `Service`.
+4. When Service is exported on subsequent clusters, no VIP allocation is done if VIP already present on `ServiceImport`.
+5. VIP is deallocated only when `ServiceImport` is deleted i.e. `Service` or `ServiceExport` are no longer present on any
+of the clusters.
+6. If Submariner is uninstalled on the cluster that allocated a given VIP, VIP is not changed on `ServiceImport`.
```
yboaron, line 39: Some corner case Qs:

Regarding 
`If Submariner is uninstalled on the cluster that allocated a given VIP, VIP is not changed on ServiceImport`
Don't we delete all serviceimport CRs when submariner is uninstalled ?


Regarding 
`VIP allocation is done by the first cluster to export the Service.` 
do we need to implement  some lock mechanism to resolve race condition in case two clusters export service at the same time ?

tpantelis: No - that’s handled by using K8s retry on conflict functionality.

vthapar: > Some corner case Qs:
> 
> Regarding `If Submariner is uninstalled on the cluster that allocated a given VIP, VIP is not changed on ServiceImport` Don't we delete all serviceimport CRs when submariner is uninstalled ?
Aggregated ServiceImport will still exist if service is exported on other clusters. Deletion will only happen if this was the last cluster to export service, in which case it will be deleted. Point here is that if the cluster allocating VIP is no longer there [deleted or submariner removed] we don't try to change the VIP as it will impact existing clients.
> 
> Do we expect that cluster will get the same VIPs cidr subset after reinstall ?
No. We don't make any assumptions and it is also possible that some other cluster gets that CIDR. We can see this happen with Globalnet.
> 
> Regarding `VIP allocation is done by the first cluster to export the Service.` do we need to implement some lock mechanism to resolve race condition in case two clusters export service at the same time ?
Not needed. As both try to create Aggregated ServiceImport, only one will be able to do so. Other will get conflict with `Resource already exists` error.

yboaron (approved):

PR #240, opened 2024-12-03
Proposal for IPV6 datapath support

yboaron:


submariner/IPV6-datapath.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Gateway
+
+To support IPV6 the gateway should:
+
+* discover publicIP, privateIP, healthcheckIP and cluster’s subnets for each IP family.  
+  * Note: gateway should address corner cases related to this change, for example in dual-stack environment only the V4 publicIP address is successfully resolved.    
+* run NAT Discovery per IP family in remote Endpoint.
+* advertise in local Endpoint IP details based on cluster networking type, for example in duals-stack cluster both V4 Public IP and V6 Public IP should be advertised in Endpoint.   
+* continue advertising a **single** Endpoint. in case of a dual-stack cluster, fields should consist of both V4 and V6 addresses separated.
```
tpantelis, line 121: I think we should change the appropriate IP fields to arrays and bump the API version and implement a webhook converter so each version can be served for backwards compatibility.

tpantelis (commented):
The image links don't work from the rich diff view.

submariner/IPV6-datapath.md
```diff
@@ -0,0 +1,129 @@
...
+
+* Dual-stack Pod networking (a single IPv4 and IPv6 address assignment per Pod)  
+* IPv4 and IPv6 enabled Services  
+* Pod off-cluster egress routing (eg. the Internet) via both IPv4 and IPv6 interfaces
+
+Currently, Submariner only supports IPV4 datapath and this proposal explains the changes required to support IPV6 and dual-stack clusters.
+
+## Inter-cluster datapath requirements
+
+The next table describes the required inter-cluster connectivity for the various combinations of clusters networking configurations:
```
sridhargaddam: What if we have more than 2 clusters - f.e., lets say we have three clusters A, B, C
A is dualStack
B is IPv4 only
C is IPv6 only

Even if A connects to B via IPv4 and to C via IPv6, the connectivity between B and C will not work. Taking this as an example, can you please include notes on how Lighthouse would support service discovery in a hybrid environment - as advertising v6 addresses to a v4 cluster (vice versa) will not help/work.

sridhargaddam: I would like you to consider another use-case. Lets say we have just two clusters D and E.
D is a dualStack cluster
E is an IPv4 only cluster.

In this scenario, if we establish IPv4 tunnels between the cluster, and if the user exports an IPv6 only service (by setting .spec.ipFamilies to "IPv6"), how will Submariner handle this scenario? Does it ignore such exported services from advertising to the Broker or is it expected that cluster E would get the serviceImport details from Broker and will ignore it as it has only IPv6 address and not v4.

submariner/IPV6-datapath.md
```diff
@@ -0,0 +1,129 @@
...
+
+The next diagram illustrates Submariner’s datapath architecture for kube-proxy based CNIs:
+![non-ovnk-architecture](./images/dual-stack-arch-for-non-ovnk.png)
+
+With the proposed architecture, Submariner needs to establish both V4 and V6 intra-cluster egress routing to the GW node in the case of dual-stack.
+
+Pod IPV4 egress packets for CNI != OVNK and cable-driver=libreswan will be:
+![non-ovnk-ipv4-egress](./images/ipv4-non-ovnk-egress-packets.png)
+
+And Pod IPV6 egress packets for the same configuration will be:
```
sridhargaddam, line 101: For IPv4 vxlan tunnels, we use 242.x.x.x cidr rangle, what is the IPv6 CIDR range you are planning to use for setting up tunnels?

submariner/IPV6-datapath.md
```diff
@@ -0,0 +1,129 @@
...
+![non-ovnk-ipv6-egress](./images/ipv6-non-ovnk-egress-packets.png)
+
+**Note**: In future, we may optimize this architecture for a dual-stack case.
```
sridhargaddam, line 106: If the cluster is IPv4 only or dualStack, probably starting with a single V4 VxLAN for both might be a better option from the initial implementation. Otherwise you have to consider the impact on upgrades for subsequent changes in future. Just a suggestion.

submariner/IPV6-datapath.md
sridhargaddam: If we are changing any CRDs, we should also consider the impact during upgrades as there will be brownfield deployments.

submariner/IPV6-datapath.md
```diff
@@ -0,0 +1,129 @@
...
+## Datapath breakdown
+
+### Gateway
+
+To support IPV6 the gateway should:
+
+* discover publicIP, privateIP, healthcheckIP and cluster’s subnets for each IP family.  
+  * Note: gateway should address corner cases related to this change, for example in dual-stack environment only the V4 publicIP address is successfully resolved.    
+* run NAT Discovery per IP family in remote Endpoint.
+* advertise in local Endpoint IP details based on cluster networking type, for example in duals-stack cluster both V4 Public IP and V6 Public IP should be advertised in Endpoint.
```
sridhargaddam: We will not be having any explicity field to specify the type of cluster (i.e., IPv4 only, IPv6 only or dualStack), but the presence of multiple addresses in the endpoint will be like a signal that remote cluster has more than one IP for an element and depending on the type of IPaddress it could be v4/v6/both. I hope this is correct?

submariner/IPV6-datapath.md
tpantelis: yes - a webhook converter would handle that, ie convert to and from both versions depending on client requests.

submariner/IPV6-datapath.md
yboaron: will update the proposal

submariner/IPV6-datapath.md
yboaron: I haven't checked the subject in depth, but I assume we'll find a similar address space for V6 as well.
Does it make sense?

submariner/IPV6-datapath.md
yboaron: Yep, that's correct.

submariner/IPV6-datapath.md
yboaron: In that case, I guess we need to be persistent and change also other resources , for example 
GlobalnetCIDRRange and ClustersetIPCIDRRange in Broker resource. right ?

tpantelis: yeah whatever is needed

vthapar: @tpantelis Why not follow the core API example add a new field with `xxxIPs` which is an array? This makes it easy to handle upgrades and allows us to depricate old `xxxIP` field over time.

vthapar (dismissed):

submariner/IPV6-datapath.md
tpantelis: We could do that but then we’d have to handle both fields in various places in code going forward. A webhook converter handles all that transparently.  Webhooks and CRD versioning were designed for cases like this.  K8s core API adds new fields and deprecates old ones b/c they’re  not CRDs and much harder to bump versions.  We can discuss further (next year 🙂)

submariner/IPV6-datapath.md
```diff
@@ -0,0 +1,185 @@
...
+Submariner’s control plane uses a central broker component to facilitate the exchange of metadata information between participating clusters.
+
+Both inter-cluster datapath and service-discovery (Lighthouse) apply their configuration based on the information received from control plane.
+
+The Broker, inter-cluster datapath and Lighthouse should be updated to support mixed clusters environment (e.g: dual-stack, V6 only , V4 only).
+
+### Control plane
+
+For mixed clusters environment we may experience connectivity issue.
+Some clusters and broker may not connect if the Broker’s cluster doesn’t support dual-stack.
```
sridhargaddam: This statement implies that Broker cluster **has** to support dualStack in mixed env. I think this is not required - f.e., if all the managed clusters support dualStack then having a Broker cluster that just supports IPv4 alone should be sufficient isnt it?
The Gateways on the managed clusters can connect to Broker over IPv4 and still exchange metadata that includes info of V4 and V6.

submariner/IPV6-datapath.md
```diff
@@ -0,0 +1,185 @@
...
+| 2 | V4 | V6 | N/A |
+| 3 | V6 | dual-stack | V6 |
+| 4 | V6 | V6 | V6 |
+| 5 | V4 | dual-stack | V4 |
+| 6 | dual-stack | dual-stack | V4,V6 |
+
+This can be achieved in one of the following ways:
+
+* Lighthouse ignores processing imported services IP addresses that don’t match local cluster networking configuration.
+* Lighthouse will process imported services regardless of local cluster networking configuration, and rely on local workloads dns requests.
```
sridhargaddam: Looks like these are contradicting statements. Can you please rephrase.

submariner/IPV6-datapath.md
```diff
@@ -0,0 +1,185 @@
...
+For example, in use case \#3 in the table above, Lighthouse DNS database will store both V4 and V6 addresses.
+However the local dns client will look for the V6 (AAAA) record.
```
sridhargaddam: Normally, all the clients/apps make DNS requests for both v4/v6 and try connecting to them in parallel with a slight preference for v6. Please see Happy Eyeballs algo - https://datatracker.ietf.org/doc/html/rfc6555

submariner/IPV6-datapath.md
yboaron: Correct, will rephrase

submariner/IPV6-datapath.md
yboaron: Is this the case also for single-stack clients? 
I mean does V6 client/app try to connect to V4 address returned from DNS ?

submariner/IPV6-datapath.md
yboaron: I've rephrased it, let me know if it looks OK to you.

submariner/IPV6-datapath.md
sridhargaddam: Yes, if the DNS server returns any v4 address, then the client will try to connect to it (which would fail if the server is not listening on it).

yboaron: Interesting, if V6 client tries to connect to IP V4 address I assume we should go with first approach

sridhargaddam: If the local cluster supports only V4 and the remote cluster is v6 only (or dualStack), then returning V6 address as part of Lighthouse may not help since we do not use any intermediate proxies (or NAT mechanisms) to translate v4 to v6. So, approach 1 seems reasonable.

yboaron: The plan is to add [validation](https://github.com/submariner-io/enhancements/pull/240/files#diff-622b5a04b96a48ea8501651697d89c87465357f36b392c61e23d8f4d024ef9feR46) for clusters connectivity, 
so a cluster should join clusterset only if it can communicate with all existing clusters in the set.

Even after adding the validation we should still handle the following special cases:

1. client : single stack V4 , DNS : V4, V6 entries
2. client : single stack V6 , DNS : V4, V6 entries
3. client : dual-stack ,         DNS : V4 entry 
4. client : dual-stack ,         DNS : V6 entry 

Now, IIUC your comments, if we go with approach '2' things may fail for ^^ use cases.

just double checking, for use case#1, the V4 client may try to connect to the V6 address, right ?  (not sure how the IP header looks like in this case)

sridhargaddam: >just double checking, for use case#1, the V4 client may try to connect to the V6 address, right ?

An IPv4 client can make a DNS query for both A (IPv4) and AAAA (IPv6) records. If there is a response to the AAAA query, the client can then check whether the network-interface/host supports IPv6 and decide accordingly whether to connect over IPv6. OTOH, some client implementations may first check the network-interface/host to see if IPv6 is supported and only then send a AAAA query. Therefore, the behavior depends on the implementation of the client.

skitt: It’s quite common to see simultaneous A/AAAA queries, and then connection attempts on whatever comes back first, with a fallback to the other, regardless of what the host claims to support (because there are systems with working local IPv6 but no external IPv6 connectivity, and likewise for IPv4).

sridhargaddam: Most of the modern client applications that support dual-stack environments implement the Happy eyeballs algorithm that tries to connect to both V4 and V6 addresses (returned from DNS) concurrently (or with a small delay in v4) and use the connection that succeeds first to improve the user-experience. In the Submariner context, since it is aware of both local and remote cluster networking and because it has control over LH DNS Server, approach 1 seems more optimal.

submariner/IPV6-datapath.md
```diff
@@ -0,0 +1,190 @@
...
+| dual-stack | dual-stack | V4,V6 |
+
+Verification of cluster connectivity should be added when cluster joins a clusterset.
+A cluster will only be added to clusterset if it can communicate with all existing clusters in the set.
+
+For example:
+
+V4 cluster A should successfully join the clusterset.  
+Dual-stack cluster B should successfully join the clusterset.  
+V6 cluster C should fail joining the clusterset (because it can't connect to cluster A).
```
sridhargaddam, line 53: In the table above, its mentioned that if we have a v6 cluster and a dual-stack cluster, v6 would be the supported connectivity. Whereas this line mentions that cluster C will be unable to join the clusterSet. Is this validation done at the time of joining the cluster to the Broker?

yboaron: Yep, the validation should be done when the cluster joins the clusterset.
Cluster C (V6) won't be able to join the clusterset , because it can't connect cluster A (V4)

submariner/IPV6-datapath.md
yboaron: Trying to summarize:

So, for use cases similar to # 1, the client will not try to connect to the server's V6 address, but it may waste time checking if the local host/network interface supports IPv6.


Approach # 1:
This may not be optimal (I assume the extra time to check the local host/network interface capabilities should be very short) but things should work in a reasonable time and make for a simpler/uniform Lighthouse implementation

Approach # 2:
Faster connection to remote server, more complex Lighthouse implementation

cc @vthapar

sridhargaddam: I think there is some misunderstanding. I'm personally suggesting that we go with approach 1 and **not** approach 2. We cannot control the behavior of the client whether its checking if the local host/network-iface supports v6 or not. However, we do control the LH DNS Server and can respond to only supported ipFamilies (i.e., return AAAA/v6 of the remote service, only if the local cluster supports v6/dualStack). Approach 1 would also make the implementation of LH DNS server simpler.

submariner/IPV6-datapath.md
tpantelis: I've done research and testing wrt CRD version bump and webhook converter. In order for migration and rolling upgrade to be seamless, we would need a webhook converter. While doable, it does present challenges wrt TLS certificate and authentication between the webhook and the API server and other configuration. I don't think there's a compelling enough reason to tackle all of that. We can simply add plural IP fields where necessary and support the singular fields for migration with minimal effort.

submariner/IPV6-datapath.md
yboaron: OK, updated the proposal.

submariner/IPV6-datapath.md
yboaron: We decided to follow K8S core API example and add new plural fields (e.g: xxxIPs)

yboaron: Updated the proposal

submariner/IPV6-datapath.md
```diff
@@ -0,0 +1,190 @@
...
+  For example, in a dual-stack environment, only the V4 public IP address might be successfully resolved.
+* run NAT Discovery per IP family in remote Endpoint.
+* Advertise IP details in the local Endpoint based on the cluster's networking type.
+For example, in a dual-stack cluster, both V4 Public IP and V6 Public IP should be advertised in the Endpoint.
+* continue advertising a **single** Endpoint. in case of a dual-stack cluster, fields should consist of both V4 and V6 addresses separated.
+
+**Note**: Three new fields (HealthCheckIPs, PrivateIPs, PublicIPs) should be added to the Endpoint resource to support dual-stack.
+ These fields will store multiple IP addresses for dual-stack scenarios.
+
+* create inter-cluster tunnel only if local endpoint networking type matches remote endpoint ipfamily.
```
sridhargaddam, line 126: Lets say we have two dual-stack clusters A and B. If the clusters are able to successfully connect over "IPv4" but fail to connect over "IPv6", what is the impact on the connection status (f.e., the user facing commands like subctl show xxx). Please add some notes about this scenario.

yboaron: Done.

sridhargaddam:
LGTM, will let others to take a look before I approve it.

tpantelis (approved):

aswinsuryan (approved):

sridhargaddam (approved):

PR #256, opened 2025-09-08
Add support for OVN-Ipsec deployments

aswinsuryan:
Add support for OVN-Ipsec deployments

### DIFF
--- a/submariner/ipsec-certificate-support.md
+++ b/submariner/ipsec-certificate-support.md
@@ -1,0 +1,85 @@
+# Submariner Enhancement Proposal: Certificate Mode for OVN-Kubernetes IPsec
+
+## Summary
+
+This enhancement enables Submariner to **share the Pluto (Libreswan) process** with OpenShift/OVN-Kubernetes when OVN
+IPsec is enabled. The primary goal is to allow both Submariner and OVN-Kubernetes to provide secure, encrypted inter-cluster
+and intra-cluster networking on the same OpenShift cluster, without process conflicts or network disruption.
+
+To achieve this, Submariner reuses the existing Pluto process started by OVN-Kubernetes, rather than starting its own.
+Because OVN-Kubernetes configures Pluto to use X.509 certificates for authentication, Submariner must also operate in
+**certificate mode** as a necessary side effect.
+
+To achieve this, Submariner is enhanced to:
+
+- Reuse the existing Pluto (Libreswan) process started by OVN-Kubernetes, rather than starting its own.
+- Operate in certificate mode, matching OVN’s use of X.509 certificates for authentication.
+- Provide an explicit deployment option to enable this mode, ensuring that Submariner and OVN-Kubernetes can both manage
+IPsec tunnels independently but harmoniously.
+
+---
+
+## Motivation
+
+- **Problem:**
+  When both Submariner and OVN-Kubernetes IPsec are enabled, each tries to start its own Pluto process, causing conflicts
+and breaking cluster networking.
+- **Goal:**
+  Allow Submariner and OVN-Kubernetes to run together, each managing their own tunnels, but sharing
+the same Pluto process and certificate-based authentication infrastructure.
+
+---
+
+## Design Details
+
+### Manual Enablement
+
+- Certificate mode must be **explicitly enabled** by passing an option (e.g., `--ipsec-cert-mode`) to `subctl` during
+deployment or join.
+
+### Pluto Process Expectation
+
+- When certificate mode is enabled, Submariner **expects** that a compatible Pluto process (as started by OVN-Kubernetes)
+is already running
+- Submariner does **not** attempt to start or manage the Pluto process in this mode. It only writes its connection
+configuration (`/etc/ipsec.d/submariner.conf`) and loads connections into the existing Pluto instance.
+- If the Pluto process is not running when Submariner is deployed in certificate mode, tunnel setup will fail and errors
+will be logged.
+
+### Certificate Management
+
+- Submariner automates certificate issuance, signing, and import into the NSS database (used by Libreswan for authentication),
+using Kubernetes Secrets and controllers.
+- The NSS database used by Submariner for certificate storage is located at `/var/lib/ipsec/nss` on the Gateway node.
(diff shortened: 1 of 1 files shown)


tpantelis (approved):

submariner/ipsec-certificate-support.md
```diff
@@ -0,0 +1,86 @@
...
+  Allow Submariner and OVN-Kubernetes to run together, each managing their own tunnels, but sharing
+the same Pluto process and certificate-based authentication infrastructure.
+
+---
+
+## Design Details
+
+### Manual Enablement
+
+- Certificate mode must be **explicitly enabled** by passing an option (e.g., `--ipsec-cert-mode`) to `subctl` during
```
yboaron, line 37: Just double-checking — a "mixed" IPSec cert mode in a clusterset should be supported, right? By "mixed" I mean one cluster joins with the --ipsec-cert-mode flag, while another joins without specifying this parameter.

submariner/ipsec-certificate-support.md
```diff
@@ -0,0 +1,86 @@
...
+will be logged.
+
+### Certificate Management
+
+- Submariner automates certificate issuance, signing, and import into the NSS database (used by Libreswan for authentication),
+using Kubernetes Secrets and controllers.
+- The NSS database used by Submariner for certificate storage is located at `/var/lib/ipsec/nss` on the Gateway node.
+- The CA is managed in the broker cluster, and certificate Secrets are synchronized between broker and local clusters.
+- **Certificate Signing Flow:**
+  - The Submariner Gateway (acting as a CSR signer) generates a Certificate Signing Request (CSR) and sends it to the broker cluster.
```
yboaron: What’s the expected flow during a gateway switchover? Should the newly active gateway repeat this process?

submariner/ipsec-certificate-support.md
```diff
@@ -0,0 +1,86 @@
...
+
+### RBAC and Cleanup
+
+- Proper RBAC is set up for certificate management, and uninstall logic ensures clean removal of Submariner’s configuration and certificates.
+
+### Status and Observability
+
+- Connection status is reported in the Gateway object, the existing logic does not require any changes for this.
+- Debug logs are added for certificate import and NSS operations.
+
```
yboaron, line 69: Do we plan to support setting/removing `--ipsec-cert-mode` in runtime ?

submariner/ipsec-certificate-support.md
```diff
@@ -0,0 +1,86 @@
...
+  ```sh
+  subctl join --ipsec-cert-mode broker-info.subm
+  ```
+
+- This will configure Submariner to:
+  - Use certificate-based authentication.
+  - Reuse the existing Pluto process.
+  - Avoid any actions that would interfere with OVN-Kubernetes IPsec.
+- It is the administrator’s responsibility to ensure that the Pluto process is running and managed by OVN-Kubernetes before deploying Submariner
+in this mode.
```
yboaron, line 85: Will this change affect Submariner upgrade in any way?

submariner/ipsec-certificate-support.md
```diff
@@ -0,0 +1,86 @@
...
+### Manual Enablement
+
+- Certificate mode must be **explicitly enabled** by passing an option (e.g., `--ipsec-cert-mode`) to `subctl` during
+deployment or join.
+
+### Pluto Process Expectation
+
+- When certificate mode is enabled, Submariner **expects** that a compatible Pluto process (as started by OVN-Kubernetes)
+is already running at the required path (`/usr/local/bin/pluto` by default).
+- The Pluto control socket is expected at `/run/pluto/pluto.ctl`.
```
skitt: Why is the daemon’s path important? If Submariner can find the control socket, that should be enough, shouldn’t it?

submariner/ipsec-certificate-support.md
skitt: Why does the Gateway act as a signer? If it only issues requests, and the Broker signs those, it shouldn’t need to be a signer.

submariner/ipsec-certificate-support.md
aswinsuryan: yes , the submariner could find it that should be enough. removed the sentance.

submariner/ipsec-certificate-support.md
aswinsuryan: oh that was an errro , gateway is the who generates CSR not sign, corrected it.

submariner/ipsec-certificate-support.md
aswinsuryan: It can be a future enhancement , listening to ovn ipsec mode we change it in submarine.

submariner/ipsec-certificate-support.md
aswinsuryan: yes the leader will have to do the same process.

submariner/ipsec-certificate-support.md
aswinsuryan: we cannot support mixed mode ,as the the other cluster needs to be in certificate mode as well, as it needs to have the CA certificate to validate submariner certificate.and there will be no PSK honored in this mode.

submariner/ipsec-certificate-support.md
aswinsuryan: we will not be able to upgrade from a psk mode to certificate mode , will need a reinstall. Upgrading from a lower version to higher version of a submariner configured in certificate mode should be transparent.

skitt (approved):

submariner/ipsec-certificate-support.md
tpantelis: So each gateway will have to issue its own signing request. So the secret name will have to include something unique to the gateway's `Endpoint`, like `Hostname`.

submariner/ipsec-certificate-support.md
tpantelis: We should probably mention this in the doc.

submariner/ipsec-certificate-support.md
yboaron: In that case maybe we should add a note to this proposal that  cert-mode  must be set in all clusters in cluster set

PR #263, opened 2025-12-19
Describe the subctl connect enhancement

### LINKED ISSUE #259: subctl: provide a connect operation combining deploy-broker and join

A `subctl connect` operation which connects two or more clusters using Submariner.

**Why is this needed**:

For small numbers of clusters, having a separate `deploy-broker` operation producing a file that needs to be provided when joining is cumbersome. On top of that, semantically, there’s an ambiguity with `join` — it joins a cluster to a clusterset (or put another way, to a broker), but users may want to “join” clusters that will communicate with each other.

Having access to all the clusters also means that some parameters can be auto-configured, or at least checked.

skitt:
Fixes #259

### DIFF
--- a/subctl/issue-259-connect.md
+++ b/subctl/issue-259-connect.md
@@ -1,0 +1,145 @@
+# Provide a `connect` operation combining `deploy-broker` and `join`
+
+## Summary
+
+For end users, the purpose of Submariner is to connect clusters to
+each other, enabling communications between them and providing
+additional services on top. The fact that a broker, or hub cluster, is
+involved, is a secondary concern; however that’s the first step
+involved in setting up Submariner with `subctl`.
+
+This enhancement proposal describes a `connect` subcommand which
+configures two or more clusters together, without the separate
+`deploy-broker` and `join` steps, or the use of a `broker-info.subm`
+file.
+
+## Proposal
+
+A new `connect` operation, taking two or more Kubernetes contexts,
+configures the corresponding clusters so that they are connected,
+where possible and where this doesn’t result in conflicts with
+existing configuration on the clusters. The contexts are specified
+using the existing patterns in `subctl`: `--contexts` followed by a
+comma-separated list of context names.
+
+## Design Details
+
+### Base scenario: no existing Submariner deployment
+
+When the contexts point to clusters with no existing Submariner setup,
+`connect` chooses one of the clusters to host the broker, deploys it
+there, and connects all the clusters using the newly-deployed broker.
+
+Broker selection should check for connectivity between the clusters:
+the broker should only be installed on a cluster which is accessible
+from all others. A `--broker-context` flag is provided to allow broker
+selection by the user.
+
+**No** `broker-info.subm` file is produced; see subsequent sections
+for scenarios describing deployments with existing brokers.
+
+`connect` supports all the existing options on `deploy-broker`,
+including component selection, Globalnet configuration etc., apart
+from those related to `broker-info.subm`. Since a single invocation of
+`connect` has knowledge of more than one cluster, some additional
+checks can be performed; in particular, the cluster CIDRs can be
+checked for overlaps, with the following behaviour:
+
+* in the default setup (no Globalnet configuration), Globalnet is
+  automatically enabled if the CIDRs overlap;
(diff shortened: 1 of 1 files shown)


subctl/issue-259-connect.md
```diff
@@ -0,0 +1,130 @@
...
+
+If the broker is inaccessible from any of the additional clusters,
+`connect` fails with an appropriate error, without changing any of the
+configuration on any of the clusters.
+
+If all the provided contexts are already connected, `connect` doesn’t
+do anything and exits successfully.
+
+If a broker is found indirectly (that is to say, it isn’t listed in
+the contexts to be connected), and the cluster hosting it isn’t
```
tpantelis: By indirectly, I assume you mean found via the `BrokerK8s` fields in the `Submariner` resource in an existing installation found amongst the to-be-connected contexts. Can we explicitly state this for clarification?

We should also check all the to-be-connected contexts to ensure that, for ones that already have submariner installed, they all point to the same broker,

subctl/issue-259-connect.md
```diff
@@ -0,0 +1,130 @@
...
+   deployments. If a Submariner broker is found, remember it; if a
+   Submariner deployment is found, determine the corresponding broker
+   and remember it.
+
+2. If more than one broker is found in the above step, fail.
+
+3. If no broker is found, run connectivity tests from each cluster to
+   all the other clusters’ API server. If any cluster is not
+   accessible from another cluster, remove it from the set of broker
+   candidates. If the set of broker candidates ends up empty, fail.
```
tpantelis: Maybe we should cap this search based on the number clusters to connect? This may get unwieldy with too many clusters, in which case we can make the user provide the `broker-context`.

subctl/issue-259-connect.md
```diff
@@ -0,0 +1,130 @@
...
+
+4. If a broker is found, run connectivity tests from each additional
+   cluster to its API server. If any of the tests fail, fail.
```
tpantelis: Wouldn't this have already been determined in step 3, ie if a broker was found that means all the provided contexts can access it?

subctl/issue-259-connect.md
```diff
@@ -0,0 +1,130 @@
...
+
+5. Check for overlapping CIDRs among the clusters. If any overlap:
+
+   1. If an existing broker was found, and Globalnet isn’t enabled,
```
tpantelis, line 118: In this case we should also ensure the new clusters don't overlap with any existing clusters already connected via the broker.

subctl/issue-259-connect.md
skitt: Yes, see the first paragraph in this section:

> If more than one broker is found, `connect` fails with an appropriate error.

subctl/issue-259-connect.md
skitt: Step 3 runs if no broker is found in step 1 (“If no broker is found”). Step 4 runs if a broker is found (“If a broker is found”). I’ll clarify that these relate to step 1.

subctl/issue-259-connect.md
tpantelis: OK but perhaps we should make this determination explicit above, ie "..._an existing Submariner deployment, all such contexts are used to find the
corresponding brokers **via examining the broker-related fields in the Submariner resource**."_. 

Also to clarify "_If a broker is found indirectly_...", how about "_If a broker isn't found amongst the contexts to be connected_...".

tpantelis (approved):

subctl/issue-259-connect.md
```diff
@@ -0,0 +1,145 @@
...
+
+Broker selection should check for connectivity between the clusters:
+the broker should only be installed on a cluster which is accessible
+from all others. A `--broker-context` flag is provided to allow broker
+selection by the user.
+
+**No** `broker-info.subm` file is produced; see subsequent sections
+for scenarios describing deployments with existing brokers.
+
+`connect` supports all the existing options on `deploy-broker`,
```
vthapar, line 41: What about different `join` options? e.g. `cable-driver`, `gateways` etc. ? There is also `cloud-prepare` and options it has.

vthapar (changes requested):

